PR #123, opened 2025-05-27
Localization setting

### LINKED ISSUE #84: Set up localization such that the dashboard is accessible in multiple languages
This issue relates to setting up localization for the project in an extendable way. I would suggest using the i18n library.

*Implementation Requirements*:
* Set-up typesafe localization for the project with initial values for 'en' and 'da'
* Add a general client-side component `LocaleText`, which can render as any text component and receive the locale key as prop. 
* Add a selector component that can change the language to the supported languages.
* if a component uses `LocaleText` the text must be reactive to changes from the selector (inside both client-side and server-side context)
* We should add a debug mode, such that the localization-keys are displayed instead of the locale-text. I recommend making it accessible through the url-query-param: `?displayLocalizationKeys=true`

flopper123:
This PR adds *SSR* localization through to the better-analytics-dashboard -  Closes #84 .

---

### Implementation overview
* I have implemented this on key-elements on the `DashboardPageClient` for now, to keep the scope minimal.
* Install the 'country-flag-icons' library - this is a very useful library for rendering country-flags.  It offers version of each flag as unicode, which can have a very positive impact on the performance of our client-side rendered graphs.
* Install the `shadcn` select-component - this is used for the `LocaleSelect`.
* Update `useSettings` composable to expose a `setSettings` method that updates the DB settings, internal settings, and force refresh the page (for SSR components to be updated).
* Update file-permissions of clickhouse folders to work on linux (missing +x and +w on log and script), this is not strictly in the scope of this PR, but its required for me to run the project since clickhouse auth changes

---

### Limitations & Future Work 
This is an intentionally scoped MVP and has the following limitations:
* No global text coverage – only dashboard text has been localized so far.
* Missing CI/CD to detect missing translations or automatic AI-powered translations.
I will open two seperate issues for these once I merge this


Lindharden:
Closing - implemented by #327

PR #159, opened 2025-06-06
Query filter UI/UX improvements

flopper123:
Added border for cleaner visual separation
image

Save / Cancel buttons are disabled if filters are unchanged
image

Save / Cancel buttons are active if filters have changes - cancelling will reset filters to current active filters
image

State if all filters have been removed but has not been saved
image

### DIFF
--- a/dashboard/src/components/filters/QueryFiltersSelector.tsx
+++ b/dashboard/src/components/filters/QueryFiltersSelector.tsx
@@ -1,82 +1,124 @@
-import { useCallback, useState } from "react";
-import { ChevronDownIcon, FilterIcon, PlusIcon, SettingsIcon } from "lucide-react";
-import { Popover, PopoverContent, PopoverTrigger } from "../ui/popover";
-import { Button } from "../ui/button";
-import { useQueryFiltersContext } from "@/contexts/QueryFiltersContextProvider";
-import { QueryFilterInputRow } from "./QueryFilterInputRow";
+import { useCallback, useMemo, useState } from 'react';
+import { ChevronDownIcon, FilterIcon, PlusIcon, SaveIcon, SettingsIcon, Undo2 } from 'lucide-react';
+import { Popover, PopoverContent, PopoverTrigger } from '../ui/popover';
+import { Button } from '../ui/button';
+import { useQueryFiltersContext } from '@/contexts/QueryFiltersContextProvider';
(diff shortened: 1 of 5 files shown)


dashboard/src/app/dashboard/[dashboardId]/pages/page.tsx
```diff
@@ -49,6 +50,7 @@ export default async function PagesPage({ params, searchParams }: PagesPageParam
           </div>
           <DashboardFilters />
         </div>
+        <ActiveQueryFilters />
```
Comment (Lindharden) [inline dashboard/src/app/dashboard/[dashboardId]/pages/page.tsx:53]: We should modify the DashboardFilters to include the list of filters, so that the list of filters are shown next to the filter dropdown.

dashboard/src/components/filters/QueryFiltersSelector.tsx
```diff
@@ -1,82 +1,125 @@
...
-  }, [addQueryFilter]);
+
+  const { queryFilters: contextQueryFilters, setQueryFilters } = useQueryFiltersContext();
+  const {
+    queryFilters,
+    setQueryFilters: setLocalQueryFilters,
+    addEmptyQueryFilter,
+    removeQueryFilter,
+    updateQueryFilter,
+  } = useInstantLocalQueryFilters();
```
Lindharden, line 22: Why do we create useInstantLocalQueryFilters and why this naming convention?  What is its purpose? Could we consider moving these inside the context or local functions?

Lindharden:
Looks a lot better now! I'm still a bit unsure why we include the id in the queryfilters type. It seems to add unnecessary complexity that we have to work-around the id by omitting it and generating temp ids, when we dont use it for anything. We don't store filters based on id, and order doesn't matter either - we should be able to omit it entirely?

Lastly, I think we need a hover effect on the filter badges, otherwise there's no feedback on what happens when clicked. I think either the X should be the only clickable part that removes it, or apply a hover effect that clearly accentuates the remove button when hovered. Otherwise it's unclear that you're removing it when clicking on the badge.

Lindharden:
> Looks a lot better now! I'm still a bit unsure why we include the id in the queryfilters type. It seems to add unnecessary complexity that we have to work-around the id by omitting it and generating temp ids, when we dont use it for anything. We don't store filters based on id, and order doesn't matter either - we should be able to omit it entirely?

I have looked a bit at the code not shown in this PR and I see it's necessity. :)

PR #173, opened 2025-06-07
Nginx reverse proxy

flopper123:


Lindharden (approved):

Lindharden:
Why don't we make it generate a SSL cert via certbot, so that anyone who wants to run this build doesn't need to provide their own SSL cert?

flopper123:
> Why don't we make it generate a SSL cert via certbot, so that anyone who wants to run this build doesn't need to provide their own SSL cert?

That would be optimal indeed - would be nice to run certbot in a Docker container

PR #192, opened 2025-06-15
Updated integration sheet with data-server-url

### LINKED ISSUE #175: Add "data-server-url" to integration sheet
As the `analytics.js` needs to now the endpoint, I've added an attribute called `data-server-url` to the analytics script.
This should also be reflected in the integration sheet

Image

flopper123:
Closes #175

### DIFF
--- a/.env.production.example
+++ b/.env.production.example
@@ -54,3 +54,8 @@ NEXT_PUBLIC_TRACKING_SERVER_ENDPOINT="https://betterlytics.io"
 # Dashboard default user credentials
 ADMIN_EMAIL=admin@example.com
 ADMIN_PASSWORD=admin
+
+# Feature flags
+ENABLE_DASHBOARD_TRACKING=false
+ENABLE_REGISTRATION=true
+IS_CLOUD=true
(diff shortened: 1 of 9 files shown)


dashboard/src/components/integration/IntegrationSheet.tsx
```diff
@@ -51,7 +51,7 @@ export function IntegrationSheet({ open, onOpenChange }: IntegrationSheetProps)
   }, [isVerified]);
 
   const trackingScript = siteId
-    ? `<script async src="https://analytics.example.com/tracker.js" data-site-id="${siteId}"></script>`
+    ? `<script async src="https://betterlytics.io/analytics.js" data-site-id="${siteId}" data-server-url="https://betterlytics.io/track"></script>`
```
Lindharden, line 54: This will show the wrong URL for self-hosters. We should use NEXT_PUBLIC_BASE_URL or a similar config to define the tracking URL, so that the integration sheet matches the environment.

Lindharden (changes requested):
See comments

PR #220, opened 2025-06-22
Fixed sidebar related issues on geography page

flopper123:


Lindharden:
Much better. But it is still possible to scroll on the page. Since the world map is draggable and panable, it should not be possible to scroll on the page

{E4DC4D18-8699-43CC-A685-9BA80E5D0694}

Lindharden (changes requested):
Looks good. See comment :)

Lindharden (approved):

PR #249, opened 2025-06-28
Added production server setup readme

flopper123:
image


Lindharden:
If we do this, I think we should update the production env example to reflect self-host configs

PR #259, opened 2025-06-29
TimeZone

flopper123:
- Added browser timezone as used timezone
- Refactored of time selector


dashboard/src/components/TimeRangeSelector/hooks/useTimeRangeHandlers.ts
```diff
@@ -0,0 +1,144 @@
...
+    },
+    [updateTempState],
+  );
+
+  const handleCompareStartDateSelect = useCallback(
+    (date: Date | undefined) => {
+      if (!date || !periodDurationDays) return;
+      // Calculate exact duration in milliseconds to preserve time
+      const durationMs = periodDurationDays * 24 * 60 * 60 * 1000;
+      const compareEnd = new Date(date.getTime() + durationMs);
```
Lindharden, line 95: Why do we need to do this? If the user selects a start date, isn't new Date(date.getTime()) sufficient? It seems that we're shifting the start day to 1 day later than the selected date?

dashboard/src/components/TimeRangeSelector/hooks/useTimeRangeHandlers.ts
```diff
@@ -0,0 +1,144 @@
...
+    [updateTempState, periodDurationDays],
+  );
+
+  const handleCompareEndDateSelect = useCallback(
+    (date: Date | undefined) => {
+      if (!date || !periodDurationDays) return;
+
+      // Calculate exact duration in milliseconds to preserve time
+      const durationMs = periodDurationDays * 24 * 60 * 60 * 1000;
+      const compareStart = new Date(date.getTime() - durationMs);
```
Lindharden, line 110: Same question as above :)

dashboard/src/components/TimeRangeSelector/hooks/useTimeRangeHandlers.ts
```diff
@@ -0,0 +1,144 @@
...
+      updateTempState({
+        compareStart: compareStart,
+        compareEnd: date,
+      });
+    },
+    [updateTempState, periodDurationDays],
+  );
+
+  const handleApply = useCallback(() => {
+    // Validate granularity before applying
```
Lindharden, line 120:
```suggestion
```

dashboard/src/components/TimeRangeSelector/index.tsx
Lindharden, line 1: Can we possibly rename this to something more descriptive than index? :)

dashboard/src/components/TimeRangeSelector/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+    handleCompareStartDateSelect,
+    handleCompareEndDateSelect,
+    handleApply,
+  } = useTimeRangeHandlers({
+    tempState,
+    updateTempState,
+    allowedGranularities,
+    periodDurationDays,
+    onApply: handleApplyChanges,
+  });
```
Lindharden, line 65: Perhaps move this next to the useTimeRangeState so all hooks are next to each other

dashboard/src/utils/dateFormatters.ts
```diff
@@ -12,6 +12,21 @@ export function toDateTimeString(date: string | Date): DateTimeString {
   return d.toISOString().replace('T', ' ').slice(0, 19);
 }
 
+/**
+ * Formats date strings to Clickhouse datetime column format
+ * @param dateTime DateTimeString
+ * @returns DateTimeString
+ */
+export function toClickHouseGridStartString(dateTime: DateTimeString): DateTimeString {
```
Lindharden, line 20: did you mean toClickHouseDateStartString (or toClickHouseDateTimeStartString)? :)

dashboard/src/lib/ba-query.ts
```diff
@@ -37,21 +39,27 @@ function getFilterQuery(queryFilters: QueryFilter[]) {
...
 function getGranularitySQLFunctionFromGranularityRange(granularity: GranularityRangeValues) {
-  const mappedGranularity = granularityMapper[granularity];
-  const validatedGranularity = GranularitySchema.parse(mappedGranularity);
-  return SQL.Unsafe(validatedGranularity);
+  const interval = granularityIntervalMapper[granularity];
+  const validatedInterval = GranularityIntervalSchema.parse(interval);
+  return (column: z.infer<typeof DateColumnSchema>, date: DateTimeString) => {
+    const validatedColumn = DateColumnSchema.parse(column);
+    const alignedDate = toClickHouseGridStartString(date);
+    return safeSql`toStartOfInterval(${SQL.Unsafe(validatedColumn)}, INTERVAL ${SQL.Unsafe(validatedInterval)}, ${SQL.DateTime({ granulairty_origin_date: alignedDate })} - INTERVAL 10 YEAR)`;
```
Lindharden, line 61: Can you explain the -10 year interval? perhaps in a comment

dashboard/src/utils/brand.ts
Lindharden, line 1: is this unused?

dashboard/src/utils/dateHelpers.ts
```diff
@@ -1,7 +1,7 @@
 export function parseClickHouseDate(dateString: string): Date {
   // Force UTC interpretation by appending 'Z' if not already present
-  const utcDateString = dateString.includes('Z') ? dateString : `${dateString}Z`;
-  return new Date(utcDateString);
+  const DateString = dateString.includes('Z') ? dateString : `${dateString}Z`;
+  return new Date(DateString);
```
Lindharden, line 4: Is this forced interpretation still needed considering we're assuming always UTC on server-side?

dashboard/src/utils/timeRanges.ts
```diff
@@ -10,73 +10,60 @@ export interface TimeRangePreset {
...
     label: 'Last 7 days',
     value: '7d',
-    getRange: (userTimezone = 'UTC') => {
-      const now = new Date();
-      const end = endOfDayInTimezone(now, userTimezone);
-      const start = startOfDayInTimezone(subDays(now, 6), userTimezone);
+    getRange: () => {
+      const helper = createTimezoneHelper();
+      const yesterday = subDays(new Date(), 1);
+      const end = helper.endOfDayInUserTimezone(yesterday);
```
Lindharden, line 29: Last 7 days should be inclusive of the current day :)

dashboard/src/utils/timeRanges.ts
```diff
@@ -10,73 +10,60 @@ export interface TimeRangePreset {
...
     label: 'Last 28 days',
     value: '28d',
-    getRange: (userTimezone = 'UTC') => {
-      const now = new Date();
-      const end = endOfDayInTimezone(now, userTimezone);
-      const start = startOfDayInTimezone(subDays(now, 27), userTimezone);
+    getRange: () => {
+      const helper = createTimezoneHelper();
+      const yesterday = subDays(new Date(), 1);
+      const end = helper.endOfDayInUserTimezone(yesterday);
```
Lindharden, line 40: Same as above :)

dashboard/src/utils/timeRanges.ts
```diff
@@ -10,73 +10,60 @@ export interface TimeRangePreset {
...
     label: 'Last 3 months',
     value: '3mo',
-    getRange: (userTimezone = 'UTC') => {
-      const now = new Date();
-      const end = endOfDayInTimezone(now, userTimezone);
-      const start = startOfDayInTimezone(subMonths(now, 3), userTimezone);
+    getRange: () => {
+      const helper = createTimezoneHelper();
+      const yesterday = subDays(new Date(), 1);
+      const end = helper.endOfDayInUserTimezone(yesterday);
```
Lindharden, line 51: Same as above :)

dashboard/src/utils/timezoneHelpers.ts
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * Convenience functions for common operations
+ */
+
+/**
+ * Create a timezone helper for a specific user timezone
+ */
+export function createTimezoneHelper(): TimezoneAwareDateHelper {
+  return new TimezoneAwareDateHelper();
+}
```
Lindharden, line 85: We don't really use class anywhere else, besides for components. Can you explain the reason for using class? :)

Lindharden (changes requested):
Seems to handle timezones a lot better than previously! I've left some comments. And here's some discoveries from my testing:

1) 
Selecting last 7 days  (and last 28 days and last 3 months) makes it select the day before the current day: 
image.


2)
Page Views summary card doesn't seem to reflect the same data as the graph:
{9D07E686-1C9E-494B-905C-FC51A0D62A61}
in summary card it states 6 page views, but graphs shows more than 20 on different data points.


3)
Also seems like it shows no comparison when selecting a comparison range with no captured data:
{717B4810-4734-4F76-98C2-C1483103099A}



4)
Selecting 24 hours and comparing with the same day makes the data appear odd:
{BC3EB71D-DDEA-4E58-B1EC-053ECBEF34AF}

Same here, where comparison range is the same dates, but it says 0 in compared period, but it should be the same given it's the same range selected:
{B93D1B37-6D87-4FA5-AC9A-AB2771A2BBEB}


5)

Selecting a smaller granularity before switching to a larger range doesn't change the selected granularity:
{42D5B680-E660-48D1-9846-354F2723C38D}
Ideally the user should get the feedback by visibly seeing that their granularity is changed. Otherwise the user might click apply and looking at the data unknowing that granularity has changed unexpectedly

6)

The user dropdown appears behind the time selector:
{E6E99307-704C-42C2-8F90-EB5457523380}

dashboard/src/utils/timeRanges.ts
flopper123: So, Google Analytics / other site analytics do not count the current day if you select more than last 24 hours, and I think it's better to leave the current day out, as there wont be the complete analytics for the day yet. Thoughts?

dashboard/src/utils/timeRanges.ts
flopper123: Same response as above :)

dashboard/src/utils/timezoneHelpers.ts
flopper123: It was originally passed on a timezone, however, now that I don't use that anymore I'll refactor it to not be a class 😄 👍

dashboard/src/utils/dateHelpers.ts
flopper123: Yeah, because this function is used both on frontend and on backend.
So if it's used on backend and the "z" is not there, it'll be treated as a local timezone date - that we do not want

dashboard/src/utils/dateFormatters.ts
flopper123: No. Clickhouse calls it "grid" alignment, so this function aligns the DateTime to the start of a grid "cell"

dashboard/src/components/TimeRangeSelector/index.tsx
flopper123: Not possible - this one requires the `handleApplyChanges`. The `handleApplyChanges` requires the other hook. The `handleApplyChanges` is a hook itself though

dashboard/src/components/TimeRangeSelector/hooks/useTimeRangeHandlers.ts
flopper123: Updated

dashboard/src/components/TimeRangeSelector/hooks/useTimeRangeHandlers.ts
flopper123: Updated

dashboard/src/components/TimeRangeSelector/index.tsx
Lindharden: You reacted with thumbs up and resolved the comment, but the file is stilled named index.tsx

Lindharden (changes requested):
Time seems to be wrong in this case:

It shows 19:54 but the event occured at 20:54
image
Seems to only be the case when selecting 24h quick select with hours. 


It also seems that comparison doesn't work in this case:
image

Also seems like comparison range is no longer restricted to match the same range length as the time selected:
image



Everything else related to dates seems to work real nice now!

dashboard/src/utils/timeRanges.ts
```diff
@@ -1,89 +1,104 @@
...
+  const hours = timeOfDayDate.getHours();
+  const minutes = timeOfDayDate.getMinutes();
+  const seconds = timeOfDayDate.getSeconds();
+  const milliseconds = timeOfDayDate.getMilliseconds();
+
+  const newDate = new Date(date);
+
+  newDate.setHours(hours, minutes, seconds, milliseconds);
+
+  return newDate;
```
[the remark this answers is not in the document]
flopper123: Probably not, but just in case. Also, then it does what the function describes.

dashboard/src/components/TimeRangeSelector/hooks/useTimeRangeHandlers.ts
```diff
@@ -0,0 +1,209 @@
...
+      }
+
+      const compareEnd = endOfDay(date);
+      const compareStart = startOfDay(new Date(compareEnd.getTime() - timeDifference));
+      updateTempState({
+        compareStart,
+        compareEnd,
+      });
+    },
+    [updateTempState, periodDurationDays],
```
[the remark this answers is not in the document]
flopper123: Sure 👍

dashboard/src/components/TimeRangeSelector/hooks/useTimeRangeHandlers.ts
```diff
@@ -0,0 +1,209 @@
...
+      }
+
+      const compareStart = startOfDay(date);
+      const compareEnd = endOfDay(new Date(compareStart.getTime() + timeDifference));
+      updateTempState({
+        compareStart,
+        compareEnd,
+      });
+    },
+    [updateTempState, periodDurationDays],
```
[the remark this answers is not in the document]
flopper123: Sure 👍

dashboard/src/components/InteractiveChart.tsx
```diff
@@ -23,6 +23,9 @@ interface InteractiveChartProps {
 
 const InteractiveChart: React.FC<InteractiveChartProps> = React.memo(
   ({ title, data, color, formatValue, granularity, comparisonMap }) => {
+    const axisFormatter =
+      granularity === undefined || granularity === 'day' ? timeFormat('%b %d') : timeFormat('%b %d - %H:%M');
```
Lindharden, line 27: Can we reuse the "defaultDateLabelFormatter" in chartUtils.ts?

Lindharden (changes requested):
Seems to work super nice now. 

There is this one bug in the comparison range, where the selected date doesn't apply:
chrome_y85cIGMT1s

dashboard/src/components/InteractiveChart.tsx
flopper123: The interface is different though. But I've extracted it 👍

Lindharden: Okay :)

Lindharden (approved):

PR #260, opened 2025-06-29
Added sitemap and robots files

Lindharden:
Added sitemap
Added robots config: disallowing /dashboard /dashboards /billing

Fixed some env variables examples in .env.production.example

Added DPA to footer


dashboard/src/app/sitemap.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export default function sitemap(): MetadataRoute.Sitemap {
+  const baseUrl = env.NEXT_PUBLIC_BASE_URL;
+
+  return [
+    /****************** PUBLIC PAGES ******************/
+
+    {
+      url: baseUrl,
+      lastModified: new Date(),
```
flopper123, line 12: Is it correct/good practice to let `lastModified` be set to `new Date()` on every page?

dashboard/src/app/sitemap.ts
```diff
@@ -0,0 +1,102 @@
...
+      changeFrequency: 'monthly',
```
flopper123, line 13: Should some of these be set to a higher change frequency? I imagine that our docs/landing page will have weekly updates for the next while. Thoughts?

flopper123 (approved):

PR #263, opened 2025-06-29
Custom event docs

Lindharden:
Added docs for custom events.

I added basic but slightly different examples for multiple frameworks and use cases.

I added a pre-load queue for events that may be sent before the script has loaded.

Added prettierignore file for docs


docs/src/content/integration/custom-events.mdx
```diff
@@ -20,7 +22,417 @@ Custom events let you track specific actions users take on your website:
...
+  category: "electronics",
+  price: 299.99,
+  source: "search-results",
+});
+```
+
+These properties let you answer questions like:
+
+- **What** was clicked? (`buttonText: "Get Started"`)
+- **Where** on the page? (`location: "header"`)
```
flopper123, line 86: Suggestion: Align the example to match the question answers in this example

flopper123 (dismissed):

PR #268, opened 2025-07-05
Fixed time format and mobile dashboard filter ordering

flopper123:


dashboard/src/components/TimeRangeSelector.tsx
```diff
@@ -210,7 +210,7 @@ export default function TimeRangeSelector({ className = '' }: { className?: stri
 
   const displayRangeLabel = () => {
     if (currentActivePreset === 'custom' && startDate && endDate) {
-      return `${format(startDate, 'P')} - ${format(endDate, 'P')}`;
+      return `${format(startDate, 'dd/MM/yyyy')} - ${format(endDate, 'dd/MM/yyyy')}`;
```
Lindharden, line 213:
```suggestion
      return `${startDate.toLocaleDateString()} - ${endDate.toLocaleDateString()}`;
```

We should use users locale instead of only appealing to a certain format :)

Lindharden (changes requested):
Nice 👍

PR #275, opened 2025-07-06
228 remove comparison option on user journey page

### LINKED ISSUE #228: Remove comparison option on User Journey page
It might not make sense to show comparison data on the User Journey page.

So - we should probably either remove the option on said page, or find a nice way to present it.

Lindharden:
Closes #228


Lindharden:
duplicate of https://github.com/betterlytics/betterlytics/pull/276

PR #276, opened 2025-07-06
228 remove comparison option on user journey page

### LINKED ISSUE #228: Remove comparison option on User Journey page

Lindharden:
Closes #228

### DIFF
--- a/dashboard/src/components/dashboard/DashboardFilters.tsx
+++ b/dashboard/src/components/dashboard/DashboardFilters.tsx
@@ -4,12 +4,16 @@ import TimeRangeSelector from '@/components/TimeRangeSelector';
 import QueryFiltersSelector from '@/components/filters/QueryFiltersSelector';
 import { ActiveQueryFilters } from '../filters/ActiveQueryFilters';
 
-export default function DashboardFilters() {
+interface DashboardFiltersProps {
+  showComparison?: boolean;
+}
+
+export default function DashboardFilters({ showComparison = true }: DashboardFiltersProps) {
   return (
     <div className='space-y-2'>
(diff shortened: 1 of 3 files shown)


dashboard/src/components/dashboard/DashboardFilters.tsx
```diff
@@ -4,12 +4,16 @@ import TimeRangeSelector from '@/components/TimeRangeSelector';
 import QueryFiltersSelector from '@/components/filters/QueryFiltersSelector';
 import { ActiveQueryFilters } from '../filters/ActiveQueryFilters';
 
-export default function DashboardFilters() {
+interface DashboardFiltersProps {
+  showComparison?: boolean;
```
flopper123, line 8: This has a code smell to it, which is caused by the components not being composable.
It's one of the things I'd like to see fixed with a cleanup of the filter system.
I can approve it, if you really want it to be a part of the release though. Let me know what you think.

flopper123 (changes requested):

dashboard/src/components/dashboard/DashboardFilters.tsx
Lindharden: Your comment doesn't provide much. Are you suggesting we entirely ignore this issue until someone has cleaned up the dashboard filters to utilize composed components? 

In which case, why dont we just make an issue on making this change to the filter component instead of blocking an already-done implementation for something? This also helps with the refactoring, knowing what some of the use cases are.

flopper123: True that

flopper123 (approved):

PR #278, opened 2025-07-06
Improved login error message display and redirection

### LINKED ISSUE #261: Failed login attempt redirects to 404
Logging in with incorrect email or password causes the user to be redirected to the a 404 page:
Image

Lindharden:
Closes #261

### DIFF
--- a/dashboard/src/components/auth/LoginForm.tsx
+++ b/dashboard/src/components/auth/LoginForm.tsx
@@ -23,11 +23,11 @@ export default function LoginForm() {
       });
 
       if (result?.error) {
-        setError('Invalid email or password');
+        setError('Invalid email or password. Please check your credentials and try again.');
         setIsLoading(false);
       }
     } catch {
-      setError('An error occurred during sign in');
+      setError('An error occurred during sign in. Please try again.');
       setIsLoading(false);
     }
   };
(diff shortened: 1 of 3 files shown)


dashboard/src/components/auth/LoginForm.tsx
```diff
@@ -23,11 +23,11 @@ export default function LoginForm() {
...
 
       if (result?.error) {
-        setError('Invalid email or password');
+        setError('Invalid email or password. Please check your credentials and try again.');
         setIsLoading(false);
       }
     } catch {
-      setError('An error occurred during sign in');
+      setError('An error occurred during sign in. Please try again.');
       setIsLoading(false);
```
flopper123, line 31: Now that you're already working on this, could you please rewrite the `isLoading` into a `useTransition`? 😇 
Because, if for some reason the redirect is not working properly, then the `isLoading` will remain true forever, as no finally block catches that case 😄

flopper123 (approved):

dashboard/src/components/auth/LoginForm.tsx
Lindharden: Yes, good suggestion 👍

flopper123 (approved):

PR #308, opened 2025-07-07
Using locale date

### LINKED ISSUE #219: Use European time-format instead of American time-format
The current format is: MM/DD/YYYY
I would expect, as a European, that it would be: DD/MM/YYYY

This issue might be fixed in https://github.com/betterlytics/betterlytics/issues/205 

The times in our application looks as:
Image

flopper123:
Closes #219


dashboard/src/components/TimeRangeSelector/TimeRangeSelector.tsx
```diff
@@ -76,8 +76,8 @@ export function TimeRangeSelector({
 
   const displayRangeLabel = () => {
     if (currentActivePreset === 'custom' && context.startDate && context.endDate) {
-      const startLabel = format(context.startDate, 'P');
-      const endLabel = format(context.endDate, 'P');
+      const startLabel = context.startDate.toLocaleDateString();
+      const endLabel = context.endDate.toLocaleDateString();
```
Woldbye, line 80: It's a bad idea to rely on toLocaleDateString formatting as it gives inconsistent output between browsers. Why not rely on the format from the already used date-fns lib??

Woldbye (approved):

PR #314, opened 2025-07-08
Added optional search column to DataTable

flopper123:
Added optional search column. If specified, an Input will be added over the table used for searching on specified column:

image

On input, it will filter on said column:

image

### DIFF
--- a/dashboard/src/components/DataTable.tsx
+++ b/dashboard/src/components/DataTable.tsx
@@ -2,17 +2,19 @@
 
 import { useState } from 'react';
 import {
-  ColumnDef,
-  SortingState,
   flexRender,
   getCoreRowModel,
   getFilteredRowModel,
   getSortedRowModel,
   useReactTable,
-  Row,
+  type ColumnDef,
+  type SortingState,
+  type Row,
+  type ColumnFiltersState,
 } from '@tanstack/react-table';
 import { Table, TableBody, TableCell, TableHead, TableHeader, TableRow } from '@/components/ui/table';
 import { ArrowUp, ArrowDown } from 'lucide-react';
+import { Input } from './ui/input';
 
 interface DataTableProps<TData, TValue> {
   columns: ColumnDef<TData, TValue>[];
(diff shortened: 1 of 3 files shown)


dashboard/src/components/DataTable.tsx
```diff
@@ -28,81 +31,96 @@ export function DataTable<TData, TValue>({
...
+          className='mb-2 max-w-sm'
+        />
+      )}
+      <div
+        className={`rounded-lg ${className || ''} overflow-hidden border border-gray-200 dark:border-slate-700`}
+      >
+        <Table>
+          <TableHeader className='bg-gray-50 dark:bg-slate-800'>
+            {table.getHeaderGroups().map((headerGroup) => (
+              <TableRow key={headerGroup.id} className='border-b border-gray-200 dark:border-slate-700'>
```
Woldbye, line 68: Quibble: it seems like the css properties are shared between many items here. If we group the  css into classes, this will make it easier to maintain and easier to read.

Woldbye, line 68: Shouldn't this also have hover effects like the previous code ?

Woldbye (approved):

dashboard/src/components/DataTable.tsx
flopper123: It still has! 😄
GitHub just shows this as replacing the old code. I have not done that. The hover effect is still on it, just below this comment
image

flopper123: This might be a good idea to look into at some point. But in this MR I haven't touched the table code. It just looks like changes since it's been tabbed one time, as it's now being wrapped in a fragment 😄

Lindharden (changes requested):
Code looks good, but the search bar introduces an awkward empty space and disalignment between controls and the datatable. The search bar should be aligned with the rest of the "filter" controls:

image

Lindharden (approved):
Looks good :)

PR #321, opened 2025-07-09
255 add account verification process

### LINKED ISSUE #255: Add account verification process

Lindharden:
Added verification flow.

A verification link is sent upon sign up, or a resend  is requested. A banner is prominently shown in the dashboard as we want users to verify early in order to properly send alerts and notifications and more. 

When the user clicks the verification link, they're redirected to /verify-email endpoint, which will verify their token and mark the account verified. If the user has a session when clicking on the link, they'll be redirected to /dashboards with ?verified=1, which is used to show a celebratory modal. If the user does not have a session, they'll be redirected to login and shown the same celebratory modal.

In order to refresh the user's authentication token (the one stored in their cookies), I had to implement a trigger to refresh their session.

Using the last stored user token in the database to add rate-limiting to prevent email spam.

Added feature flag to disable verification. We dont want verification for self-hosters or dev.

Restriction modal that appears when billing is proceeded without verified account
{081111F9-4013-4BA2-A3F8-183B845615A5}

Celebratory modal to show the user they have now verified their account:
{C3597272-0DF7-479B-B092-E742FA82B215}

Verification Email:
{91B005E7-D1D1-4765-8641-8FB7317C75EA}

Verification banner in dashboard and /billing page
{C8AA3D93-1A63-4D92-BF1B-0AC5BFED1271}

Closes #255


dashboard/src/app/actions/verification.ts
```diff
@@ -1,18 +1,64 @@
...
+
+export async function resendVerificationEmailAction(formData: SendVerificationEmailData) {
+  try {
+    const validatedData = SendVerificationEmailSchema.parse(formData);
+    const { email } = validatedData;
+
+    const rateLimitCheck = await checkRateLimit(email);
+
+    if (!rateLimitCheck.allowed && rateLimitCheck.nextAllowedAt) {
+      const waitTime = Math.ceil((rateLimitCheck.nextAllowedAt?.getTime() - Date.now()) / 60000);
```
flopper123, line 48: Quibble: The variable `rateLimitCheck.nextAllowedAt` is always defined here:
```suggestion
      const waitTime = Math.ceil((rateLimitCheck.nextAllowedAt.getTime() - Date.now()) / 60000);
```

dashboard/src/app/billing/BillingInteractive.tsx
```diff
@@ -30,6 +34,11 @@ export function BillingInteractive({ billingData }: BillingInteractiveProps) {
         return;
       }
 
+      if (!session?.user?.emailVerified) {
+        setShowVerificationModal(true);
+        return;
+      }
```
flopper123, line 40: Suggestion/discussion: Create a hook for handling user email management. Maybe called `useUserEmail` or `useUserVerification`.
It could use `isFeatureEnabled('enableAccountVerification')`, `session?.user?.email`, and `session?.user?.emailVerified` to create a cleaner verification API for the frontend.

dashboard/src/app/dashboard/[dashboardId]/layout.tsx
```diff
@@ -56,6 +57,13 @@ export default async function DashboardLayout({ children, params }: DashboardLay
                 <UsageUpgradeBanner billingDataPromise={getUserBillingData()} />
               </Suspense>
             )}
+            {isFeatureEnabled('enableAccountVerification') &&
+              session.user?.email &&
+              !session.user?.emailVerified && (
```
Comment (flopper123) [inline dashboard/src/app/dashboard/[dashboardId]/layout.tsx:62]: Regarding the hook from the other comment, it would be nice if this could be simplified to:
```suggestion
            {userEmail.verified === false && (
```

If the `isFeatureEnabled('enableAccountVerification')` is true, then `verified` could also just return true for instance

dashboard/src/services/email/template/email-verification-mail.ts
```diff
@@ -0,0 +1,139 @@
...
+    .join('');
+
+  return `<ul style="${emailStyles.unorderedList}">${listItems}</ul>`;
+}
+
+export function generateEmailVerificationText(data: EmailVerificationData): string {
+  const content = `
+Verify Your Email Address
+
+Hi ${data.userName},
```
flopper123, line 97: Nitpick:
```suggestion
Hi ${escapeHtml(data.userName)},
```

dashboard/src/services/verification.service.ts
```diff
@@ -0,0 +1,167 @@
...
+  }
+}
+
+export async function verifyEmail(data: VerifyEmailData): Promise<VerificationResult> {
+  if (!isFeatureEnabled('enableAccountVerification')) {
+    return {
+      success: false,
+      error: 'Account verification is not enabled',
+    };
+  }
```
flopper123, line 79: Suggestion/discussion: Should this return true? Since if account verification is not enabled, it would be nice to still have access to all features. Let me know what you think 😄

dashboard/src/services/verification.service.ts
```diff
@@ -0,0 +1,167 @@
...
+      return { allowed: true };
+    }
+
+    const cooldownAgo = subMinutes(now, RESEND_COOLDOWN_MINUTES);
+    const tokenExpiryAgo = subHours(now, VERIFICATION_TOKEN_EXPIRY_HOURS);
+
+    if (isBefore(recentToken.createdAt, tokenExpiryAgo)) {
+      const nextAllowedAt = addHours(recentToken.createdAt, VERIFICATION_TOKEN_EXPIRY_HOURS);
+      return { allowed: false, nextAllowedAt };
+    }
```
flopper123, line 155: It would be nice with a comment on the first case here, as I'm a little unsure exactly why the next statement is not enough 😄

Woldbye:
Nice work on this PR!

I haven't worked with Mailersend before, as I usually use Twilio's SendGrid (they offer the same number of free emails per month). That said, I have two main remarks:

* One of the biggest challenges with email delivery is avoiding the spam folder. I didn't see any specific handling for this in the PR—do you know if Mailersend addresses this directly on the account level (e.g., via domain authentication, SPF/DKIM setup, etc.)?

* Email services are notoriously inconsistent, and with only 3,000 free emails per month, I think it's important to have some form of automatic monitoring in place. Ideally, the Mailersend account should notify us when we're approaching our quota or if the service experiences downtime. This seems outside the scope of this PR, but should probably be added as a new issue upon completion of this.

Lindharden:
> * One of the biggest challenges with email delivery is avoiding the spam folder. I didn't see any specific handling for this in the PR - Have you taken any special actions to handle this on the Mailersend account (via domain authentication, SPF/DKIM setup, etc.)?

SPF/DKIM is setup for this! I'm planning to enable DMARC in the future as well :)

> * Email services are notoriously inconsistent, and with only 3,000 free emails per month, I think it's important to have some form of automatic monitoring in place. Ideally, the Mailersend account should notify us when we're approaching our quota or if the service experiences downtime. This seems outside the scope of this PR, but should probably be added as a new issue upon completion of this.

Thanks for being considerate! I've reached out to MailerSend to hear about alerts. Currently I'm monitoring limits and we have a usage-based pricing setup with a limit to avoid unexpected excessive use.

dashboard/src/app/billing/BillingInteractive.tsx
Lindharden: I thought a bit about this. Would be nice to have some smart function for this. But I don't know if I'm a fan of maintaining a client-sided hook and a server function for this. 

I also don't think that featureEnabled should be mixed with isVerified as these technically serve two-different purposes.

Lmk if you disagree :)

dashboard/src/components/accountVerification/VerificationCelebrationModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+              <CheckCircle className='h-16 w-16 text-blue-600' />
+            </div>
+
+            <div
+              className={cn(
+                'space-y-4 transition-all delay-200 duration-700',
+                showContent ? 'translate-y-0 opacity-100' : 'translate-y-4 opacity-0',
+              )}
+            >
+              <h2 className='text-2xl font-bold text-blue-900'>Email Verified Successfully, {displayName}!</h2>
```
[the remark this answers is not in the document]
Lindharden: Yess that's a oversight from refactoring. Fixed :) 

I created a userUtils for a better consistent fallback

dashboard/src/services/email/template/email-verification-mail.ts
Lindharden: This is the text version of the email for those who cannot render HTML -  I believe we cannot escape here :)

dashboard/src/services/verification.service.ts
Lindharden: These are services for setting users verified using the service. It shouldn't matter what it returns tbh. But I believe feedback is better when used unexpectedly. It will only appear if mails are sent, which also wont happen :)

dashboard/src/services/verification.service.ts
Lindharden: This case was a brainfart. I removed it entirely :)

dashboard/src/app/billing/BillingInteractive.tsx
flopper123: Very reasonable argument. Let's keep it as is for now, then we can always refactor later 😄

PR #327, opened 2025-07-10
Add Danish/English locale dictionaries

### LINKED ISSUE #84: Set up localization such that the dashboard is accessible in multiple languages

Woldbye:
This PR adds *SSR* localization through to the better-analytics-dashboard -  Closes #84 .

---

### Implementation overview
* Add full dictionary setup to contain SSR translations for any supported-locale. To support a locale simply add a ${supported-locale}.json dictionary file with same keys as 'en.json' to the project.
* Add custom translation function, this serves two purposes:
1. Typesafety, and editor auto-completion of dictionary keys
2. Allows us to fully customize it anyway we see fit. For example, add support for custom dictionary values, error handling etc. 
* Implemented translations on key-elements on the Dashboad Overview
* Add DEFAULT_LANGUAGE as an .env option

---

### Limitations & Future Work 
This is an intentionally scoped MVP and has the following limitations:
* No global text coverage – only dashboard text has been localized so far.
* Missing CI/CD to detect missing translations or automatic AI-powered translations.
* Add support for number argument to `t` function, that resolves to plural or singular value.
* Add section about available languages to docs

I will open four seperate issues for these once I merge this


dashboard/src/app/dashboard/[dashboardId]/layout.tsx
```diff
@@ -44,28 +47,30 @@ export default async function DashboardLayout({ children, params }: DashboardLay
...
-            {billingEnabled && (
-              <Suspense fallback={null}>
-                <UsageUpgradeBanner billingDataPromise={getUserBillingData()} />
-              </Suspense>
-            )}
-            <div className='flex w-full justify-center'>{children}</div>
-          </main>
-          {/* Conditionally render tracking script based on server-side feature flag */}
-          {shouldEnableTracking && siteId && <TrackingScript siteId={siteId} />}
+      <DictionaryProvider dictionary={dictionary} initialLanguage={language}>
```
Comment (flopper123) [inline dashboard/src/app/dashboard/[dashboardId]/layout.tsx:50]: Suggestion: Instead of wrapping it in here and in `dashboards/layout.tsx`, this should be added in `dashboard\src\app\Providers.tsx`. That would wrap the entire application in the `DictionaryProvider` 😄

dashboard/src/app/dashboards/layout.tsx
```diff
@@ -1,15 +1,21 @@
...
-export default function DashboardsLayout({ children }: DashboardsLayoutProps) {
+export default async function DashboardsLayout({ children }: DashboardsLayoutProps) {
+  const { dictionary, language } = await getDictionary();
+
   return (
-    <section className='h-full w-full'>
-      <BATopbar />
-      {children}
-    </section>
+    <DictionaryProvider dictionary={dictionary} initialLanguage={language}>
```
flopper123, line 14: ^ Read comment above

dashboard/src/components/charts/ChartTooltip.tsx
```diff
@@ -61,18 +64,20 @@ export function ChartTooltip({
...
-          <div className='text-muted-foreground text-xs'>vs {labelFormatter(comparisonData.compareDate)}</div>
+          <div className='text-muted-foreground text-xs'>
+            {dictionary.charts.tooltip.vs} {labelFormatter(comparisonData.compareDate)}
+          </div>
         )}
       </div>
 
       <div className='mb-3'>
-        <div className='text-muted-foreground mb-1 text-xs'>Current Period</div>
+        <div className='text-muted-foreground mb-1 text-xs'>{dictionary.charts.tooltip.currentPeriod}</div>
```
flopper123, line 74: Question: Why is it `dictionary.charts.tooltip.currentPeriod` here instead of `dictionary.t('charts.tooltip.currentPeriod')`?

dashboard/src/components/language/LanguageSelect.tsx
```diff
@@ -19,28 +20,30 @@ const LANGUAGE_TO_COUNTRYCODE = {
 } satisfies Record<SupportedLanguages, FlagIconProps['countryCode']>;
 
 export function LanguageSelect({ onUpdate, value: language, id }: LanguageSelectProps) {
+  const { dictionary, isLoading } = useDictionary();
```
flopper123, line 23: Question: Why is loading required here?

dashboard/src/components/language/LanguageSelect.tsx
```diff
@@ -19,28 +20,30 @@ const LANGUAGE_TO_COUNTRYCODE = {
...
+
   return (
-    <Select value={language} onValueChange={onUpdate}>
+    <Select value={language} onValueChange={onUpdate} disabled={isLoading}>
```
flopper123, line 26: Suggestion: For this file I would hard-code the languages to each be their own representation. For instance it would say:
Dansk
English
Deutsch

That way, each user can easily find their own language. Thoughts?

dashboard/src/contexts/DictionaryContextProvider.tsx
```diff
@@ -0,0 +1,77 @@
...
+  initialLanguage: SupportedLanguages;
+  children: React.ReactNode;
+};
+
+function addTFunction(dict: RawDictionary): BADictionary {
+  return {
+    ...dict,
+    t: (key: string) => {
+      return key.split('.').reduce((current: any, k) => current?.[k], dict) || key;
+    },
```
flopper123, line 26: Suggestion: Consider precompute this. You could simply create an object that is a "flat" version of the dictionary, such that the keys still match perfectly. That way the `t` function could work exactly the same for the server side where the hook is not available

dashboard/src/contexts/DictionaryContextProvider.tsx
```diff
@@ -0,0 +1,77 @@
...
+        setDictionary(newDictionary);
+        setCurrentLanguage(language);
+      } catch (error) {
+        console.error('Failed to load dictionary:', error);
+      } finally {
+        setIsLoading(false);
+      }
+    },
+    [currentLanguage],
+  );
```
flopper123, line 55: Suggestion: Consider using `useSuspense` instead of manual loading state management

dashboard/src/dictionaries/dictionaries.ts
```diff
@@ -0,0 +1,55 @@
+export const SUPPORTED_LANGUAGES = ['en', 'da'] as const;
+
+export const dictionaries = {
+  en: () => import(`@/dictionaries/en.json`).then((module) => module.default),
+  da: () => import('@/dictionaries/da.json').then((module) => module.default),
+};
```
flopper123, line 6: Suggestion: If these get precomputed, I think the typing will also become quite a bit simpler, since you won't have to check for nested key types 😁

dashboard/src/dictionaries/dictionaries.ts
```diff
@@ -0,0 +1,55 @@
...
+    ? `${Key}` | `${Key}.${NestedKeyOf<ObjectType[Key]>}`
+    : `${Key}`;
+}[keyof ObjectType & (string | number)];
+
+export type BADictionary = RawDictionary & {
+  t: (key: DictionaryKeys, ...args: unknown[]) => string;
+};
+
+export type SupportedLanguages = keyof typeof dictionaries;
+export const DEFAULT_LANGUAGE: SupportedLanguages = 'en';
```
flopper123, line 22: Suggestion: I think this should be an environment variable, such that self-hoster can just default to their own language

flopper123 (changes requested):
Overall a very nice PR! 🔥
Have a few comments I think that could improve it even further 😄

dashboard/src/components/language/LanguageSelect.tsx
Lindharden: I thought about this as well - I have seen this done elsewhere. Quite nice 👍

dashboard/src/dictionaries/dictionaries.ts
Woldbye: Resolving this, as it doesn't help to precompute this :( 

One alternative solution is to declare the keys as a interface somewhere, then we would have them at compile-time and run-time.

dashboard/src/contexts/DictionaryContextProvider.tsx
Woldbye: Resolving this as flattening the keys loses string interpolation advantages otherwise gained.

dashboard/src/components/language/LanguageSelect.tsx
Woldbye: Good catch - I will remove this.
We’re doing a full SSR navigation after the language is selected, so the component unmounts immediately. That means the disabled={isLoading} doesn’t provide any actual UX benefit in this flow.

dashboard/src/dictionaries/dictionaries.ts
Woldbye: I like the suggestion, but we need to ensure that the possible enumerations are well-documented, to ensure the self-hoster doesn't choose an unsupported value or format.

dashboard/src/contexts/DictionaryContextProvider.tsx
flopper123: Whops, meant `useTransition`. This is fine as is though 😄

flopper123 (dismissed):
Nice, looks good! 👏

Lindharden (approved):

PR #333, opened 2025-07-11
Track production app

flopper123:
Setup env for tracking entire app


dashboard/src/app/layout.tsx
```diff
@@ -34,6 +36,17 @@ export default function RootLayout({
...
   return (
     <html lang='en' suppressHydrationWarning>
+      {env.ENABLE_APP_TRACKING && (
+        <head>
+          <Script
+            async
+            src={`${process.env.NEXT_PUBLIC_ANALYTICS_BASE_URL}/analytics.js`}
+            data-site-id={env.APP_TRACKING_SITE_ID}
+            data-server-url={`${process.env.NEXT_PUBLIC_TRACKING_SERVER_ENDPOINT}/track`}
+            data-dynamic-urls='/dashboard/*'
```
Lindharden, line 46: We should add /dashboard/*/funnels/* as well. Otherwise we won't be getting aggregrated funnel data.

URL for funnel:
/dashboard/cmcut2v980005tsrwfzqad2io/funnels/cmcz0vi950001tsekogtwp29m

dashboard/src/app/layout.tsx
```diff
@@ -34,6 +36,17 @@ export default function RootLayout({
 }>) {
   return (
     <html lang='en' suppressHydrationWarning>
+      {env.ENABLE_APP_TRACKING && (
+        <head>
+          <Script
+            async
+            src={`${process.env.NEXT_PUBLIC_ANALYTICS_BASE_URL}/analytics.js`}
+            data-site-id={env.APP_TRACKING_SITE_ID}
+            data-server-url={`${process.env.NEXT_PUBLIC_TRACKING_SERVER_ENDPOINT}/track`}
```
Lindharden, line 45: These should use env as well instead of process.env

Lindharden (changes requested):
Nice. Few remarks :)

Lindharden (approved):
Looks good 👍

PR #335, opened 2025-07-11
Initial about page draft

### LINKED ISSUE #313: Create /about page
The about page needs to be created at /about. It's currently linked to in the footer.

The about page should contain an introduction to the concept of betterlytics. The team.

Consider the following:
- Explanation what makes us more private-friendly than competitors 
- Brief message on our GDPR compliance and how we strive to protect data etc.

Lindharden:
Closes #313


flopper123 (commented):
I generally like it very much!
A nice addition could be the fact that the data never leaves the EU, or even our own servers.

PR #336, opened 2025-07-11
324 dashboard basidebar not greyed out when usersettings integration setup dialog box is open

### LINKED ISSUE #323: Dashboard Geography Tab overlaps setting overlays
The base-map on the LeafletMap on the Geography Tab from the Dashboard overlaps with dialog and menu overlays (see screenshots below).

This is likely due to the z-indice on the Geography Tab not being set correctly, as it doesn't account for the LeafletMap having a default z-indice of 999 in the map library.

#### UI Screenshots
Image
Image

#### Suggested Solution to Geography Tab
To resolve the z-index on the geography tab, we can probably do something like this globally: 

```
  [role="dialog"], [role="menu"] {
    z-index: 1000 !important;
  }

  div[data-radix-popper-content-wrapper] {
    z-index: 1001 !important;
  }
```

It requires extensive testing to resolve this, as we need to ensure this change doesn't have unwanted side-effects.

### LINKED ISSUE #324: Dashboard BASideBar not greyed out when UserSettings / Integration setup dialog-box is open
The BASideBar should clearly be greyed out when the UserSettings or Integration modal is open, as it's not CLICKABLE while the modal is open. 

#### UI Screenshots of interaction between UserSettings and BASideBar
Image

#### UI Screenshots of interaction between UserSettings and Integration Setup
Image

### LINKED ISSUE #325: Zoom-controls should be greyed out when a dialog-box opens from the Geography Tab
The zoom-controls should be greyed out to indicate they are not clickable when a dialog-box / modal opens from the Geography Tab on the Dashboard (see screenshot).

#### UI Screenshot
Image

### LINKED ISSUE #329: Fix z-index issues regarding sidebar
# Z-INDEX
Note, we have z-index issues regarding the side-bar.
This is due to the geography having some high built in z-index.
This has caused multiple z-index issues, as on phone the sidebar needs to be on top of the geography page, hence why the sidebar has large z-index values. I've set the values as low as I can on the sidebar, so we probably need to specify larger z-index values on the places where issues arise.

## User settings
User settings on smaller screens is rendered behind sidebar.
Image

## Integration setup
Integration is behind sidebar on mobile:
Image

On desktop, the sidebar is not "grayed out". Fixing the integration sidebar z-index should fix this issue as well
Image

## Dashboard filters
On smaller screens the sidebar is on top of the dashboard filters, making the filters impossible to fill out
Image

Lindharden:
Closes #324 
Closes #323 
Closes #329 
Closes #325 

- Removed most z-indexes now that z-index hierarchy is fixed.
- I gave topbar z-50 to avoid the border not appearing on world map.
- I made the sidebar extend to full height of screen on mobile view.
- I made world-map fixed to avoid the scrolling issues. I can't find the fundamental issue that is causing overflow on all pages. Likely caused by sidebar and topbar somehow.


Woldbye (commented):
Nice to see multiple bugs resolved in one PR – great cleanup! 👏

That said, I’m a bit concerned about setting `z-index: 0 !important` on core Leaflet classes like .leaflet-pane and .leaflet-control. These layers are part of Leaflet’s built-in z-index hierarchy, which ensures correct stacking of map tiles, overlays, tooltips, and controls. Overriding them globally could introduce subtle bugs or break expected behavior -especially with things like popups, tooltips. It might even break leaflets own inteernal logic, as it most likely relies on the z-indices. 

I think it would be safer to instead raise the z-index of our modal/dialog layers rather than lowering the map’s internals? That might be a more targeted fix with fewer side effects.

Lindharden:
> I think it would be safer to instead raise the z-index of our modal/dialog layers rather than lowering the map’s internals? That might be a more targeted fix with fewer side effects.

This was previously done in attempt to fix this, because leaflet's map internally uses z-index 1000.

But this caused an avalanche of z-index issues, hence why I could re,remove explicitly defined z-indexes from all our existing components. It seems to conflict with most of shadcn's native components and logic. 

I found this Stackoverflow post that I followed - it seems to be a general problem that many have experienced! I tested everything tho, on all screen sizes, and everything seemed to work perfectly :)

https://stackoverflow.com/a/73083646

flopper123 (approved):
Tested it - everything seems to be fixed now. Very nice!

PR #352, opened 2025-07-13
Updated API for custom events in tracking script

### LINKED ISSUE #351: Update API for custom events in tracking script
Instead of users calling:
`window.betterlytics(...)`

It should be:
`window.betterlytics.event(...)`

This will allow us to handle better backwards compatibility if we want to add multiple method later on

flopper123:
Closes #351 

Instead of `betterlytics(...)`, you now call `betterlytics.event(...)` to track custom events

- [x] Docs updated
- [x] Added placeholder prior to NPM package
- [x] Tracked custom event on metric change in dashboard page as local test

image


dashboard/src/lib/betterlytics.ts
Lindharden, line 1: We should remove this and instead use npm package if we want to send custom events

dashboard/src/app/dashboard/[dashboardId]/(dashboard)/SummaryAndChartSection.tsx
```diff
@@ -30,6 +31,7 @@ export default function SummaryAndChartSection({ data }: SummaryAndChartSectionP
   const { dictionary } = useDictionary();
 
   const handleMetricChange = useCallback((metric: string) => {
+    betterlytics.event('dashboard-overview-metric-change', { metric });
```
Comment (Lindharden) [inline dashboard/src/app/dashboard/[dashboardId]/(dashboard)/SummaryAndChartSection.tsx:34]: we shouldn't send events here unless we put it behind a configurable environment variable

Lindharden (changes requested):

dashboard/src/app/dashboard/[dashboardId]/(dashboard)/SummaryAndChartSection.tsx
Comment (flopper123) [reply dashboard/src/app/dashboard/[dashboardId]/(dashboard)/SummaryAndChartSection.tsx]: Removed 👍

PR #356, opened 2025-07-13
User setting for avatar

flopper123:


dashboard/src/components/avatar/BAAvatar.tsx
```diff
@@ -0,0 +1,44 @@
+'use client';
+
+import { useUserSettings } from '@/hooks/useUserSettings';
+import { useMemo } from 'react';
+import { Avatar } from '../ui/avatar';
+import { User } from 'lucide-react';
+import { GravatarImage } from '../ui/gravatar-image';
```
Lindharden, line 7: Can we use relative import here? @

dashboard/src/components/avatar/BAAvatar.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { useSession } from 'next-auth/react';
+
+export function BAAvatar() {
+  const { settings } = useUserSettings();
+  const avatarMode = useMemo(() => settings?.avatar || 'default', [settings?.avatar]);
```
Lindharden, line 12: is 'default' fallback here required? Considering avatar should always be either Gravatar or Default

dashboard/src/components/avatar/BAAvatar.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const { data: session } = useSession();
+
+  if (!session) {
+    return <DefaultAvatar />;
+  }
+
+  return (
+    <>
+      <DefaultAvatar />
+      <GravatarImage email={session.user?.email} alt={session.user?.name || 'User'} className='bg-red absolute' />
```
Lindharden, line 41: Is it correct that this returns both Default and GravatarImage?

dashboard/src/components/userSettings/UserPreferencesSettings.tsx
```diff
@@ -68,6 +73,36 @@ export default function UserPreferencesSettings({ formData, onUpdate }: UserPref
...
+                    <BookUser className='h-4 w-4' />
+                    <span>Gravatar</span>
+                  </div>
+                </SelectItem>
+              </SelectContent>
+            </Select>
+          </div>
+          {settings?.avatar === 'gravatar' && (
+            <div className='text-muted-foreground text-xs'>
+              Note: Gravatar is a third-party library - by enabling it you agree to their terms and services
```
Lindharden, line 102:
```suggestion
Note: Gravatar is a third-party service — enabling it may share your email hash with their servers. By enabling, you agree to their terms.
```

What do you think about this? :) And perhaps even go as far as make terms a hyperlink to their terms of service

dashboard/src/components/avatar/BAAvatar.tsx
flopper123: Yeah, it was like this before. The `GravatarImage` fallsback to `DefaultAvatar` if not found.

Lindharden (approved):

PR #363, opened 2025-07-14
Add CSS to reposition scrollable-content below TopBar for Dashboard and Public layouts

Woldbye:
Add CSS to reposition scrollable-content below topbar on all Dashboard and Public layouts.

#### Changes
* Adjust overlay designs to make them correctly set overflow and size via. flex layout for all layouts on the dashboard and landing-page. 
* Adjust map size to fit the viewport, instead of overflowing. 
* Adjust map-controls as position: absolute, instead of sticky or fixed. This is a huge improvement for long term maintainability.
* Make `<Footer>` part of scrollable section on all pages.
* Add border shadow on topbar, this looks a lot better now that it's not transparent 

#### Notes
* The blur effect on the Topbars are gone - it's beyond me how to fix the layout while pertaining the blur-effects. I highly advice to skip them for now, as it offers very little to the UX, and its complicated to implement correctly.
* I believe we have an issue to connect to this, but I couldn't quite find it? Please help attach it, if you see it.
* I have not touched the `/docs` section, but I did confirm that it still works 

##### Refactor `/dashboard/src/ScrollReset.tsx`?
We may need to refactor /dashboard/src/ScrollReset.tsx to target specific scrollable containers, especially if there are nested or multiple scrollable sections on the page. The current implementation resets the scroll position using window.scrollTo(0, 0), which only works for the main document scroller.

If scrolling areas are nested or custom (e.g., not using `<main>` as the primary scroll container), each one would need to be handled individually.

That said, I couldn't reproduce the scroll reset bug described in the [Next.js issue](https://github.com/vercel/next.js/issues/45187), so I haven't made this change yet. My assumption is that Next.js may already be handling scroll restoration for inner scroll containers in our case.

### UI Screenshots from dashboard

I have manually checked the relevant pages, but I would like if reviewer helped me double check, as I have touched a lot of pages, and its hard to ensure it works on all browsers alone.

#### Desktop
image
image

image

#### Mobile
image
image

### DIFF
--- a/dashboard/src/app/dashboard/[dashboardId]/layout.tsx
+++ b/dashboard/src/app/dashboard/[dashboardId]/layout.tsx
@@ -49,12 +49,12 @@ export default async function DashboardLayout({ children, params }: DashboardLay
   return (
     <DashboardProvider>
       <DictionaryProvider dictionary={dictionary} initialLanguage={language}>
-        <section className="flex flex-col h-screen">
+        <section className='h-screen w-full'>
           <BATopbar />
           <SidebarProvider>
             <BASidebar dashboardId={dashboardId} />
             <BAMobileSidebarTrigger />
-            <main className='flex flex-col min-h-0 flex-1 bg-background w-full pt-[56px]'>
+            <main className='bg-background flex min-h-0 w-full flex-1 flex-col pt-[56px]'>
               <ScrollReset />
               {billingEnabled && (
                 <Suspense fallback={null}>
@@ -68,9 +68,7 @@ export default async function DashboardLayout({ children, params }: DashboardLay
                     <VerificationBanner email={session.user.email} userName={session.user.name || undefined} />
                   </div>
                 )}
-              <div className='scrollable-y'>
-                {children}
-              </div>
+              <div className='scrollable-y'>{children}</div>
             </main>
             {/* Conditionally render tracking script based on server-side feature flag */}
             {shouldEnableTracking && siteId && <TrackingScript siteId={siteId} />}
--- a/dashboard/src/app/dashboard/[dashboardId]/layout.tsx
+++ b/dashboard/src/app/dashboard/[dashboardId]/layout.tsx
@@ -49,12 +49,12 @@ export default async function DashboardLayout({ children, params }: DashboardLay
   return (
     <DashboardProvider>
       <DictionaryProvider dictionary={dictionary} initialLanguage={language}>
-        <section>
+        <section className="flex flex-col h-screen">
           <BATopbar />
           <SidebarProvider>
             <BASidebar dashboardId={dashboardId} />
             <BAMobileSidebarTrigger />
-            <main className='bg-background w-full overflow-x-hidden'>
+            <main className='flex flex-col min-h-0 flex-1 bg-background w-full overflow-x-hidden overflow-y-auto'>
               <ScrollReset />
               {billingEnabled && (
                 <Suspense fallback={null}>
@@ -68,7 +68,7 @@ export default async function DashboardLayout({ children, params }: DashboardLay
                     <VerificationBanner email={session.user.email} userName={session.user.name || undefined} />
                   </div>
                 )}
-              <div className='flex w-full justify-center'>{children}</div>
+              <div className='flex flex-1 overflow-y-auto min-h-0 w-full justify-center'>{children}</div>
             </main>
             {/* Conditionally render tracking script based on server-side feature flag */}
             {shouldEnableTracking && siteId && <TrackingScript siteId={siteId} />}
--- a/dashboard/src/app/dashboard/[dashboardId]/layout.tsx
+++ b/dashboard/src/app/dashboard/[dashboardId]/layout.tsx
@@ -49,12 +49,12 @@ export default async function DashboardLayout({ children, params }: DashboardLay
   return (
     <DashboardProvider>
       <DictionaryProvider dictionary={dictionary} initialLanguage={language}>
-        <section className='h-screen w-full'>
+        <section>
           <BATopbar />
           <SidebarProvider>
             <BASidebar dashboardId={dashboardId} />
             <BAMobileSidebarTrigger />
-            <main className='bg-background flex min-h-0 w-full flex-1 flex-col pt-[56px]'>
+            <main className='bg-background box-border flex w-full flex-col overflow-x-hidden'>
               <ScrollReset />
               {billingEnabled && (
                 <Suspense fallback={null}>
@@ -68,7 +68,8 @@ export default async function DashboardLayout({ children, params }: DashboardLay
                     <VerificationBanner email={session.user.email} userName={session.user.name || undefined} />
                   </div>
                 )}
-              <div className='scrollable-y'>{children}</div>
+
+              <div className='relative flex h-full w-full justify-center'>{children}</div>
             </main>
             {/* Conditionally render tracking script based on server-side feature flag */}
             {shouldEnableTracking && siteId && <TrackingScript siteId={siteId} />}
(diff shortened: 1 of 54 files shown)


dashboard/src/app/dashboard/[dashboardId]/layout.tsx
```diff
@@ -68,7 +68,9 @@ export default async function DashboardLayout({ children, params }: DashboardLay
                     <VerificationBanner email={session.user.email} userName={session.user.name || undefined} />
                   </div>
                 )}
-              <div className='flex w-full justify-center'>{children}</div>
+              <div className='scrollable-y'>
```
Comment (flopper123) [inline dashboard/src/app/dashboard/[dashboardId]/layout.tsx:71]: Can we still have the justify-center on it? Because when I test it, the main content is not aligned.
This is what it used to look like:
image

Now, the content looks like it's left-aligned:
image

dashboard/src/app/dashboard/[dashboardId]/layout.tsx
```diff
@@ -49,12 +49,12 @@ export default async function DashboardLayout({ children, params }: DashboardLay
   return (
     <DashboardProvider>
       <DictionaryProvider dictionary={dictionary} initialLanguage={language}>
-        <section>
+        <section className="flex flex-col h-screen">
           <BATopbar />
```
Comment (flopper123) [inline dashboard/src/app/dashboard/[dashboardId]/layout.tsx:53]: The top-bar is missing the border somehow, this is how it looked like before:
image

Whereas now it's missing and is replaced by some gradient:
image

dashboard/src/app/dashboard/[dashboardId]/layout.tsx
```diff
@@ -49,12 +49,12 @@ export default async function DashboardLayout({ children, params }: DashboardLay
   return (
     <DashboardProvider>
       <DictionaryProvider dictionary={dictionary} initialLanguage={language}>
-        <section>
+        <section className="flex flex-col h-screen">
```
Comment (flopper123) [inline dashboard/src/app/dashboard/[dashboardId]/layout.tsx:52]: The `flex flex-col` seems unnecessary. I just tried to remove it locally, and it seems to not change anything. I think the `h-screen` fixed the issue with this one 😄

dashboard/src/app/dashboard/[dashboardId]/layout.tsx
```diff
@@ -49,12 +49,12 @@ export default async function DashboardLayout({ children, params }: DashboardLay
...
           <BATopbar />
           <SidebarProvider>
             <BASidebar dashboardId={dashboardId} />
             <BAMobileSidebarTrigger />
-            <main className='bg-background w-full overflow-x-hidden'>
+            <main className='flex flex-col min-h-0 flex-1 bg-background w-full pt-[56px]'>
```
Comment (flopper123) [inline dashboard/src/app/dashboard/[dashboardId]/layout.tsx:57]: Same with this one regarding `flex flex-col` removing them locally did not seem to change anything 😁

dashboard/src/app/dashboards/layout.tsx
```diff
@@ -12,9 +12,11 @@ export default async function DashboardsLayout({ children }: DashboardsLayoutPro
 
   return (
     <DictionaryProvider dictionary={dictionary} initialLanguage={language}>
-      <section className='h-full w-full'>
+      <section className='flex flex-col h-screen w-full'>
```
flopper123, line 15: This `flex flex-col` can also be removed.

dashboard/src/components/topbar/BATopbar.tsx
```diff
@@ -31,7 +31,7 @@ export default function BATopbar() {
 
   return (
     <>
-      <header className='bg-background/95 supports-[backdrop-filter]:bg-background/60 sticky top-0 z-50 w-full border-b backdrop-blur'>
+      <header className='fixed top-0 left-0 z-50 bg-background/95 supports-[backdrop-filter]:bg-background/60 w-full backdrop-blur-md shadow-md shadow-black/5 dark:shadow-white/4'>
```
flopper123, line 34: This is the one regarding the border. I just switched to light mode, and it looks nice. However, on dark mode it should probably look differently, as it's hard to see 😄 
Readding the `border-b` looks nice on both light and dark mode, however, I would customize the shadow for the dark mode a bit

dashboard/src/app/dashboard/[dashboardId]/layout.tsx
Comment (Woldbye) [reply dashboard/src/app/dashboard/[dashboardId]/layout.tsx:52]: You could remove it, but I added it to ensure the layout is flex for predictability for future changes 👌👌

dashboard/src/app/dashboard/[dashboardId]/layout.tsx
Comment (Woldbye) [reply dashboard/src/app/dashboard/[dashboardId]/layout.tsx:71]: image
Here's a screenshot of the changes - Please note I have chosen weird dimensions, as my screen is not wide enough to see this error usually :D

dashboard/src/app/dashboard/[dashboardId]/layout.tsx
Comment (Woldbye) [reply dashboard/src/app/dashboard/[dashboardId]/layout.tsx:57]: You __cannot__ remove this `flex flex-col` - it's necessary to correctly place the scrollbar. If you remove it, it starts below the topbar.

dashboard/src/app/dashboards/layout.tsx
Woldbye: @flopper123 This is fine to remove - I have removed it

dashboard/src/components/topbar/BATopbar.tsx
Woldbye: I have adjusted the dark variant shadow and readded the border-b

image

Woldbye:
@flopper123 I have attempted to resolve the issues you mentioned, please have a look at your convenience! :)

Lindharden (changes requested):
I think you made some accidental changes to the layout. Here are some things for instance:

The frontpage now has enlarged space causing key-takeaways to be hidden:
{8B370E2D-F86D-4B4E-A3B8-18A9822FDDDF}

compared to
{5F4F6A76-E60E-47CC-99C1-DABDA15F2841}

Can be accentuated by zooming out, showcasing how the content is grouped nicely together while the hero section no longer plays nice with the rest.
chrome_6iVgzej2YV

Unless this was intended? 

The footer on the sign in page is also hidden far away:
{BC5F8265-848F-43F8-9242-F0F42B219DC2}

compared to:
{3C20C796-FB06-4755-9A61-EB872677C0E4}

Also seems like the breadcrumb section in the billing page no longer spans the entire width:
{B34CAE97-79BC-4E7F-9D0C-5F4C1AFE1C82}

Compared to:
{28992E59-C8CD-43F9-967D-AB19AF2941C5}


I'm a bit concerned the scrollbar no longer extends the entire container height, meaning you can't scroll when your mouse hovers the topbar. This is not common practice and it looks a bit odd that the scrollbar doesn't extend the page.

See facebook, github, microsoft etc as examples:
{32E18BBF-D78A-4971-B45F-DA557D2D94D1}

{1ACB31CE-6576-497F-98CF-287DB1EDA094}

{CD6E4602-D420-4027-A3F9-1A0A3D0671AF}

{9B142366-43C4-4C9F-AEA5-0C57928DED6C}

But other than that, the annoying scrolling on pages that doesn't require scrolling is gone, so that's quite nice!

dashboard/src/app/globals.css
```diff
@@ -172,3 +172,23 @@
 .leaflet-bottom {
   z-index: 0 !important;
 }
+
+html,
+body {
+  height: 100%;
+  margin: 0;
+  padding: 0;
+  overflow: hidden;
```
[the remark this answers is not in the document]
Woldbye: No major implications, on the contrary this makes it controllable. Without this you can't control inner scrolling sections as you want in modals. 

Individual scrolling such as in modals is handled by setting modal as flex view and applying flex shrink / flex grow appropriately.

PR #369, opened 2025-07-16
Setup prod logging

flopper123:


Lindharden (approved):

Lindharden:
Before you merge, please test if the dashboards are importable and can properly find the datasource :)

docker-compose.yml
```diff
@@ -70,6 +70,27 @@ services:
     networks:
       - monitoring_network
 
+  promtail:
+    image: grafana/promtail:latest
```
Lindharden, line 74:
```suggestion
    image: grafana/promtail:3.2.1
```

Promtail is set to be deprecated, so we should ensure we don't use latest tag here :)

promtail/promtail-config.yml
```diff
@@ -0,0 +1,20 @@
+server:
+  http_listen_port: 9080
+
+positions:
+  filename: /tmp/positions.yaml
```
Lindharden, line 5: When writing to tmp, aren't the positions lost on container restarts unless we add it to a persistent volume?

docker-compose.yml
```diff
@@ -70,6 +70,27 @@ services:
...
+    restart: unless-stopped
+    volumes:
+      - /var/lib/docker/containers:/var/lib/docker/containers:ro
+      - /var/run/docker.sock:/var/run/docker.sock:ro
+      - ./promtail/promtail-config.yml:/etc/promtail/config.yml:ro
+    networks:
+      - monitoring_network
+
+  loki:
+    image: grafana/loki:latest
```
Lindharden, line 86: Perhaps also change to a persistent version here instead of latest. Consider seeing what latest evaluates to and specify that version as we know it's compatible

Lindharden (changes requested):

Lindharden (approved):

PR #372, opened 2025-07-17
Added default filters fallback to invalid filter state

### LINKED ISSUE #361: Handle invalid filter url
Deleting half of the filter URL results in application error.

You can reproduce by changing your url to any invalid state, such as:

/dashboard/cmd3cgm8s0009r908xo6t722q?filters=asd

The error should be handled and the filters should fallback to default filters

Lindharden:
Fixes application error for invalid filter states

Closes #361


flopper123 (approved):
Works like a charm 😄

PR #373, opened 2025-07-17
22 add rate limiting to server endpoints

### LINKED ISSUE #22: Add rate-limiting to server endpoints
We should find a decent lenient rate limit for events so that the tracking endpoint can not be abused.

Lindharden:
Implements rate-limit (Up to 30 requests immediately, then 10 requests per second thereafter).

Implements event validation that limits length of properties, validates correct input such as proper IP & URL, existing user agent & event name, JSON-parsable payload for custom events, control characters, timestamps are not older than 5 minutes etc.

Added metric logs and output so that we can monitor reason for reject events, how many exceeds rate limits and the time it takes to validate input.

Example log output for promtail:
2025-07-17T14:15:09.930807Z  WARN betterlytics::validation: Event rejected by validation rejection_reason="invalid_user_agent" site_id=test-md781oez event_name=product-clicked ip_hash=12ca17b49af22894 url_domain=localhost url_path=/dashboard user_agent_prefix= properties_size=23
2025-07-17T14:15:09.930819Z  WARN betterlytics::validation: Event rejected by validation rejection_reason="invalid_user_agent" site_id=test-md781oez event_name=pageview ip_hash=12ca17b49af22894 url_domain=localhost url_path=/dashboard user_agent_prefix= properties_size=2
2025-07-17T14:15:09.930846Z  WARN betterlytics::validation: Event rejected by validation rejection_reason="invalid_user_agent" site_id=test-md781oez event_name=pageview ip_hash=12ca17b49af22894 url_domain=localhost url_path=/dashboard user_agent_prefix= properties_size=2
2025-07-17T14:15:09.930884Z  WARN betterlytics::validation: Event rejected by validation rejection_reason="invalid_user_agent" site_id=test-md781oez event_name=pageview ip_hash=12ca17b49af22894 url_domain=localhost url_path=/dashboard user_agent_prefix= properties_size=2

Closes #22

### DIFF
--- a/backend/src/validation/mod.rs
+++ b/backend/src/validation/mod.rs
@@ -4,6 +4,8 @@ use std::net::IpAddr;
 use std::str::FromStr;
 use url::Url;
 use crate::analytics::RawTrackingEvent;
+use sha2::{Digest, Sha256};
+use tracing::warn;
 
 #[derive(Debug, Clone)]
 pub struct ValidationConfig {
@@ -68,9 +70,24 @@ impl EventValidator {
         raw_event: RawTrackingEvent,
         ip_address: String,
     ) -> Result<ValidatedTrackingEvent, ValidationError> {
-        self.validate_required_fields(&raw_event)?;
-        self.validate_payload_sizes(&raw_event)?;
-        self.validate_formats(&raw_event, &ip_address)?;
+        let result = self.validate_event_internal(&raw_event, &ip_address);
+        
+        // Log sanitized rejection details if validation fails
+        if let Err(ref error) = result {
+            self.log_sanitized_rejection(error, &raw_event, &ip_address);
+        }
+        
+        result
+    }
+
+    fn validate_event_internal(
+        &self,
+        raw_event: &RawTrackingEvent,
+        ip_address: &str,
+    ) -> Result<ValidatedTrackingEvent, ValidationError> {
+        self.validate_required_fields(raw_event)?;
+        self.validate_payload_sizes(raw_event)?;
+        self.validate_formats(raw_event, ip_address)?;
 
         // only present for custom events
         if !raw_event.properties.is_empty() {
@@ -78,8 +95,8 @@ impl EventValidator {
         }
 
         Ok(ValidatedTrackingEvent {
-            raw: raw_event,
-            ip_address,
+            raw: raw_event.clone(),
+            ip_address: ip_address.to_string(),
         })
     }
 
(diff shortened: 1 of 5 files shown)


backend/src/validation/mod.rs
```diff
@@ -0,0 +1,290 @@
...
+            rejection_reason = self.get_rejection_reason(error),
+            site_id = %raw_event.site_id,
+            event_name = %raw_event.event_name,
+            ip_hash = %ip_hash,
+            url_domain = %self.extract_domain(&raw_event.url),
+            url_path = %self.extract_path(&sanitized_url),
+            user_agent_prefix = %sanitized_user_agent,
+            properties_size = %raw_event.properties.len(),
+            "Event rejected by validation"
+        );
```
flopper123, line 226: This logs the actual control characters, which I don't think we want.

For instance, sending `window.betterlytics.event("hel\n\nlo\t\tworld")` will cause it to print:

image

flopper123 (changes requested):
Lovely! 😄 
Only found the one thing that allows control characters to be printed to console.

One slight concern - I wasn't really able to test whether the rate limit works by IP properly. Has this been tested, that it rate limits by IP? Because else someone can DoS us very easily 🤣

flopper123:
> > One slight concern - I wasn't really able to test whether the rate limit works by IP properly. Has this been tested, that it rate limits by IP? Because else someone can DoS us very easily 🤣
> 
> It is per IP. I confirmed that it works with my own IP, naturally couldn't test multiple IPs at once. You can read more here :)
> 
> https://github.com/benwis/tower-governor

Yeah, but the IP from the header is from from the NGINX reverse proxy. That's why the `parse_id` uses the custom `x-forwarded-for` header, which NGINX sets. So if Governor uses the actual header's IP, then this will rate limit all users at once 😄

Lindharden:
> > > One slight concern - I wasn't really able to test whether the rate limit works by IP properly. Has this been tested, that it rate limits by IP? Because else someone can DoS us very easily 🤣
> > 
> > 
> > It is per IP. I confirmed that it works with my own IP, naturally couldn't test multiple IPs at once. You can read more here :)
> > https://github.com/benwis/tower-governor
> 
> Yeah, but the IP from the header is from from the NGINX reverse proxy. That's why the `parse_id` uses the custom `x-forwarded-for` header, which NGINX sets. So if Governor uses the actual header's IP, then this will rate limit all users at once 😄

Good catch. Fixed

backend/src/validation/mod.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    /// Log sanitized rejection details for debugging
+    fn log_sanitized_rejection(
+        &self,
+        error: &ValidationError,
+        raw_event: &RawTrackingEvent,
+        ip_address: &str,
+    ) {
+        // Hash IP address for privacy
+        let ip_hash = self.hash_ip_address(ip_address);
```
Woldbye, line 208: Why print the ip address if you obscure hash it anyway?

Lindharden: Yeah the idea is that we don’t need the Ip for any distinct purposes in terms of telling why a request was invalidated, but it provides traceability in terms of telling if it’s a single user that sends malicious events that gets rejected, or if there are multiple users that gets their requests invalidated (indicative that there exists some user agent or event payload that we don’t properly handle). 

But it may not be useful at all. We can always remove in the future if we never actually use it

flopper123 (approved):

PR #377, opened 2025-07-18
Update `simulate-events.js` to simulate users from all over the world

### LINKED ISSUE #370: Extend `simulate-events.js` to generate random GeoIps
It would improve our development experience greatly to be able to simulate random GeoIps in the `simulate-event.js` script. 

This would enable us to quickly test changes to the geography tabs.

Woldbye:
Closes #370 

Update `simulate-events.js` script to simulate random global GeoIP's for users all around the world.

The resulting geographic distribution is realistic - countries with large public IP allocations (e.g., US, Russia, China) receive more simulated traffic. Sparse regions like Greenland or Antarctica may appear as zero, not due to filtering, but simply due to their minimal public IP space.

## Changes
* Update Rust GeoIP parsing to check for `enable_geolocation` flag before defaulting `127.0.0.1` to `localhost`.
* Assign a random public IP address per user, ensuring consistent GeoIP for each simulated user. The public IP address is determined by exclude all private/reserved CIDRs based on [IANA IPv4 Special Registry](https://www.iana.org/assignments/iana-ipv4-special-registry) 
* Add support for custom CLI flags for simulation parameters to `simulate-events.js`:
```
--events (total number of events)
--users (number of unique users)
--batch-size (requests per batch)
--event-freq (custom event ratio)
```

* Updated CONTRIBUTING.md with detailed usage documentation and examples

## Additional Notes
* Our rust backend uses a local MaxMind GeoLite2 DB to resolve IP → country, so it's not a concern to lookup this many posts
* The timestamp validation changes from #22 has broken our `simulate-events.js` days functionality, so I have defaulted it to use `Date.now()` to circumvent the validation. 
* Simulating more than ~10,000 events may trigger ingestion usage limits on the local analytics server
image

## Screenshots of result of large simulation
image

image


.gitignore
```diff
@@ -36,6 +36,7 @@
 # System Files
 .DS_Store
 Thumbs.db
+/backend/assets/geoip/GeoLite2-Country.mmdb
```
Woldbye, line 39: Remove as this have been added to main 
```suggestion
```

backend/src/geoip/mod.rs
```diff
@@ -116,7 +119,8 @@ impl GeoIpService {
     }
 
     pub fn lookup_country_code(&self, ip_address: &str) -> Option<String> {
-        if ip_address == "127.0.0.1" || ip_address == "::1" {
+        if !self.config.enable_geolocation && (ip_address == "127.0.0.1" || ip_address == "::1") {
```
Lindharden, line 122: What is the resulting country when enable_geolocation is true and ip address corresponds to localhost? Does it just return Unknown now instead of Localhost cause the IP doesn't exist in MaxMind DB? In which case, don't we want it to return localhost regardless of config for geolocation?

Lindharden (commented):
This is too complex to review on mobile abroad, and this is the script @flopper123 created - so I'll let him review it 🤣

backend/src/geoip/mod.rs
Woldbye: I found out that this wasn't necessary, and have therefore reverted all changes to the backend

flopper123 (approved):
Works super nice!

PR #379, opened 2025-07-19
Allow the user to setup and login using a TOTP

RaMin0:
# Setup TOTP
https://github.com/user-attachments/assets/a8f8ebb5-090f-450d-a936-ebd8d15e22d3

# Login with TOTP
https://github.com/user-attachments/assets/0dcb5e77-3c03-4b86-9ab6-e8e873c091c9

# Disable TOTP
https://github.com/user-attachments/assets/fb96adb8-d0c3-422a-87a7-1164c1ad767a


dashboard/src/components/auth/LoginForm.tsx
```diff
@@ -70,20 +111,48 @@ export default function LoginForm() {
...
+              ref={totpInputRef}
+              autoSubmit
+              form='login'
+            />
+          )}
+          <AlertDialogFooter>
+            <AlertDialogCancel disabled={isPending}>Cancel</AlertDialogCancel>
+          </AlertDialogFooter>
+        </AlertDialogContent>
+      </AlertDialog>
```
flopper123, line 154: Suggestion: It works super nice! However, I would move this into a separate page, such that you get redirected from the `/signin` page to `/mfa` for instance. That would both help simplifying this page and flow, optimize for accessibility, and show the correct "Save password" at the correct timing. Currently, my password manager only asks to save the password after the TOTP has been filled in😄

dashboard/src/lib/crypto.ts
```diff
@@ -0,0 +1,38 @@
...
+  const iv = Buffer.from(ivHex, OUTPUT_ENCODING);
+  const authTag = Buffer.from(authTagHex, OUTPUT_ENCODING);
+  const keyBuffer = getKey(key);
+
+  const decipher = crypto.createDecipheriv(ALGORITHM, keyBuffer, iv);
+  decipher.setAuthTag(authTag);
+
+  const decrypted = decipher.update(encrypted, OUTPUT_ENCODING, INPUT_ENCODING) + decipher.final(INPUT_ENCODING);
+  return decrypted;
+}
```
flopper123, line 32: 😎 🔥

dashboard/src/repositories/postgres/user.ts
```diff
@@ -12,18 +13,24 @@ import {
...
 
+export async function findUserById(userId: string): Promise<User | null> {
+  return await findUserBy({ id: userId });
+}
+
 export async function findUserByEmail(email: string): Promise<User | null> {
+  return await findUserBy({ email });
+}
+
+async function findUserBy(where: Prisma.UserWhereUniqueInput): Promise<User | null> {
```
flopper123, line 24: This is neat! 🤩

flopper123 (changes requested):
Very nice addition! Works super well! :)

dashboard/prisma/schema.prisma
```diff
@@ -49,6 +49,8 @@ model User {
   sessions      Session[]
 
   passwordHash String?
+  totpEnabled Boolean @default(false)
```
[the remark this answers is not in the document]
RaMin0: Thought about it, but then `totpSecret` would be here and `totpEnabled` would be there, and it felt like it's cognitively simpler to have them both live in the same place. Wdyt?

dashboard/src/components/auth/LoginForm.tsx
RaMin0: Given how this is handled via `next-auth`, the `email`, `password` and `totp` have to be passed together to `verifyCredentials`, and it seemed challenging to achieve this when redirecting the user to another page. I'd have to save the password somewhere as plain text to resubmit it, or add some logic to the session to be in a state mid-way between not being signed in and being fully signed in

Imho, the password manager's current behavior makes sense, because it should prompt to save the credentials after a successful login attempt, i.e. after entering a valid TOTP. Imagine a scenario where a hacker manages to figure out a user's credentials, but fails to provide a valid TOTP - why would they save the credentials in this case? 🤔

flopper123 (approved):

dashboard/prisma/schema.prisma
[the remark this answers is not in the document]
Lindharden: Seems fine how it's currently done 😀

PR #380, opened 2025-07-19
Save Changes button should be disabled when there are no changes to save

### LINKED ISSUE #330: Save Changes button should be disabled when there are no changes to save
This is the case for both Dashboard Settings and User settings:

Image

Image

RaMin0:
# User Profile
https://github.com/user-attachments/assets/8d15464b-a410-41d3-a28d-973f74299f78

# User Preferences
https://github.com/user-attachments/assets/59c58c00-b3bd-4a70-949c-d48f4915ed0b

# Dashboard Settings
https://github.com/user-attachments/assets/07ccbd67-80f6-484a-8171-0dec9e0b3cc4

Closes #330

### DIFF
--- a/dashboard/src/hooks/use-is-changed.ts
+++ b/dashboard/src/hooks/use-is-changed.ts
@@ -1,21 +1,15 @@
-import { useEffect, useState } from 'react';
+import { useMemo } from 'react';
 
 export default function useIsChanged<T extends Record<string, any>>(
   currentData: T | null | undefined,
   originalData: T | null | undefined,
 ): boolean {
-  const [isChanged, setIsChanged] = useState(false);
-
-  useEffect(() => {
+  return useMemo(() => {
     if (!originalData || !currentData) {
-      setIsChanged(false);
-      return;
+      return false;
     }
 
     const keys = Object.keys(currentData) as (keyof T)[];
-    const changes = keys.some((key) => currentData[key] !== originalData[key]);
-    setIsChanged(changes);
+    return keys.some((key) => currentData[key] !== originalData[key]);
   }, [currentData, originalData]);
-
-  return isChanged;
 }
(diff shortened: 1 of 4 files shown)


Woldbye:
Looks good to me :) nice work

dashboard/src/hooks/use-is-changed.ts
```diff
@@ -0,0 +1,21 @@
...
+      setIsChanged(false);
+      return;
+    }
+
+    const keys = Object.keys(currentData) as (keyof T)[];
+    const changes = keys.some((key) => currentData[key] !== originalData[key]);
+    setIsChanged(changes);
+  }, [currentData, originalData]);
+
+  return isChanged;
```
flopper123, line 20: Nitpick: This can be simplified by the use of `useMemo` 😄 
```suggestion
  return useMemo(() => {
    if (!originalData || !currentData) {
      return false;
    }

    const keys = Object.keys(currentData) as (keyof T)[];
    return keys.some((key) => currentData[key] !== originalData[key]);
  }, [currentData, originalData]);
```

dashboard/src/hooks/use-is-changed.ts
```diff
@@ -0,0 +1,21 @@
...
+}
```
flopper123, line 21: Nitpick: This can be simplified using `useMemo` 😄 Not a blocker by any means though!

```suggestion
export default function useIsChanged<T extends Record<string, any>>(
  currentData: T | null | undefined,
  originalData: T | null | undefined,
): boolean {
  return useMemo(() => {
    if (!originalData || !currentData) {
      return false;
    }
    const keys = Object.keys(currentData) as (keyof T)[];
    return keys.some((key) => currentData[key] !== originalData[key]);
  }, [currentData, originalData]);
}
```

flopper123 (approved):

dashboard/src/hooks/use-is-changed.ts
RaMin0: Done in f6d14c38007b81ca115d311dfd6fc1608849919e 👌🏻

PR #383, opened 2025-07-20
Fix Gravatar hint margin and reflect avatar setting changes upon save

### LINKED ISSUE #382: Gravatar hint spacing and Avatar setting instant feedback
On small desktop screens, the hint for the usage of Gravatar in the User Settings is not properly spaced from surrounding elements and changes to the Avatar setting are not reflected upon save, but instead require a full page reload

https://github.com/user-attachments/assets/f22a44a4-2bce-437d-9828-90cb94ec1f32

RaMin0:
https://github.com/user-attachments/assets/4bdd2754-2da2-4255-8f5c-492fed37cdc8

Closes #382


Lindharden (approved):
Nice improvement

dashboard/src/components/userSettings/UserPreferencesSettings.tsx
```diff
@@ -98,7 +98,7 @@ export default function UserPreferencesSettings({ formData, onUpdate }: UserPref
             </Select>
           </div>
           {settings?.avatar === 'gravatar' && (
-            <div className='text-muted-foreground text-xs'>
+            <div className='text-muted-foreground me-34 text-xs'>
               Note: Gravatar is a third-party service - enabling it may share your email hash with their servers.
               <br />
```
RaMin0, line 103: I think this looks better without the new line

image

Woldbye:
It seems like the “share” hint should be an approval dialog rather than just static text.
* Once the text is displayed, the data may have already been shared with Gravatar, making the message ineffective as a preventative measure.
* Visually, the text feels disconnected from the rest of the settings UI and disrupts the overall flow.

dashboard/src/components/userSettings/UserPreferencesSettings.tsx
Woldbye: Agreed! 😃

flopper123: Hey! 😄 
A suggestion to this is to use the tailwind `text-balance` class.
I would add the following classes to the parent and remove the br: `pt-2 text-balance md:pt-0`

That would make it fully responsive, and automatically handle the text breaks:
image

image

flopper123 (changes requested):

dashboard/src/components/userSettings/UserPreferencesSettings.tsx
```diff
@@ -98,7 +98,7 @@ export default function UserPreferencesSettings({ formData, onUpdate }: UserPref
             </Select>
           </div>
           {settings?.avatar === 'gravatar' && (
-            <div className='text-muted-foreground text-xs'>
+            <div className='text-muted-foreground me-34 text-xs'>
```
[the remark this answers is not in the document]
RaMin0: I tried using `flex-col` first, but I noticed the label jumps up when I choose Gravatar and the hint shows, which I didn't really like.

The reason I used the fixed `me-34` is to counteract the dropdown's fixed `w-32` and `py-2`. I tested it on different screen sizes and it behaved the same

dashboard/src/components/userSettings/UserPreferencesSettings.tsx
Woldbye: Thank you for the thorough testing @RaMin0 ! 

1. I think it's fine that the that the hint text flows underneath the dropdown menu for these screen sizes as long as we have the extra padding.
2. text-pretty seems like the more semantic approach - I tried both in my browser on the element, and couldn't spot any noticeable difference on edge between the two.

RaMin0: <table>
<thead>
<tr>
<th><code>default</code></th>
<th><code>text-balance</code></th>
<th><code>text-pretty</code></th>
</tr>
</thead>
<tbody>
<tr valign="top">
<td>image<br />Breaks sentences once the element's boundaries are reached</td>
<td>image<br />Tries to fill all lines of the paragraph equally to maintain a block feel</td>
<td>image<br />Flows the text naturally and tries to keep sentences together on the same line</td>
</tr>
</tbody>
</table>

RaMin0: @Woldbye Done in 9e5b397055af0ecddeed80e78f2feb4304f9693d

flopper123:
Very nice!

PR #387, opened 2025-07-22
Resolve hydration issue on Dashboard by lazy-rendering Date

### LINKED ISSUE #338: Runtime Error: Hydration failed because the server rendered HTML didn't match the client
Safari 18.5 on macOS 15.5

Image

Woldbye:
Closes #338 
Resolve the issues by lazy-rendering the Date on dashboard.

### Changes
* Add `useFormattedDate`-hook to delay rendering of date-string until client side. All future components that needs to display a 'date' to the client, should use this hook to avoid hydration errors. 
* Add Localization to Dashboard main page - The only text that should be localized seems to be the the 'created at date'.

### Notes
The error only occurs if the server-side component has knowledge of the actual value being printed client-side. So only components that print server-side rendered values such as `dashboard.created`, will need to utilize `useFormattedDate`.

### Image of Loading
Please note that it loads for a while, since I am simulating slow 4G network for the screen capture.

#### Dark Mode Video
msedge_IvABdDJp5P

#### Light Mode Video
msedge_kV7tpRu5r4


dashboard/src/components/LazyDate.tsx
```diff
@@ -0,0 +1,24 @@
...
+}) {
+  const { dictionary } = useDictionary();
+  const [formattedDate, setFormattedDate] = useState<string>('');
+
+  // Not using useMemo here, because we need to delay the formatting until after the component has mounted
+  useEffect(() => {
+    setFormattedDate(date ? date.toLocaleDateString(locales, options) : dictionary.misc.unknown);
+  }, [date]);
+
+  return <>{formattedDate || dictionary.misc.loading}</>;
```
flopper123, line 23: Suggestion: Set the loading state as the default state, and let the formatted date / unknown override that state 😄 

```suggestion
  const [formattedDate, setFormattedDate] = useState<string>(dictionary.misc.loading);

  // Not using useMemo here, because we need to delay the formatting until after the component has mounted
  useEffect(() => {
    setFormattedDate(date ? date.toLocaleDateString(locales, options) : dictionary.misc.unknown);
  }, [date]);

  return formattedDate;
```

dashboard/src/app/dashboards/DashboardCard.tsx
```diff
@@ -52,7 +47,13 @@ export default function DashboardCard({ dashboard }: DashboardCardProps) {
               <div className='flex w-9 justify-center'>
                 <Calendar className='text-muted-foreground h-3 w-3' />
               </div>
-              <span className='text-muted-foreground text-xs'>Created {formattedDate}</span>
+              <span className='text-muted-foreground text-xs'>
+                {dictionary.misc.createdAt}{' '}
+                <LazyDate
+                  date={dashboard.createdAt}
+                  options={{ year: 'numeric', month: 'short', day: 'numeric' }}
+                />
```
flopper123, line 55: It looks a little off that it prints "Created Loading..." while it's loading. I could see this issue also arising in the future, when using this.
Could the `LazyDate` be used as a "formatter" maybe? I'm thinking something along the lines of:
```tsx
<LazyDate
  formatted={(date) => `${dictionary.misc.createdAt} ${date}`}
  date={dashboard.createdAt}
  options={{ year: 'numeric', month: 'short', day: 'numeric' }}
/>
```
That way, it would show `Loading...` while loading, and then show `Created Jul 23, 2025` once the date has been loaded. Thoughts? 😁

flopper123 (changes requested):

dashboard/src/app/dashboards/DashboardCard.tsx
Woldbye: I agree that the loading looks a bit odd. I just copies the current implementation (which already prints Created Loading...) as this a bug related pr not enhancement. 

That being said, I think it's a great suggestion! Maybe the format should be optional, and default to just display the date

dashboard/src/components/LazyDate.tsx
Woldbye: Good suggestion - cleans it up nicely. I have implemented this!

dashboard/src/components/LazyDate.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { interpolateReactTemplate } from '@/utils/reactTemplateStrings';
+import { useEffect, useMemo, useState, ReactNode } from 'react';
+
+type LazyDateProps = {
+  date?: Date;
+  locales?: Intl.LocalesArgument;
+  options?: Intl.DateTimeFormatOptions;
+  formatted?:
+    | ((formattedDateString: string) => React.ReactNode)
+    | [(formattedDateString: string) => string, Record<string, React.ReactNode>];
```
flopper123, line 12: Suggestion: I would leave it as this:
```suggestion
  formatted?: (formattedDateString: string) => React.ReactNode;
```

We can always extend it later, if needed

flopper123 (approved):

PR #388, opened 2025-07-24
Added Google and Github social logins

Devleed:
**Funtionality Changes:**
Added Google and Github adapters in Next Auth configuration

**ENV Changes**
4 new env variables are introduced

GITHUB_ID= /* Github client ID */
GITHUB_SECRET= /* Github secret */
GOOGLE_CLIENT_ID= /* Google client ID */
GOOGLE_CLIENT_SECRET= /* Google client secret */

**UI Changes**
Added Social login buttons in login form

image

New UI after changes:

image

### DIFF
--- a/dashboard/src/components/auth/LoginForm.tsx
+++ b/dashboard/src/components/auth/LoginForm.tsx
@@ -15,6 +15,7 @@ import {
 } from '@/components/ui/alert-dialog';
 import { Spinner } from '@/components/ui/spinner';
 import { useIsMobile } from '@/hooks/use-mobile';
+import { getProviders } from 'next-auth/react';
 
 export default function LoginForm() {
   const router = useRouter();
@@ -25,10 +26,16 @@ export default function LoginForm() {
   const [totp, setTotp] = useState('');
   const [error, setError] = useState('');
   const [isDialogOpen, setIsDialogOpen] = useState(false);
+  const [providers, setProviders] = useState<Record<string, any> | null>(null);
+
   const [isPending, startTransition] = useTransition();
 
   const [isGooglePending, startGoogleTransition] = useTransition();
   const [isGithubPending, startGithubTransition] = useTransition();
+
+  useEffect(() => {
+    getProviders().then(setProviders);
+  }, []);
 
   useEffect(() => {
     if (isMobile) return;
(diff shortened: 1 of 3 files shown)


dashboard/src/components/auth/LoginForm.tsx
```diff
@@ -126,6 +149,45 @@ export default function LoginForm() {
...
+
+      <div className='space-y-3'>
+        {/* Google Login Button */}
+        <button
+          type='button'
+          onClick={() => handleOAuthLogin('google')}
+          className='flex w-full cursor-pointer items-center justify-center gap-3 border border-gray-400 bg-transparent px-6 py-2 text-sm font-medium text-white focus:ring-2 focus:ring-gray-500 focus:outline-none'
+        >
+          {/* Google G logo */}
+          <img width='24' height='24' src='https://img.icons8.com/fluency/48/google-logo.png' alt='google-logo' />
```
Lindharden, line 172: I believe the Google icon may be available as svg here: https://developers.google.com/identity/branding-guidelines

But it needs to use the correct one according to dark/light theme it seems (when skimming the guidelines)!

Can you look into if this is a better option, or if we should stick with the external url that you use here? @Devleed

dashboard/src/components/auth/LoginForm.tsx
Lindharden, line 1: Can we conditionally hide the OAuth button(s) if configuration keys are not set? I.e if GitHub client id and secret is not set, then it shouldn't render the GitHub login button 😄

Lindharden (dismissed):
I left a few minor comments! Amazing work @Devleed !!

dashboard/src/components/auth/LoginForm.tsx
Devleed: Sure

dashboard/src/lib/auth.ts
```diff
@@ -36,6 +38,21 @@ export const authOptions: NextAuthOptions = {
...
+      clientId: process.env.GOOGLE_CLIENT_ID!,
+      clientSecret: process.env.GOOGLE_CLIENT_SECRET!,
+      authorization: {
+        params: {
+          prompt: 'consent',
+          access_type: 'offline',
+          response_type: 'code',
+        },
+      },
+    }),
```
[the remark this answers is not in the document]
Devleed: Sure

Lindharden (approved):
I can't run it locally to check if it all works because I'm abroad, but looks good!! @flopper123 should have tested it 😄 

Thank you so much for your contribution @Devleed !! 😍

PR #389, opened 2025-07-24
Strips "www." postfix and trailing "/", if present

flopper123:
The referrer `www.betterlytics.io/dashboards/` would now be stored as `betterlytics.io/dashboards`


backend/src/referrer/mod.rs
```diff
@@ -229,4 +229,20 @@ fn extract_search_term(url: &Url, param_names: &[String]) -> Option<String> {
...
+
+/// Strips protocol prefixes (http://, https://, www.) and trailing /
+fn normalize_url(url: &String) -> String {
+    let stripped_prefix_url = strip_protocol_prefixes(url);
+    strip_url_postfix_forwardslash(&stripped_prefix_url)
+}
+
+/// Strips protocol prefixes (http://, https://, www.) and trailing /
+fn strip_protocol_prefixes(url: &String) -> String {
+    url.replace("https://", "").replace("http://", "").replace("www.", "")
```
Woldbye, line 242: I know the previous code looked like this, but aren't we missing a ton of protocols here?? data: ftp: etc.

flopper123: Practically speaking no, these won't be put on the `referrer`, as that's the "link" to the website you just came from 😄

Woldbye: I want you to add the note :D also the phrasing of each sentence is different, it's more explicit in the text here what is actually done

Lindharden (changes requested):
I think we should utilize the URL library like it's utilized elsewhere, to ensure robust handling.

See this generated code snippet as a rough example for inspiration:

use url::Url;

fn normalize_url(url_str: &str) -> Option<String> {
    if let Ok(mut url) = Url::parse(url_str) {
        // Remove the scheme (e.g., http, https)
        url.set_scheme("").ok()?;
        // Remove "www." prefix
        let host = url.host_str()?.trim_start_matches("www.");
        url.set_host(Some(host)).ok()?;
        // Remove trailing slash
        let normalized = url.as_str().trim_end_matches('/').to_string();
        Some(normalized)
    } else {
        None // Invalid URL
    }
}

Woldbye (dismissed):

Lindharden (approved):
Approved, but I cannot test locally as I'm abroad without a PC. Please thoroughly test with many different cases before merging to ensure it works!😎

PR #398, opened 2025-07-27
Style adjustments for Flag icons, LeafletMap tooltips and LeafletMap legend

Woldbye:
Addressing some small style misalignments introduced by my previous PRs. 

### Changes
* The Flag icons are a little bit too small on all devices, and their borders are too thick and overwhelming.
   * Make flag-icons slightly larger
   * Add opacity to flag-borders
   * FlagIcon fallback icon are now colored by the Theme

* Public page's Map's legend-text is too small - xss is simply unreadable, so I have removed it
* Leaflet Map tooltips are wrapping 'Visitors: x' on two lines on some countries for public page
   * I have set max-width of tooltips to a fixed width, as 50vw is simply too large to play nicely if its not fullscreen
* The colors on the Leaflet Map's tooltips now follow the theme (dark, light etc.)

### Screenshots
image

image

image
image

image

image

image
image


Lindharden (approved):
Changes looks good!

Does the tooltip also look nice on countries with long names? 😄

PR #400, opened 2025-07-28
Sets up user account properly on NextAuth provider sign-in

flopper123:


dashboard/prisma/schema.prisma
```diff
@@ -13,17 +13,21 @@ datasource db {
...
-  refreshToken       String?
-  accessToken        String?
-  accessTokenExpires DateTime?
+  refresh_token      String?
+  access_token       String?
+  expires_at         Int?
+  token_type         String?
+  scope              String?
+  id_token           String?
+  session_state      String?
```
RaMin0, line 25: Should we be mixing letter casing? `userId` vs. `user_id`

dashboard/prisma/schema.prisma
```diff
@@ -13,17 +13,21 @@ datasource db {
...
+  provider           String
   providerAccountId  String
-  refreshToken       String?
-  accessToken        String?
-  accessTokenExpires DateTime?
+  refresh_token      String?
+  access_token       String?
+  expires_at         Int?
+  token_type         String?
+  scope              String?
```
RaMin0, line 23: Should these be prefixed with `token` to indicate they're related to the `access_token` and not the whole `Account` record?

dashboard/prisma/schema.prisma
flopper123: Yes, this model is used by the Prisma adapter. This model has changed from version 3 to version 4.
We had accidentally set it up originally using the version 3 model, which then failed when Oauth was setup.
I've taken this from https://authjs.dev/getting-started/adapters/prisma#schema > PostgreSQL
So the model needs to look like this, for it to work! But nice catch nonetheless! 😄

RaMin0: It can be done using a custom Prisma adapter that maps the fields Auth.js expects to their camelCase counterparts, but this can be done in a separate PR as an enhancement 👌🏻

For the time being, maybe [this](https://authjs.dev/getting-started/adapters/prisma#naming-conventions) can clean things up a bit?

flopper123: Update - this sadly won't work with `@map` 😭 
Map only maps Prisma schema column definition to database columns name.
However, the Auth Prisma Adapter uses the Prisma definition...
Just looked at their example again. They don't actually change the casing of the Prisma model, just the casing for the actual DB schema, if that makes sense :/

image

So this seems to simply be used to ensure a consistent DB model schema.

RaMin0 (approved):

dashboard/src/repositories/postgres/verification.ts
```diff
@@ -58,7 +58,7 @@ export async function deleteVerificationToken(token: string): Promise<void> {
 
 export async function deleteExpiredVerificationTokens(): Promise<void> {
   try {
-    await prisma.verificationRequest.deleteMany({
+    await prisma.verificationToken.deleteMany({
       where: {
         expires: {
           lt: new Date(),
```
Woldbye, line 64: Not directly related to your changes but important to clear up. I assume all dates stored here are using the servers date-timestamps (I.e. NOT the clients)??

Woldbye (approved):
Nice work!! Code looks spot on to me 😀

Lindharden (changes requested):
It seems a lot of column values are dropped. We need to ensure no user data exists for these columns in production before this is merged to main. 

Are you using the official migration script that migrates next-auth schema definitions from v3 to v4? https://next-auth.js.org/getting-started/upgrade-v4#postgres

flopper123:
> It seems a lot of column values are dropped. We need to ensure no user data exists for these columns in production before this is merged to main.
> 
> Are you using the official migration script that migrates next-auth schema definitions from v3 to v4? https://next-auth.js.org/getting-started/upgrade-v4#postgres

@Lindharden No, I updated them manually.
I'm a little unsure of what the actual best practices are, because using the raw SQL migration we'd have to update use `prisma db pull` to update the the prisma schema based on the actual database.
I think this approach is safe, since we don't have any data in these tables on prod or anywhere, since it was never configured properly.
So I think this just counts as an initial database migration in some sense.
What do you think?

Lindharden (approved):

PR #402, opened 2025-07-28
Ensure IP is never logged

flopper123:
Removed any IP prints I could find.

Ensure NGINX doesn't print any output for `/analytics.js` and `/track`.
This seems to be the most transparent/privacy focused approach, from what I could find.

### DIFF
--- a/backend/src/geoip/mod.rs
+++ b/backend/src/geoip/mod.rs
@@ -121,11 +121,11 @@ impl GeoIpService {
         }
 
         if let Some(cached_result) = self.ip_cache.get(ip_address) {
-            debug!("GeoIP cache hit for IP: {}", ip_address);
+            debug!("GeoIP cache hit");
             return cached_result;
         }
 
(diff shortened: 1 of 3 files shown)


flopper123:
> > Quick question: Why would repeated requests from the same IP be a strong signal for malicious behavior? 🤔
> > If a person is using a site container Betterlytics, we are going to a see a lot of repeated requests.
> 
> Good point and you’re absolutely right that repeated requests aren’t malicious in itself.
> 
> Though, I do believe repeated requests from private IPs to sometimes be an early signal of abusive behavior. My concern is about being able to trace and identify a general pattern of abuse.
> 
> Logging is about observability: we need just enough insight to investigate if/when something looks off

It's a valid concern, but I'm not sure what they should try to abuse though? 🤔 
This will only disable the logging for the reverse proxy layer for `/analytics.js` and `/track`, and remove IP logging from the backend.
We still get the full visibility for the entire dashboard / the rest of the application 😄

Lindharden (approved):

PR #404, opened 2025-07-28
Added comparison to progress value table on hover

### LINKED ISSUE #403: Add comparison to progress bar tables
The design should follow the same vibe as on the regular pages we have:

Image

Should be added to the `Overview` page on the following tables:
- `Top Pages`
- `Geography`
- `Devices Breakdown`
- Traffic Sources
- Custom Events

flopper123:
Closes #403 

Added comparison to multi progress tables.
Added trend indicator.
Added hover effect to show trend.
Added text truncation on hover, if trend does not fit with text

image

Also added comparison to events table:

image


Lindharden:
This looks good 😄 we could consider adding the trend indicator icon to the very right in each row so that the trend for each row becomes apparent without having to hover each row. Then the specific percentage increase/decrease could be shown on hover as currently (but without the icon). What do you think of this? @flopper123

flopper123:
> This looks good 😄 we could consider adding the trend indicator icon to the very right in each row so that the trend for each row becomes apparent without having to hover each row. Then the specific percentage increase/decrease could be shown on hover as currently (but without the icon). What do you think of this? @flopper123

@Lindharden @Woldbye 

What about this:
image

We only show the trend, and nothing changes regarding that in the table.

Because, since this is the overview page, you can quickly see the data, however, if you're looking for more in-depth data, you can always go to the `Pages` page/whatever you are looking for:

image

Woldbye:
> > This looks good 😄 we could consider adding the trend indicator icon to the very right in each row so that the trend for each row becomes apparent without having to hover each row. Then the specific percentage increase/decrease could be shown on hover as currently (but without the icon). What do you think of this? @flopper123
> 
> 
> 
> @Lindharden @Woldbye 
> 
> 
> 
> What about this:
> 
> image
> 
> 
> 
> We only show the trend, and nothing changes regarding that in the table.
> 
> 
> 
> Because, since this is the overview page, you can quickly see the data, however, if you're looking for more in-depth data, you can always go to the `Pages` page/whatever you are looking for:
> 
> 
> 
> image
> 
> 

This looks good, but maybe we could add the trend percentage as a native html title tooltip on hover of the trend button

Lindharden:
> @Lindharden @Woldbye
> 
> What about this: image
> 
> We only show the trend, and nothing changes regarding that in the table.
> 
> Because, since this is the overview page, you can quickly see the data, however, if you're looking for more in-depth data, you can always go to the `Pages` page/whatever you are looking for:
> 
> image

I think this looks perfect @flopper123 !! But I definitely think the value should still appear on row-hover as you had implemented before.

I.e. the icon is always visible as seen on your image and the actual percentage change is then visible for each row upon hover. Given this is an overview page, this will allow the user to quickly scan negative/positive trends and see if they’re worth looking into - in which case they can dive deeper in the breakdown pages.

dashboard/src/app/actions/geography.ts
```diff
@@ -43,3 +46,51 @@ export const getWorldMapDataAlpha2 = withDashboardAuthContext(
     }
   },
 );
+
```
Woldbye, line 49: You mentioned yourself that we will need the same comparison for the tooltips on the map.

Therefore, I suggest generalizing this, as currently the logic is very specific to "top countries" and not reusable for viewing all countries.

dashboard/src/app/actions/geography.ts
```diff
@@ -43,3 +46,51 @@ export const getWorldMapDataAlpha2 = withDashboardAuthContext(
...
+export const getTopCountryVisitsAction = withDashboardAuthContext(
+  async (ctx: AuthContext, params: Omit<z.infer<typeof queryParamsSchema>, 'siteId'>) => {
```
Woldbye, line 51: Consider receiving the number of countries to return as a prop that defaults to 10

Woldbye (commented):
Very nice!! I like the design a lot <3 

I had a few remarks that I have left as comments.

dashboard/src/app/actions/geography.ts
```diff
@@ -43,3 +46,51 @@ export const getWorldMapDataAlpha2 = withDashboardAuthContext(
...
+        compareStartDate &&
+        compareEndDate &&
+        worldMapResponseSchema
+          .parse(
+            dataToWorldMap(
+              await fetchVisitorsByGeography(ctx.siteId, compareStartDate, compareEndDate, queryFilters),
+              CountryCodeFormat.Original,
+            ),
+          )
+          .visitorData.filter((row) => topCountries.includes(row.country_code));
```
Lindharden, line 85: It's a bit hard to interpret with context on phone. But is there a reason why we first parse this to world map format and then to the table format? Can't we avoid the world
map parsing entirely here to make it faster?

dashboard/src/app/actions/geography.ts
flopper123: I separated it, as I don't feel like they're showing the same data 😄 
And if I generalized them both to the same action, then I'd need to update all the world maps.
So I think in this case, that they should not be generalized.
Thoughts?

Woldbye: You can keep both actions seperate. What I am really requesting is an until method that handles comparing visitor data that this uses. Then we can modify the world data to use the same later

flopper123: I'm not too sure what you mean. The comparison is merged in the `toDataTable`. It's already a utility function that handles this. It can directly be put on the other action, once we're adding comparison to the world map, if that was what you meant 😄

Woldbye (approved):
Nice job adjusting the code!! Looks ready for merge

Woldbye (approved):

Lindharden (approved):

PR #405, opened 2025-07-29
Open Dashboard Settings in a dialog

RaMin0:
## Description

* Instead of its own dedicated page, "Dashboard Settings" now opens in a dialog to allow for better UX and help keep the user's context intact
* "Quick Settings" was removed since it only had the "Theme" toggle, which is already part of the "User Settings" (@flopper123)

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Style/UI changes

## Testing

- [x] Tests pass locally
- [ ] Added new tests
- [x] Manual testing completed
- [ ] Performance impact assessed

## Screenshots

https://github.com/user-attachments/assets/0925d0fd-db75-4b48-bab4-03a67e037cbc


Woldbye:
Great idea to open the dashboard settings in a dialogue! I agree that this is the right approach 

That said, I still believe quick settings are important. As the number of settings grows, the need for quick access will become even more critical. I understand that only a single setting is available currently - I think the right approach here is to add more settings not remove it.

dashboard/src/app/dashboards/DashboardCard.tsx
```diff
@@ -63,7 +63,7 @@ export default function DashboardCard({ dashboard }: DashboardCardProps) {
             </div>
             <div className='flex w-8 justify-center'>
               <Link
-                href={`/dashboard/${dashboard.id}/settings`}
+                href={`/dashboard/${dashboard.id}#settings`}
```
Woldbye, line 66: Question: What’s the purpose of using the browser native hash navigating to open the settings??

Woldbye: Thanks for the explanation! Since browsers natively support navigating to specific DOM elements via the \# hash, I’m a bit concerned that using `#settings` might unintentionally interfere with those native behaviors or with accessibility tools.

On a related note, in many projects I’ve worked on, we handle modals and overlays through a single, generic global overlay component. Usually it's controlled  it via a shared `useOverlay` hook or context, which manages opening, closing, and even updates the URL in a centralized way. This helps avoid issues like multiple overlays stacking up or z-index conflicts, and keeps URL handling consistent.

Maybe something like that could be a more sustainable approach here in the long run?

RaMin0: Your concern about accessibility is sound and justifies the extra server overhead of using search params. Let me switch to that 👌🏻

I like the idea of centralizing the modal handling as we depend on this UX a lot!

dashboard/src/hooks/use-hash-key.ts
```diff
@@ -0,0 +1,30 @@
...
+    return () => {
+      window.removeEventListener('hashchange', checkHash);
+    };
+  }, [key]);
+
+  const toggleKey = (state: boolean) => {
+    const hash = state ? `#${key}` : '';
+    router.push(`${window.location.search}${hash}`);
+    setIsKeySet(state);
+  };
```
flopper123, line 27: I've just tested it again after seeing this.
It looks like this causes it to a perform a full page reload/refresh, every time you add / remove the hash key.
I think that by pushing a URL to the router like this, it handles it as a redirect, causing the entire page to have to be fetched and loaded.

flopper123 (changes requested):

dashboard/src/app/dashboards/DashboardCard.tsx
Woldbye: I have not experienced any bugs with the global overlayy. Though, let me stress that I'm usually writing in vue3, not nextjs / react. 
I think you are correct in suggestion a more native react pattern than a global-overlay.

I think something like:
- useModal hook
- ModalProvider
- Modal component, with one instance per overlay

Is more flexible and idiomatic in nextjs

RaMin0:
> I still believe quick settings are important

@flopper123 Since we aligned on this, do you have any last words before I revert the change?

dashboard/src/app/dashboards/DashboardCard.tsx
RaMin0: Switched to using `searchParams` in e6875945845ada300f9f85d9ac2b06bb45b70493

dashboard/src/components/SettingsButton.tsx
```diff
@@ -5,24 +5,38 @@ import { Settings } from 'lucide-react';
...
+import DashboardSettingsDialog from '@/components/dashboardSettings/DashboardSettingsDialog';
+import { useSearchParamKey } from '@/hooks/use-search-param-key';
 
 export default function SettingsButton() {
   const [isPopoverOpen, setIsPopoverOpen] = useState(false);
+  const [showSettingsDialog, setShowSettingsDialog] = useSearchParamKey('settings');
+
+  const handleAdvancedSettingsClicked = () => {
+    setShowSettingsDialog(true);
+    setIsPopoverOpen(false);
```
RaMin0, line 17: Not sure if this should be here or in the `SettingsPopover` 🤔

flopper123 (approved):
Very nice! ⭐

flopper123:
> > I still believe quick settings are important
> 
> @flopper123 Since we aligned on this, do you have any last words before I revert the change?

I personally don't believe we'll have any quick dashboard settings that you'd want to be able to quickly change, especially now that we have a modal, but we can always change that 😄

Woldbye:
> > > I still believe quick settings are important
> 
> > 
> 
> > @flopper123 Since we aligned on this, do you have any last words before I revert the change?
> 
> 
> 
> I personally don't believe we'll have any quick dashboard settings that you'd want to be able to quickly change, especially now that we have a modal, but we can always change that 😄  

As long as the quick settings is available in the sidebar I am fine with this !!

Woldbye (approved):

PR #409, opened 2025-07-29
Fix country flags on Safari

### LINKED ISSUE #408: Safari is not very happy with the country flags
<table>
<thead>
<tr><th>Safari</th><th>Chrome</th>
</thead>
<tbody>
<tr>
<th>Image</th>
<th>Image</th>
</tr>
</tbody>
</table>

RaMin0:
## Description

Fix the width of country flags on Safari spanning more than it should. Since we're using `country-flag-icons/react/3x2`, hardcoding the flags' aspect ration to 3/2 seems to fix the issue

Closes #408

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Style/UI changes

## Testing

- [x] Tests pass locally
- [ ] Added new tests
- [x] Manual testing completed
- [ ] Performance impact assessed

## Screenshots

image

### DIFF
--- a/dashboard/src/components/language/CountryDisplay.tsx
+++ b/dashboard/src/components/language/CountryDisplay.tsx
@@ -19,7 +19,7 @@ export const CountryDisplay = ({
       <FlagIcon
         countryCode={countryCode} 
         countryName={countryName}
-        className='flex-shrink-0 w-[1lvw]' 
+        className='flex-shrink-0 aspect-[3/2] w-auto!' 
       />
       <span className='truncate max-w-full'>{countryName}</span>
     </div>
--- a/dashboard/src/components/language/CountryDisplay.tsx
+++ b/dashboard/src/components/language/CountryDisplay.tsx
@@ -19,7 +19,7 @@ export const CountryDisplay = ({
       <FlagIcon
         countryCode={countryCode} 
         countryName={countryName}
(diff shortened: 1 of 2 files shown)


dashboard/src/components/language/CountryDisplay.tsx
```diff
@@ -19,7 +19,7 @@ export const CountryDisplay = ({
       <FlagIcon
         countryCode={countryCode} 
         countryName={countryName}
-        className='flex-shrink-0 w-[1lvw]' 
+        className='flex-shrink-0 aspect-[3/2] w-auto!'
```
Woldbye, line 22: Strange strange strange bug :D Seems to be Safari misbehaving with `fit-content`. Unbeknownst to me this is a common bug. Switching to w-auto seems to be the most consistent fix across browsers. 

I think its very important to handle this everywhere, so all usages of the Icon is displayed correctly. Therefore, I suggest implementing it directly inside the `FlagIcon` component, and simply remove the w-[1lw] inside `CountryDisplay`, then you shouldn't need to add important either!

Sadly I dont have a Mac available to test this myself, but would appreciate if you could try it!

RaMin0: Done in ff3545a2721531bdc539bad945562391a9683ed5

Issue #410, opened 2025-07-30
Ensure metadata is updated based on environment variables


flopper123:
Currently `dashboard\src\app\robots.ts` and `dashboard\src\app\sitemap.ts` are generated at built-time, meaning that all self-hosters will have sitemaps.xml and robots.txt pointing to `https://betterlytics.io`, which is not great.

Optimally, these should be updated/regenerated if the environment variable `PUBLIC_BASE_URL` is updated.

Lindharden:
Hey @flopper123, sitemap is only for us. It is not relevant at all for self-hosters. Robots.txt is largely also irrelevant to self-hosters, although it may be a good idea to disallow robots altogether in the self-hosted version

flopper123:
> Hey [@flopper123](https://github.com/flopper123), sitemap is only for us. It is not relevant at all for self-hosters. Robots.txt is largely also irrelevant to self-hosters, although it may be a good idea to disallow robots altogether in the self-hosted version

They are both still generate for self-hosters though, as they get baked in the application.
It can probably be fixed by removing them if `IS_PUBLIC` is false.

Lindharden:
@flopper123 what should ideally happen is:

Robots.txt should have the following set for self-host version only:
```
user-agent: *
Disallow: /
```
This is to disable indexing/scanning of bots. For cloud-version, the robots.txt should remain as currently is.

For sitemap:
Don’t generate one, or if it exists, it should not point to anything publicly indexable.

PR #411, opened 2025-07-30
Building dashboard Docker image

flopper123:
This simplifies the Docker setup for self hosting setup.

There are still some remaining issues, which I've added here:
https://github.com/betterlytics/betterlytics/issues/410

The main issue is that metadata is generated at built-time and still inlines environment variables, as they are treated as build variables in these cases.


Lindharden (changes requested):
Nice simplification with the string Boolean type for Zod schemas.

Is there a reason why the .env file itself isn’t just copied instead of creating a third build env file? That seems more foolproof and we avoid having to maintain more files

Lindharden:
> In a Dockerfile it doesn't seem like you can use `COPY` to copy from a parent folder.
> 
> So it might need some more planning, but I just put it in as the "base" environment file for self hosting environment.
> 
> Thoughts?

I still think we should avoid the separate .env files as we'll end up having to maintain multiple files like we did in the beginning. But if it works for now, we can go with it. I think we should address it in the near future tho 😄

Lindharden (dismissed):

flopper123:
> > In a Dockerfile it doesn't seem like you can use `COPY` to copy from a parent folder.
> > So it might need some more planning, but I just put it in as the "base" environment file for self hosting environment.
> > Thoughts?
> 
> I still think we should avoid the separate .env files as we'll end up having to maintain multiple files like we did in the beginning. But if it works for now, we can go with it. I think we should address it in the near future tho 😄

You're correct. I've removed the `.env.build`, as we can simply do a copy of the production env file into the folder, once we build the Docker image 😄

PR #417, opened 2025-07-31
Added highlight on sidebar page when on page

### LINKED ISSUE #416: Highlight active sidebar item
For example, if the user is at the "Pages" page, it should highlight the "Pages" in the sidebar:

Image

Also, the sidebar should disappear when changing page on mobile devices

flopper123:
Closes #416

Sidebar link is now highlighted on active page:

image

Closes sidebar automatically on phone on page change

### DIFF
--- a/dashboard/src/components/sidebar/BAMobileSidebarTrigger.tsx
+++ b/dashboard/src/components/sidebar/BAMobileSidebarTrigger.tsx
@@ -4,9 +4,16 @@ import { Button } from '@/components/ui/button';
 import { useSidebar } from '@/components/ui/sidebar';
 import { cn } from '@/lib/utils';
 import { MenuIcon } from 'lucide-react';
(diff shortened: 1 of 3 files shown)


RaMin0 (dismissed):
Fancaayyy! ✨

Woldbye (commented):
Code looks good, but I think its a bad style choice to highlight the entire background of the active route for two reasons:

1. When it's desktop mode, and the sidebar becomes wide, the selection becomes very dominating
2. It interferes with hover-styles, as the coloring scheme is nearly identical to the one applied by hover.  On my monitor (which is always on night-mode with low colors), it actually appears completely identical.
 
Instead I want to suggest removing the bg-muted, and instead bold the text and add a yellow-ish text-shadow on both the icon and text.

Woldbye (approved):
Approving

PR #418, opened 2025-07-31
Add topbar-loader to all internal transitions between pages

### LINKED ISSUE #393: Fix transitions between pages
There's a delay between clicking on a page in the dashboard until any loading feedback is given.
The user should get instant feedback, instead of waiting for the entire page to load.

Woldbye:
Closes #393

Add a universal topbar-loader at the root, that displays a loading indicator for all internal navigations. Note that it's hidden for the initial navigation to the page. I am importing the component used from [nextjs-toploader](https://www.npmjs.com/package/nextjs-toploader) (Internally it relys on the popular [NProgress](https://www.npmjs.com/package/nprogress) library). 

Navigations that rely on the router from `next/navigation`, which is exposed through `useRouter`, does not trigger the loader by default. To circumvent this, I have added a custom hook `useBARouter` that wraps `useRouter`, and manually triggers the loader before navigation.

### GIFs
* In the GIFs below I am running on localhost with `reactStrictMode: false`. When you run it yourself on `localhost` it displays the loader twice in a row, (since nextjs double loads the page for debugging).
* The nextjs error in the bottom-left-corner of the GIF is unrelated to this.

msedge_DZ4h6cuzzZ

msedge_7D6b6NhHCB

### Question
* The toploader adds a spinner at the top-right-corner aswell. We can choose to hide it through a prop to the `TopLoaderBar` - I personally think it looks great, and integrates perfectly with the Topbar. WDYT? 
* It's also an option to follow popular pages such as `youtube` who position their `TopLoaderBar` below the `TopBar` (see below). I think this looks slightly better
msedge_9YCUULdQA0


Woldbye:
> Where else does the spinner appear @Woldbye? I only see it in the top right corner (in the top bar)

It only appears here, :D 

Did you miss the blue bar above the topbar that displays loading upon navigation?

Lindharden (commented):
I think this is good. Code looks good and it seems that NProgess is indeed a common way to handle route transitions when needed. I don't believe it's needed in our case in the first place, but I guess it doesn't hurt! 

Is it possible to make it into a progress bar that spans the entire top bar, so that the bottom border of the toolbar basically becomes a loading lightsaber?

I think this would add a more pleasant and modern feel, similarly to how Facebook, YouTube etc handles loading in modern UI's.

If this sounds too complex, we shouldn't spend too much time on it as we don't really need this loading indicator to begin with 😄

Lindharden:
> Is it possible to make it into a progress bar that spans the entire top bar, so that the bottom border of the toolbar basically becomes a loading lightsaber?
> 
> I think this would add a more pleasant and modern feel, similarly to how Facebook, YouTube etc handles loading in modern UI's.
> 
> If this sounds too complex, we shouldn't spend too much time on it as we don't really need this loading indicator to begin with 😄


I just noticed the process bar loading animation! This is what I was hoping for, nice! 😍

dashboard/src/hooks/use-ba-router.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  const push = async (...args: Parameters<typeof router.push>) => {
+    start();
+    return router.push(...args);
+  };
+
+  const replace = async (...args: Parameters<typeof router.replace>) => {
+    start();
+    return router.replace(...args);
+  };
```
flopper123, line 18: Suggestion: Wrap each in a `useCallback` 😄

flopper123, line 18: Suggestion: Wrap each in a useCallback 😄

flopper123 (changes requested):
Awesome addition! 😁 
Only had a tiny code comment.

Woldbye:
> I'm fine with having it everywhere if it can be consistent.
> 
> That would mean that it also needs to be in the docs. And that it can handle the transition between the landing page/dashboard/docs.
> 
> Because right now, if I click on the `Documentation`, there's no feedback 😄 

I will try to resolve why it's not appearing on navigation to docs from landing on this pr. 

Though, I will leave installation of the topbar on the documentation pages out of this pr.
Instead I will open an issue to include it there 😀

dashboard/src/app/layout.tsx
```diff
@@ -59,6 +60,7 @@ export default function RootLayout({
         />
       </head>
       <body className={`${geistSans.variable} ${geistMono.variable} antialiased`}>
+        <NextTopLoader color='var(--primary)' height={3} showSpinner={false} shadow={false} />
```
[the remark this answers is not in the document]
Woldbye: I don't see hows

Height for the TopLoader is the actual height of the loading bar in px.

The variable topbar-height denotes the height of the topbar.

dashboard/src/components/ExternalLink.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+  const handleClick = (e: React.MouseEvent<HTMLAnchorElement, MouseEvent>) => {
+    startLoader();
+    if (props.onClick) {
+      props.onClick(e);
+    }
+  };
+
+  return (
+    <a onClick={handleClick} {...props}>
```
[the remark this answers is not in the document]
flopper123: I again, don't know if this is just a comment for yourself, but if you don't want to propagate the `onClick`, you can do:

```ts
  const handleClick = (e: React.MouseEvent<HTMLAnchorElement, MouseEvent>) => {
    e.preventDefault(); // <--- This one here will stop the event from propagating
    startLoader();
    if (props.onClick) {
      props.onClick(e);
    }
  };
```

dashboard/src/app/globals.css
```diff
@@ -42,6 +42,7 @@
   --radius-md: calc(var(--radius) - 2px);
   --radius-lg: var(--radius);
   --radius-xl: calc(var(--radius) + 4px);
+  --topbar-height: calc(var(--spacing) * 14);
```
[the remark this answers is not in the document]
Woldbye: > Would it be possible to specify the height as being `h-14` here? 🤔  

Yes, it is. It's exactly what the code you are suggesting to change is already doing.

But you can't refer to h-14 in a css variable - this is the way

dashboard/src/components/ExternalLink.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+/** Wrapper for the anchor HTML-element (<a>), that triggers the TopLoader's loading animation before navigating */
+const ExternalLink: React.FC<ExternalLinkProps> = ({ children, ...props }) => {
+  const { start: startLoader } = useTopLoader();
+
+  const handleClick = (e: React.MouseEvent<HTMLAnchorElement, MouseEvent>) => {
+    startLoader();
+    if (props.onClick) {
+      props.onClick(e);
+    }
```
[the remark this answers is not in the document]
Woldbye: Well the consumer might do:

```
<ExternalLink onClick={ 
async (e) => 
   e.preventDefault();
   await something();
}/>
```

To support this the `handleClick()` inside ExternalLink would need to be async and await it.

Woldbye:
> Very nice! I have added some small code comments.
> 
> On phone, when the sidebar is open, it looks a bit misplaced, as you can see here:
> 
> image

I have moved the Loader to the top of the screen on mobile views. This seemed like the most consistent way to resolve this

dashboard/src/components/topbar/PublicTopBar.tsx
```diff
@@ -31,23 +32,23 @@ export default function PublicTopBar() {
...
               href='/docs'
               title='Complete Betterlytics Documentation'
               className='text-muted-foreground hover:text-foreground text-sm font-medium transition-colors'
             >
               Documentation
-            </a>
+            </ExternalLink>
             <Link
-              href='#pricing'
+              href='/#pricing'
```
flopper123, line 51: Question: Why change this? 😄

dashboard/src/components/ExternalLink.tsx
```diff
@@ -0,0 +1,29 @@
...
+  );
+
+  return (
+    <a {...props} onClick={handleClick}>
+      {children}
+    </a>
+  );
+};
+
+export default ExternalLink;
```
flopper123, line 29: Suggestion: Rewrite this as a `function` for consistency.

flopper123 (changes requested):
Very nice! I'll first be able to test it our later, but I have some small code comments in the meantime 😄

dashboard/src/components/topbar/PublicTopBar.tsx
flopper123: That makes sense.
However, it seems to introduce a new bug on the landing page.
If the landing page is already loaded, and I click on the "pricing", it needs to fetch almost 1mb of data + do a full navigation before being able to scroll down.
So it seems to be doing a full re-fetch of something, before it can even scroll - it potentially has to re-fetch the entire landing page. This feels a little clunky instead of just scrolling down to already loaded content, if that makes sense 😄 

image

flopper123 (changes requested):
Found one small bug 😅 
Other than that it's ready to be approved - GJGJ 😎 🔥

dashboard/src/components/topbar/PublicTopBar.tsx
flopper123: Nevermind, this also happens on `main`. It must be a dev thing

PR #419, opened 2025-08-01
Reenabled sorting on tables

### LINKED ISSUE #415: Fix table not being sortable after adding comparison
The component `TableCompareCell` currently breaks the sorting on colums where it's used.

It should still be able to sort based on the value.

flopper123:
Closes #415

You can now sort on tables again 🎉


Lindharden (approved):
Nice👍

PR #433, opened 2025-08-12
431 add next intl with app routing on public pages

### LINKED ISSUE #350: Add support for pluralization Localization Rule in dictionary translations via argument to `t`
Currently, the `t` function in our localization system supports key-based lookups for nested dictionary entries. However, it lacks built-in support for singular/plural variations based on a numeric value. This makes it difficult to handle cases where the translation needs to change depending on quantity.

### Proposal
Enhance the `t` function to support a second argument: a number that determines whether the singular or plural form of a string should be returned.

### Example
Assume this dictionary structure and `const DAYS = ['Monday', 'Sunday']` 
```
{
  "misc": {
    "day": "day | days"
  }
}
```

| Call | Expected Output |
| -- | -- |
| `t('misc.day')` | "day" |
| `t('misc.day', 1)` | "day" |
| `t('misc.day', 2)` | "days" |
| `t('misc.day', DAYS.length)` | "days" |  

The second argument should be optional. If omitted or equal to 1, the singular form is returned. If greater than 1 (or 0), the plural form is returned. The delimiter for splitting singular/plural strings is " | " in this example, but other delimiters could be used. 

### Acceptance criteria
* t can accept an optional number as the second argument.
* Strings containing " | " are split into singular/plural variants.
* The correct form is returned based on the number passed.
* If the key does not contain " | ", the original string is returned as-is, to preserve backward compatibility.

Note that " | " could be any separator of choice.

### Suggested implementation
This should probably be implemented on the `t` function, `addTFunction()` to delegate pluralization logic to a separate helper, making the solution modular and extensible. We need to ensure it can support other localization-rules also in the future.

### LINKED ISSUE #429: Change localization to use 'next-intl' without '18n-app-routing' on Dashboard and '18n-app-routing' on Public
The localization setup should be changed to use the `next-intl` library.

The idea is to use 'next-intl' without '18n-app-routing' on the dashboard, and use '18n-routing' on the public pages. Setting up 'next-intl' without '18n-routing', and migrating the public pages to use '18n-app-routing'  should be separated into different issues to keep the changes at a minimum.
 
Using 'next-intl' is much more flexible, and requires no manual setup. It also allows us to more smoothly integrate the localization dictionary for the dashboard and public pages. Further it would complete #384 #350 , as these features are available straight out of the box.

#### Resources
The official `next-intl` documentation offers guides to setup [with](https://next-intl.dev/docs/getting-started/app-router/with-i18n-routing) and [without](https://next-intl.dev/docs/getting-started/app-router/without-i18n-routing) `i18n-app-routing`.

### LINKED ISSUE #430: 🌐 Change localization to use 'next-intl' on Dashboard
Replace our localization setup on the Dashboard to use `next-intl`  without `18n-app-routing`. The locale should be chosen exactly as it is now (from user-settings  / environment).

### Requirements
* Auto-completion on `t` function like our existing setup.
* Load locale from user-settings like our existing setup

### LINKED ISSUE #431: Add `next-intl` with `app-routing` on public-pages
Add `next-intl` localization on the public-pages [with `18n-app-routing`](https://next-intl.dev/docs/getting-started/app-router/with-i18n-routing). 

### Requirements
* App routing on public pages
* Initial locale parsed from browser settings
* Should use the same localization dictionary and locales as the dashboard

### LINKED ISSUE #441: Fix links in Integration Sheet
The links to our `npm` package has been removed since adding localization.
Once you can use HTML tags in localization, these should be re-added.

Image

Lindharden:
- Replace existing setup with Next i18n on the dashboard
- Added canonical linking for SEO optimization, so that /locale has /locale defined as canonical url.
- Added next-intl navigation wrapper for Nextjs Link to create locale aware Links.
- Moved all public pages to [locale] path directory.
- Removed conditional footer and topbar.
- Added translations for all public pages.
- Cache UserSettings in auth session. Cache is invalidated every 30 minutes or if the client updates it. This ensures data can still be updated between devices. Note that this change is necessary to keep our navigations fast, even though we access user-settings on each dashboard request.

This PR is largely attributed to @Woldbye. The public pages are translated by @Lindharden, while the next-intl setup, private page translations & more is introduced by @Woldbye. 

Closes #431 
Closes #441 
Closes #430 
Closes #350 
Closes #429


dashboard/messages/da.json
```diff
@@ -0,0 +1,703 @@
+{
```
flopper123, line 1: Question: Is there a reason why `dashboard/src/dictionaries/da.json` has been moved outside of the `src` folder into here? Because it seems a little weird to point to something from inside the `src` folder to outside of the `src` folder.

And also, why rename it to `messages` instead of `dictionaries`? I personally feel like `dictionaries` is more aligned with `intl` rather than `messages` 😄

dashboard/src/app/Providers.tsx
```diff
@@ -5,6 +5,8 @@ import { SessionProvider } from 'next-auth/react';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { ThemeProvider } from '@/app/ThemeProvider';
 import { TooltipProvider } from '@/components/ui/tooltip';
+import { NextIntlClientProvider } from 'next-intl';
+import { getLocale } from 'next-intl/server';
```
flopper123, line 9: These seems to be unused 😄

flopper123 (changes requested):
Would it be possible to use the English text as the key?
Cause I'm afraid of this quickly becoming messy, as we no longer can search for the text we see on the page.
I'm fine with having the `public.landing.` for instance, but instead of doing:
`public.landing.title` it could be `public.landing.'Works with your favorite tools'`

That would also make it way easier for other people to translate, as `Works with your favorite tools` is way easier to understand than `public.landing.title` where you'd have to look up into the English dictionary every time.

dashboard/messages/da.json
Woldbye: Messages is the name used internally by next-i18n, but we could definitely change it. I personally agree that dictionary is more universal, but as this is open source, I thought it's better to stick with the conventions from their docs

dashboard/src/i18n/request.ts
```diff
@@ -0,0 +1,25 @@
+import { getUserSettingsAction } from '@/app/actions/userSettings';
+import { SupportedLanguages } from '@/constants/supportedLanguages';
+import { DEFAULT_LANGUAGE } from '@/constants/supportedLanguages';
+import { getRequestConfig } from 'next-intl/server';
+
+export default getRequestConfig(async ({ requestLocale }) => {
+  let locale = await requestLocale;
+  if (!locale) {
+    try {
+      const result = await getUserSettingsAction();
```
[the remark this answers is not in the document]
Woldbye: I gotta say I disagree with this for one, you risk client side hydration issues if the user settings gets updated within cache ttl

dashboard/src/components/topbar/PublicTopBar.tsx
```diff
@@ -84,33 +82,31 @@ export default function PublicTopBar() {
...
+                title={t('documentation')}
               >
-                Documentation
+                {t('documentation')}
               </ExternalLink>
 
               <div className='border-t pt-3'>
                 {session ? (
-                  <Link href='/dashboards' onClick={closeMobileMenu}>
+                  <Link href='/dashboards' onClick={closeMobileMenu} locale={'false' as unknown as any}>
```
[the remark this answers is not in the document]
Woldbye: @Lindharden could you help resolve this. I personally don't think it's necessary at all to pass undefined locale - the middleware / i18n router should take care of it

dashboard/src/app/layout.tsx
```diff
@@ -35,13 +33,15 @@ const organizationStructuredData = generateStructuredData('organization', {
   path: '/',
 });
 
-export default function RootLayout({
+export default async function RootLayout({
```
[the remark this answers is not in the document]
Lindharden: I'm actually a bit concerned about this. I think this makes every page dynamically rendered. I believe we must avoid request-bound APIs in the app root in order to benefit from static page generation. 

Thus we likely have to remove next intl provider from rootlayout.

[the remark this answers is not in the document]
Lindharden: I've made a PR of a suggestion on how we can possibly fix this. Please take a look and see if you both agree @Woldbye @flopper123 

https://github.com/betterlytics/betterlytics/pull/454

dashboard/src/components/topbar/PublicTopBar.tsx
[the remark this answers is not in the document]
Lindharden: I have changed refactored the Link component to a Next Link component now :)

dashboard/src/app/[locale]/dpa/page.tsx
```diff
@@ -0,0 +1,218 @@
...
+          <div className='border-border border-b px-6 py-8'>
+            <StructuredData config={seoConfig} />
+            <h1 className='text-foreground text-3xl font-bold'>{t('title')}</h1>
+            <p className='text-muted-foreground mt-2 text-lg'>
+              {t('lastUpdated')}{' '}
+              {new Date().toLocaleDateString(locale, {
+                year: 'numeric',
+                month: 'long',
+                day: 'numeric',
+              })}
```
Comment (flopper123) [inline dashboard/src/app/[locale]/dpa/page.tsx:37]: Question: Will this get statically generated or will it say "Last updated" to the current day, every day? 😄

dashboard/src/app/[locale]/terms/page.tsx
```diff
@@ -0,0 +1,280 @@
...
+          <div className='border-border border-b px-6 py-8'>
+            <StructuredData config={seoConfig} />
+            <h1 className='text-foreground text-3xl font-bold'>{t('title')}</h1>
+            <p className='text-muted-foreground mt-2 text-lg'>
+              {t('lastUpdated')}{' '}
+              {new Date().toLocaleDateString(locale, {
+                year: 'numeric',
+                month: 'long',
+                day: 'numeric',
+              })}
```
Comment (flopper123) [inline dashboard/src/app/[locale]/terms/page.tsx:38]: Same question again

dashboard/src/i18n/navigation.ts
```diff
@@ -0,0 +1,58 @@
...
+  getPathname,
+} = createNavigation(routing);
+
+//! TODO: Resolve types in below
+// const unlocalizedPattern = new RegExp(UNLOCALIZED_REGEX);
+
+// function extractPath(href: string | { pathname?: string }): string {
+//   if (typeof href === 'string') return href;
+//   return href?.pathname ?? '';
+// }
```
flopper123, line 20: Question: Should these comments be removed?

dashboard/src/app/[locale]/dpa/page.tsx
Comment (Woldbye) [reply dashboard/src/app/[locale]/dpa/page.tsx:37]: I also dont understand this code at all. Shouldnt it be a constant indicating when we last updated DPA??

Comment (Lindharden) [reply dashboard/src/app/[locale]/dpa/page.tsx:37]: It shows the date of when it was statically generated. I.e. when we recompile these, the date changes to today. Ideally it should reflect the actual date something was updated, but it's not too important.

In this PR I just updated the date to work with the locale so that it properly translates the displayed date. The logic remains the same.

@Woldbye @flopper123

dashboard/src/app/[locale]/terms/page.tsx
Comment (Lindharden) [reply dashboard/src/app/[locale]/terms/page.tsx:38]: See answers above.

PR #446, opened 2025-08-17
Added onboarding flow

flopper123:
image

image

image


Lindharden:
It's still not centered:
{8645A8C5-B235-4CBE-AEB2-C5367D380C8D}

{55479FA6-95D4-4D63-BE1B-7E4A781B5D3C}

Also, when you log in via oauth on the register page, you get redirected to the register page again. Sometimes I'm redirected to the second step, sometimes I'm not.

PR #452, opened 2025-08-18
Made theme switch depend on resolvedTheme to get proper initial state

### LINKED ISSUE #437: Light / Dark mode toggle issues on System
## Light / Dark mode quick switch issue

If you set your mode to be System (and it's dark mode), then the color mode is set to dark mode correctly:

Image

However, if you flip the toggle, it says that it's light mode, but it's still in dark mode:

Image

If you then toggle it again, it switches to light mode, but says that it's dark mode:

Image

Lindharden:
Closes #437


flopper123 (approved):
Super nice!

PR #463, opened 2025-08-22
331 track websites core web vitals

### LINKED ISSUE #331: Track Websites Core Web Vitals
Track core web vitals, such as:
LCP, INP, CLA, etc.

This will help users track their websites performance.

Lindharden:
{C35AC3C1-C242-44A6-9F84-2C180C154816}

Closes #331


backend/src/db/mod.rs
```diff
@@ -286,6 +286,8 @@ async fn run_inserter_worker(
             site_id = %row.site_id, 
             visitor_id = %row.visitor_id,
             session_id = %row.session_id,
+            event_type = ?row.event_type,
+            custom_event_name = ?row.custom_event_name,
```
flopper123, line 290: Question: Why use `?` instead of `%` here? 😄

backend/src/processing/mod.rs
```diff
@@ -177,7 +177,10 @@ impl EventProcessor {
...
     async fn handle_event_types(&self, processed: &mut ProcessedEvent) -> Result<()> {
         let event_name = processed.event.raw.event_name.clone();
-        if processed.event.raw.is_custom_event {
+        if event_name.eq_ignore_ascii_case("cwv") {
+            processed.event_type = "cwv".to_string();
+            processed.custom_event_json = processed.event.raw.properties.clone();
+        } else if processed.event.raw.is_custom_event {
             processed.event_type = "custom".to_string();
             processed.custom_event_name = event_name;
             processed.custom_event_json = processed.event.raw.properties.clone();
```
flopper123, line 186: This is a bit stupid design choice from my side. But this event name check should be checked after checking if the event is a custom event to prevent clashing in case of users using a custom event name called `cwv` 😄 It's definitely something I want to update soon, such that this is not necessary. I've just opened an issue for this https://github.com/betterlytics/betterlytics/issues/473

backend/src/processing/mod.rs
```diff
@@ -177,7 +177,10 @@ impl EventProcessor {
     /// Handle different event types
     async fn handle_event_types(&self, processed: &mut ProcessedEvent) -> Result<()> {
         let event_name = processed.event.raw.event_name.clone();
-        if processed.event.raw.is_custom_event {
+        if event_name.eq_ignore_ascii_case("cwv") {
+            processed.event_type = "cwv".to_string();
+            processed.custom_event_json = processed.event.raw.properties.clone();
```
flopper123, line 182: Suggestion: I would definitely prefer if we typed out the properties, as this `properties` is meant for `custom_event_json`, which in turn is meant to store arbitrary data for the custom events for users.
I think it'd be better if we added the `cwv` fields to the ClickHouse events table instead, as it'll give us much better validation / safety / performance 😄

dashboard/src/app/dashboard/[dashboardId]/web-vitals/interactiveWebVitalsChartSection.tsx
```diff
@@ -0,0 +1,230 @@
...
+  const [active, setActive] = useState<CoreWebVitalName>('CLS');
+  const seriesByMetric = use(seriesPromise);
+
+  function formatCardValue(metric: CoreWebVitalName, value: number | null): React.ReactNode {
+    if (value === null) return '—';
+    const colorStyle: React.CSSProperties = { color: getCwvStatusColor(metric, value) };
+
+    const display = metric === 'CLS' ? value.toFixed(3) : formatCompactFromMilliseconds(value);
+    return <span style={colorStyle}>{display}</span>;
+  }
```
Comment (flopper123) [inline dashboard/src/app/dashboard/[dashboardId]/web-vitals/interactiveWebVitalsChartSection.tsx:33]: Suggestion: Move this to outside of component function to prevent recomputing it on every re-render.

dashboard/src/app/dashboard/[dashboardId]/web-vitals/interactiveWebVitalsChartSection.tsx
```diff
@@ -0,0 +1,230 @@
...
+import { formatCWV, getCwvStatusColor } from '@/utils/formatters';
+import { CWV_THRESHOLDS } from '@/constants/coreWebVitals';
+import MetricInfo from '@/app/dashboard/[dashboardId]/web-vitals/MetricInfo';
+
+type Props = {
+  summaryPromise: Promise<CoreWebVitalsSummary>;
+  seriesPromise: Promise<CoreWebVitalsSeries>;
+};
+
+export default function InteractiveWebVitalsChartSection({ summaryPromise, seriesPromise }: Props) {
```
Comment (flopper123) [inline dashboard/src/app/dashboard/[dashboardId]/web-vitals/interactiveWebVitalsChartSection.tsx:21]: Suggestion: As this file is a component, uppercase the first letter of the file 😄

dashboard/src/app/dashboard/[dashboardId]/web-vitals/interactiveWebVitalsChartSection.tsx
```diff
@@ -0,0 +1,230 @@
...
+    if (value === null) return '—';
+    const colorStyle: React.CSSProperties = { color: getCwvStatusColor(metric, value) };
+
+    const display = metric === 'CLS' ? value.toFixed(3) : formatCompactFromMilliseconds(value);
+    return <span style={colorStyle}>{display}</span>;
+  }
+
+  const P75Badge = () => <span className='text-muted-foreground ml-2 text-xs font-medium'>p75</span>;
+
+  const cards: SummaryCardData[] = [
```
Comment (flopper123) [inline dashboard/src/app/dashboard/[dashboardId]/web-vitals/interactiveWebVitalsChartSection.tsx:37]: Suggestion: Wrap in a `useMemo`

dashboard/src/app/dashboard/[dashboardId]/web-vitals/interactiveWebVitalsChartSection.tsx
```diff
@@ -0,0 +1,230 @@
...
+    const stroke = idx === 0 ? 'var(--cwv-threshold-good)' : 'var(--cwv-threshold-ni)';
+    return { y, label, stroke, strokeDasharray: '6 6', labelFill: stroke };
+  });
+
+  const SERIES_DEFS: ReadonlyArray<MultiSeriesConfig> = [
+    { dataKey: 'value.0', stroke: 'var(--cwv-p50)', name: 'P50' },
+    { dataKey: 'value.1', stroke: 'var(--cwv-p75)', name: 'P75' },
+    { dataKey: 'value.2', stroke: 'var(--cwv-p90)', name: 'P90' },
+    { dataKey: 'value.3', stroke: 'var(--cwv-p99)', name: 'P99' },
+  ] as const;
```
Comment (flopper123) [inline dashboard/src/app/dashboard/[dashboardId]/web-vitals/interactiveWebVitalsChartSection.tsx:156]: Suggestion:
- Move the functions without dependencies out of the function to declutter
- Move constants outside of component for performance / decluttering
- If `referenceLines` is expensive to compute, wrap in useMemo

dashboard/src/app/dashboard/[dashboardId]/web-vitals/page.tsx
```diff
@@ -0,0 +1,80 @@
...
+  );
+  const perOsPromise = fetchCoreWebVitalsByDimensionAction(dashboardId, startDate, endDate, queryFilters, 'os');
+
+  return (
+    <div className='container space-y-6 p-6'>
+      <DashboardFilters />
+      <Suspense fallback={<SummaryCardsSkeleton />}>
+        <InteractiveWebVitalsChartSection summaryPromise={summaryPromise} seriesPromise={seriesPromise} />
+      </Suspense>
+      <Suspense>
```
Comment (flopper123) [inline dashboard/src/app/dashboard/[dashboardId]/web-vitals/page.tsx:69]: Suggestion: Add a table skeleton as fallback

dashboard/src/app/dashboard/[dashboardId]/web-vitals/webVitalsTableSection.tsx
```diff
@@ -0,0 +1,200 @@
...
+  perCountryPromise,
+  perBrowserPromise,
+  perOsPromise,
+}: Props) {
+  const data = use(perPagePromise);
+  const devices = use(perDevicePromise);
+  const countries = use(perCountryPromise);
+  const browsers = use(perBrowserPromise);
+  const operatingSystems = use(perOsPromise);
+  type PercentileKey = 'p50' | 'p75' | 'p90' | 'p99';
```
Comment (flopper123) [inline dashboard/src/app/dashboard/[dashboardId]/web-vitals/webVitalsTableSection.tsx:35]: Suggestion: Move type definition outside of function 😄

dashboard/src/app/dashboard/[dashboardId]/web-vitals/webVitalsTableSection.tsx
```diff
@@ -0,0 +1,200 @@
...
+  const [activePercentile, setActivePercentile] = useState<PercentileKey>('p75');
+  const percentileIndex: Record<PercentileKey, number> = { p50: 0, p75: 1, p90: 2, p99: 3 };
+
+  const makeColumns = (label: string, renderIcon?: (key: string) => React.ReactNode): ColumnDef<Row>[] => {
```
Comment (flopper123) [inline dashboard/src/app/dashboard/[dashboardId]/web-vitals/webVitalsTableSection.tsx:39]: Suggestion: Wrap this in a useCallback, as it looks a little expensive to recompute 😄

dashboard/src/app/dashboard/[dashboardId]/web-vitals/webVitalsTableSection.tsx
```diff
@@ -0,0 +1,200 @@
...
+      defaultSorting,
+    },
+    {
+      key: 'os',
+      label: 'Operating Systems',
+      data: operatingSystems,
+      columns: osColumns,
+      defaultSorting,
+    },
+  ];
```
Comment (flopper123) [inline dashboard/src/app/dashboard/[dashboardId]/web-vitals/webVitalsTableSection.tsx:165]: Suggestion: Wrap in useMemo as it uses data

dashboard/src/components/InteractiveChart.tsx
```diff
@@ -46,7 +46,9 @@ const InteractiveChart: React.FC<InteractiveChartProps> = React.memo(
                   tickLine={false}
                   axisLine={false}
                   className='text-muted-foreground'
-                  tickFormatter={axisFormatter}
+                  tickFormatter={(value) =>
+                    axisFormatter(new Date(typeof value === 'number' ? value : Date.parse(String(value))))
```
flopper123, line 50: Nitpick: If the string is already parseable, just skip the parse to number part 😄 
```suggestion
                    axisFormatter(new Date(typeof value === 'number' ? value : String(value)))
```

dashboard/src/components/MultiSeriesChart.tsx
```diff
@@ -0,0 +1,132 @@
...
+              <ComposedChart data={data} margin={{ top: 10, right: 30, left: 0, bottom: 0 }}>
+                <CartesianGrid strokeDasharray='3 3' className='opacity-10' />
+                <XAxis
+                  dataKey='date'
+                  fontSize={12}
+                  tickLine={false}
+                  axisLine={false}
+                  className='text-muted-foreground'
+                  tickFormatter={(value) =>
+                    axisFormatter(new Date(typeof value === 'number' ? value : Date.parse(String(value))))
```
flopper123, line 71: Same nitpick as before - by no means a blocker though 🤣

dashboard/src/components/MultiSeriesChart.tsx
```diff
@@ -0,0 +1,132 @@
...
+                    dataKey={s.dataKey}
+                    stroke={s.stroke}
+                    strokeWidth={s.strokeWidth ?? 2}
+                    dot={s.dot ?? false}
+                    name={s.name}
+                  />
+                ))}
+                {referenceLines?.map((r, i) => (
+                  <ReferenceLine
+                    key={`ref-${i}`}
```
flopper123, line 104: Question: Is there another property you can use than index for key? It looks like the `label` might always be present, even though this type suggests it can be undefined

flopper123 (changes requested):
Overall works super nice! 🤩 
Have some overall comments:

The performance bars should all be aligned:
image

The table should be responsive on phone:
image

It seems like the CWV are often not sent. I tried to go to different sites / refreshed, and noticed that it didn't show any data for some of the pages I had visited. I checked the network tab and saw these. They are CWV events (which in this case seems to be working). However, on other pages I saw no CWV events, even though I refreshed / navigated / etc:
image

Also, this page should probably have comparison on it, such that you can easily see how your page's performance has changed relative to the last timeframe. Thoughts?

I've also added some initial code comments, as I couldn't finish the review in one go 🙏

dashboard/src/app/dashboard/[dashboardId]/web-vitals/InteractiveWebVitalsChartSection.tsx
```diff
@@ -0,0 +1,254 @@
...
+            <div style={{ color: getCwvStatusColor('TTFB', summary.ttfbP75) }}>
+              <CoreWebVitalBar metric='TTFB' value={summary.ttfbP75} thresholds={CWV_THRESHOLDS} />
+            </div>
+          ),
+        chartColor: 'var(--chart-5)',
+        isActive: active === 'TTFB',
+        onClick: () => setActive('TTFB'),
+      },
+    ],
+    [summary, active],
```
Comment (flopper123) [inline dashboard/src/app/dashboard/[dashboardId]/web-vitals/InteractiveWebVitalsChartSection.tsx:157]: Same with this one

PR #465, opened 2025-08-23
Added scroll depth

### LINKED ISSUE #14: Add scroll depth tracking to tracking script
We need to enhance the existing tracking script by adding scroll depth tracking. This will allow us to measure how far users scroll down a page. 

Requirements
- Track scroll depth as a percentage.
- Send scroll depth events only after a certain amount of time without activity to avoid redundant data.
- Ensure the scroll tracking is lightweight and does not negatively affect performance.
Acceptance Criteria
- Each scroll threshold is only tracked once per page load. I.e if the user scrolls down and then up, only the farthest scroll percentage is sent. This is up for discussion.
- Script behaves correctly on different screen sizes and devices.
- Consider debouncing or throttling to optimize performance.

flopper123:
Closes #14

### DIFF
--- a/dashboard/messages/da.json
+++ b/dashboard/messages/da.json
@@ -18,6 +18,11 @@
         ]
       }
     },
+    "notFound": {
+      "title": "404 - Siden blev ikke fundet",
+      "description": "Denne side ser ud til at være vandret ud i analytics-dimensionen. Men bare rolig, vi har styr på det!",
+      "goHome": "Tilbage til forsiden"
+    },
     "about": {
       "seo": {
         "title": "Om Betterlytics | Open Source, Privatlivsfokuseret Webanalyse",
@@ -410,12 +415,9 @@
       },
       "sla": {
         "title": "Serviceniveau og support",
-        "availabilityTitle": "Service-tilgængelighed",
         "availabilityBody": "Vi stræber efter høj tilgængelighed, men leverer tjenesten \"som den er\" og \"som tilgængelig\". Vi garanterer ikke uafbrudt drift eller fejlfri ydeevne.",
         "availabilityBodySuffix": "Vi kan planlægge vedligeholdelse; vi vil bestræbe os på at give forudgående besked.",
-        "supportTitle": "Support",
         "supportBody": "Teknisk support leveres via e-mail efter rimelig indsats. Vi tilstræber hurtige svar, men garanterer ikke specifikke svartider.",
-        "bugsTitle": "Fejl og problemer",
         "bugsBody": "Vi tester vores funktioner grundigt, men ingen software er perfekt. Vi registrerer rapporterede fejl og prioriterer rettelser, især for sikkerheds- eller privatlivsproblemer. Ikke alle rapporterede fejl vil blive rettet med det samme."
       },
       "liability": {
@@ -486,11 +488,6 @@
         "li1": "Anonyme data (standard, anonymiserede analyser), og",
         "li2": "Personlige eller pseudo-anonyme data (valgfri funktioner, eller enhver utilsigtet inklusion af personlige eller pseudo-anonyme data i URL'er, forespørgselsparametre eller hændelses-payloads), f.eks. session replays, brugerdefinerede identifikatorer eller brugerdefinerede hændelser."
       },
-      "intro": {
-        "title": "Introduktion",
-        "p1": "Denne Databehandleraftale (\"DPA\") er en del af Betterlytics' Servicevilkår og regulerer databehandlingen, når du bruger vores analysetjeneste. Denne aftale er udformet for at sikre overholdelse af Databeskyttelsesforordningen (GDPR) og andre gældende love om databeskyttelse.",
-        "p2": "Vigtigt: Betterlytics er specifikt designet til kun at behandle anonyme, aggregerede data. Ingen persondata indsamles, lagres eller behandles af vores tjeneste."
-      },
       "definitions": {
         "title": "Definitioner",
         "controller": "\"Dataansvarlig\": Dig (kunden), som fastlægger formål og midler til behandling af data fra dine website-besøgende.",
@@ -502,7 +499,6 @@
       "processingDetails": {
         "title": "Detaljer om databehandlingen",
         "naturePurposeTitle": "Behandlingens karakter og formål",
-        "body": "Betterlytics behandler data på vegne af den Dataansvarlige udelukkende til følgende formål:",
         "npLi1": "At levere analyser af webtrafik og besøgsadfærd.",
         "npLi2": "At generere aggregerede rapporter om webtrafik og brugsmønstre",
         "npLi3": "At detektere og filtrere bottrafik for nøjagtige statistikker",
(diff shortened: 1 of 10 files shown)


static/analytics.js
```diff
@@ -207,4 +213,39 @@
...
+
+    const windowHeight = window.innerHeight;
+    const docHeight = document.documentElement.scrollHeight;
+
+    if (docHeight === 0) {
+      return 100;
+    }
+
+    return 100 * (scrollTop / (docHeight - windowHeight));
+  }
```
Woldbye, line 238: This would also capture scroll depth on the main window scroll area. It seems like we need to account for layouts where it's not window. 

Consider:

[Header]
[ScrollableBody]
[Footer]

Type setups

Woldbye (commented):
I think we need to reconsider what scroll depths mean in the context of SPAs, PWAs, and “dashboard-like” apps w. (header/footer pinned, scrollable middle content).

For these cases the window scroll it not enough

static/analytics.js
flopper123: If you're thinking of layouts like the dashboard, where you have a sticky header, then it still works 😄

static/analytics.js
```diff
@@ -272,6 +278,45 @@
...
+    if (docHeight === 0) {
+      return 100;
+    }
+
+    if (docHeight - windowHeight <= 0) {
+      return 100;
+    }
+
+    return 100 * (scrollTop / (docHeight - windowHeight));
+  }
```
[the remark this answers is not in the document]
flopper123: Fixed 👍

dashboard/src/repositories/clickhouse/pages.ts
```diff
@@ -204,6 +220,7 @@ export async function getPageMetrics(
     pageviews: Number(row.pageviews),
     bounceRate: row.bounceRate,
     avgTime: row.avgTime,
+    scrollDepthPercentage: Number(row.scrollDepthPercentage),
```
Lindharden: You parse it to Number here, but you also added a preprocess step in the zod entity for parsing it as a number. We should probably just remove this? 😄

backend/src/db/models.rs
```diff
@@ -39,6 +39,8 @@ pub struct EventRow {
     pub cwv_inp: Option<f32>,
     pub cwv_fcp: Option<f32>,
     pub cwv_ttfb: Option<f32>,
+    pub scroll_depth_percentage: Option<f32>,
+    pub scroll_depth_pixels: Option<f32>,
```
Lindharden: In the migration file, you explicitly put these after custom_event_json row, so we should probably move these up in the model.rs so that it semantically reflects the clickhouse model 😄

static/analytics.js
```diff
@@ -272,6 +279,53 @@
...
+
+    if (docHeight === 0) {
+      return 100;
+    }
+
+    if (docHeight - windowHeight <= 0) {
+      return 100;
+    }
+
+    return 100 * (maxCapturedScrollPixels / (docHeight - windowHeight));
```
Lindharden: I think it makes more sense to use a bottom-of-viewport approach here, it better represents how much of the page the user has actually seen.

For example, on a 4K ultrawide monitor, up to 70% of the page might be visible at once, but a top-of-viewport calculation would still report 0% until the user scrolls. Using the bottom-of-viewport normalizes differences in viewport sizes and gives a more accurate picture of content visibility 😄

There’s also the edge case where if the entire page fits within the viewport, we correctly report 100%. But if only 90% fits, a top-of-viewport approach would still report 0% until scrolling occurs, whereas bottom-of-viewport would report 90%, which aligns much better with what the user has actually seen 😄 

Something like:

`
  const scrollTop = window.scrollY || document.documentElement.scrollTop;
  const viewportHeight = window.innerHeight;
  const pageHeight = document.documentElement.scrollHeight;

  const bottomOfViewport = scrollTop + viewportHeight;
  const depth = (bottomOfViewport / pageHeight) * 100;

  return Math.min(100, Math.round(depth));
`

static/analytics.js
```diff
@@ -225,6 +229,9 @@
     // Override pushState to track navigation
     var originalPushState = history.pushState;
     history.pushState = function () {
+      // Send scroll depth event before changing
+      sendScrollDepthEvent();
+
```
Lindharden: I believe we should also send the event on popstate, otherwise we won't be collecting events from users that navigated back/forth in the browser history 😄 

In addition, the scroll depth value currently leaks between pages. For example, if I visit a page that fully fits within the viewport (showing 100% scroll depth), then navigate back to a previous scrollable page and trigger the pagehide event, it still reports 100%. This happens because the max scroll depth isn’t reset when navigating back 😄

dashboard/src/repositories/clickhouse/pages.ts
```diff
@@ -172,16 +172,32 @@ export async function getPageMetrics(
...
+          session_id,
+          url,
+          avg(scroll_depth_percentage) as scroll_depth_percentage
+        FROM analytics.events
+        WHERE site_id = {site_id:String}
+          AND timestamp BETWEEN {start:DateTime} AND {end:DateTime}
+          AND event_type = 'scroll_depth'
+          AND ${SQL.AND(filters)}
+        GROUP BY session_id, url 
+      ),
```
Lindharden: Wouldn’t it make more sense to take the max scroll_depth_percentage per page per session rather than averaging across the session?

For example, if a user scrolls 100% of the page on one visit, then comes back a minute later and scrolls only 0%, averaging would report 50%, even though they’ve actually seen the full page.

Similarly, if a session emits incremental depths like 20%, 40%, 60%, 80%, 100%, the average is 60%, but what we really want to capture is how far they got, which is 100%.

dashboard/src/repositories/clickhouse/pages.ts
```diff
@@ -172,16 +172,32 @@ export async function getPageMetrics(
...
       page_aggregates AS (
         SELECT pvd.path, uniq(pvd.session_id) as visitors, count() as pageviews,
                 avgIf(pvd.duration_seconds, pvd.duration_seconds IS NOT NULL) as avg_time_seconds,
-                countIf(spc.page_count = 1) as single_page_sessions
-        FROM page_view_durations pvd JOIN session_page_counts spc ON pvd.session_id = spc.session_id
+                countIf(spc.page_count = 1) as single_page_sessions,
+                avgIf(psd.scroll_depth_percentage, psd.scroll_depth_percentage IS NOT NULL) as scroll_depth_percentage
+        FROM page_view_durations pvd
+        JOIN session_page_counts spc ON pvd.session_id = spc.session_id
+        JOIN page_scroll_depth psd ON pvd.session_id = psd.session_id AND pvd.path = psd.url
```
Lindharden: This is a no-go 😢 This will cause pages with no scroll depth events to be dropped. We need to change it to a LEFT JOIN and handle nulls, so that we don't drop pages with no scroll depth events.

You can reproduce this by visiting 3 pages with scroll depths enabled. Then comment out the `sendScrollDepthEvent()` function in the analytics script, and then visit 3 other pages. You'll now see 6 pages in the top pages table on the overview page, but in the pages table you'll only see 3 pages, because the remaining 3 with no scroll-depth events are dropped 😄

backend/src/validation/mod.rs
```diff
@@ -207,6 +212,28 @@ impl EventValidator {
...
+            None => return Err(ValidationError::InvalidScrollDepth("Missing scroll_depth_pixels value".to_string())),
+        };
+
+        if *scroll_depth_percentage < 0.0 || *scroll_depth_percentage > 100.0 {
+            return Err(ValidationError::InvalidScrollDepth("scroll_depth_percentage must be between 0 and 100".to_string()));
+        }
+
+        if *scroll_depth_pixels < 0.0 {
+            return Err(ValidationError::InvalidScrollDepth("scroll_depth_pixels must be greater than 0".to_string()));
+        }
```
Lindharden: I believe we should add a max validation on scroll_depth_pixels as well. It is currently possible to send an arbritary large number which causes the column value in Clickhouse to become "Inf"
`2025-10-25T18:18:10.450526Z DEBUG betterlytics::processing: scroll_depth_pixels: Some(inf)`

{DD8A4B0F-2EA5-4135-BF69-8419E29E8CED}

Lindharden (changes requested):

PR #503, opened 2025-09-16
Add log10 transformation to colors-schema on map

Woldbye:
Color map regions by `log10(visitors)`.  We could technically use other logarithm scales, but `log10` is usually the best approach. Another good candidate is the natural log scale, as this actually expresses the spread of IP-addresses quite nicely.

I also filter the top value with a interpolated quantile, to ensure huge outliers, don't take up all the color-scale.

To ensure the maximum values still show a slightly different shade, I have added a flat normalizing increment to the max-value of the d3-colorscale after applying the interpolated quantile.


dashboard/src/hooks/use-color-scales.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { ScaleLinear, ScaleLogarithmic, scaleLinear, scaleLog } from 'd3-scale';
+import { useMemo } from 'react';
+
+export const SCALE_TYPES = ['linear', 'log10', 'log5', 'ln', 'log2'] as const;
+export type ScaleType = (typeof SCALE_TYPES)[number];
+export type ColorScale = ScaleLinear<string, string, never> | ScaleLogarithmic<string, string, never>;
+export type UseColorScalesProps = {
+  maxVisitors: number;
+  scaleType?: ScaleType;
+};
```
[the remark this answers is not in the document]
Woldbye: Linear being used on public :) but the idea is to choose the correct based on data spread

dashboard/src/components/map/MapLegend.tsx
```diff
@@ -10,10 +11,10 @@ function MapLegendComponent({ maxVisitors = 1 }: { maxVisitors?: number }) {
         <div
           className='h-2 w-24 rounded'
           style={{
-            background: `linear-gradient(to right, ${MAP_VISITOR_COLORS.NO_VISITORS} 0%, ${MAP_VISITOR_COLORS.NO_VISITORS} 2%, ${MAP_VISITOR_COLORS.LOW_VISITORS} 3%, ${MAP_VISITOR_COLORS.HIGH_VISITORS} 100%)`,
+            background: `linear-gradient(to right, ${MAP_VISITOR_COLORS.NO_VISITORS} 0%, ${MAP_VISITOR_COLORS.NO_VISITORS} 2%, ${MAP_VISITOR_COLORS.LOW_VISITORS} 3%, ${MAP_VISITOR_COLORS.MEDIUM_VISITORS} 50%, ${MAP_VISITOR_COLORS.HIGH_VISITORS} 100%)`,
```
Woldbye, line 14: NO_VISITORS twice?

dashboard/src/components/map/MapLegend.tsx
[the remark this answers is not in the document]
Woldbye: Good idea - I have added the translated value

dashboard/src/hooks/use-color-scales.tsx
```diff
@@ -0,0 +1,49 @@
...
+    if (scaleType.startsWith('log') || scaleType === 'ln') {
+      return scaleLog<string>()
+        .base(scaleType === 'ln' ? Math.E : parseInt(scaleType.slice(3), 10))
+        .domain([1, Math.max(1, maxVisitors) / 2, Math.max(1, maxVisitors)])
+        .range(range)
+        .clamp(true);
+    }
+
+    return scaleLinear<string>().domain([0, 1, maxVisitors]).range(range).clamp(true);
+  };
```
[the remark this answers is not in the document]
Woldbye: I tested the color scale you proposed extensively. It seems like log10 is a lot better to display small differences in visitor counts. I added the pow-scale as a `ScaleType`, so we can always swap this if you don't like log10 :)

Woldbye:
Bump

dashboard/src/presenters/toWeeklyHeatmapMatrix.ts
```diff
@@ -28,7 +35,10 @@ export function toWeeklyHeatmapMatrix(rows: WeeklyHeatmapRow[]): PresentedWeekly
   const orderedWeekdays: Array<keyof typeof matrix> = [1, 2, 3, 4, 5, 6, 7];
   const resultMatrix = orderedWeekdays.map((key) => ({ weekday: Number(key), hours: matrix[key] }));
 
-  const maxFromRows = Math.max(...rows.map((r) => r.value), 0);
+  const maxFromRows = computeNormalizedMax(
+    rows.map((r) => r.value),
+    hiQuantile,
+  );
```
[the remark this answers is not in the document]
Woldbye: I removed the interpolation things from the PR. Now I just use max

dashboard/src/hooks/use-color-scale.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return useMemo(() => {
+    if (!cssColors || cssColors.length < 3) return null;
+    const range = cssColors.slice(0, 3) as [string, string, string];
+
+    if (scaleType.startsWith('log') || scaleType === 'ln') {
+      const base = scaleType === 'ln' ? Math.E : parseInt(scaleType.slice(3), 10);
+
+      return scaleLog<string>()
+        .base(base)
+        .domain([1, ...colors.slice(1).map((_, i) => Math.max(1, maxVisitors) / (colors.length - i))])
```
flopper123: Question: Why does it slice the first color?

dashboard/src/lib/statistics.ts
```diff
@@ -0,0 +1,51 @@
...
+ * - niceMax(604.3222) → 700
+ * - niceMax(9732) → 10000
+ * - niceMax(0.84) → 1
+ *
+ * @param value The raw maximum value
+ * @returns A rounded "nice" maximum value suitable for human-eyes to read.
+ */
+export function niceMax(value: number): number {
+  return scaleLinear().domain([0, value]).nice().domain()[1];
+}
```
flopper123: All exported functions seems to be unused. So it should probably be remove 🫡

dashboard/src/utils/colorUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+        .map((v) =>
+          Math.max(0, Math.min(255, Math.round(v)))
+            .toString(16)
+            .padStart(2, '0'),
+        )
+        .join('')
+    );
+  }
+
+  return color || raw;
```
flopper123: Question: Is this a valid return? Because in case that this is a valid return, does that make the match case above irrelevant? 😄

flopper123 (changes requested):
Overall a very nice visual update 🔥 (when it works 😅 )
I've added code comments, and have some visual comments:

1. On the landing page, the world map is pure black
image

2. This is not super important, but it seems like there's not "holes" between some of the countries, that changes based on the zoom level:
image

3. The Weekly Trends also has the same issue as the landing page:
image

4. Now that it somehow works again, I'd suggest finding a more muted color for the no visitors on this:
image

It could for instance match the previous one:
image

I can't really figure out when / when it doesn't work, it seems to happen sporadically.

dashboard/src/hooks/use-color-scale.tsx
Woldbye: Its incorrect. Its left-over from when I refactored the empty color type out from it. I have removed the slicing.

dashboard/src/utils/colorUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+      '#' +
+      [r, g, b]
+        .map((v) =>
+          Math.max(0, Math.min(255, Math.round(v)))
+            .toString(16)
+            .padStart(2, '0'),
+        )
+        .join('')
+    );
+  }
```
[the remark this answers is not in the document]
Woldbye: * **1.**  The function is literally named `getResolvedHexColor` - it's specifically meant to return colors in hex format (#RRGGBB), as this is what works best with D3-scales.
* **2 & 3.** I updated the regex to handle rgba() and convert to 8-character hex (#RRGGBBAA) when alpha is present. The previous code was technically functionally correct as rgba() colors passed through and D3 still accepted them, but the function now consistently returns hex as promised. **Note**: using colors with alpha in color scales is generally bad practise.

dashboard/src/hooks/use-color-scale.tsx
```diff
@@ -0,0 +1,58 @@
...
+  | ScaleLinear<string, string, never>
+  | ScaleLogarithmic<string, string, never>
+  | ScalePower<string, string, never>;
+
+export function useColorScale({
+  colors = DEFAULT_COLORS,
+  maxVisitors,
+  scaleType = 'pow-4/10',
+}: UseColorScaleProps): ColorScale | null {
+  const cssColors = useCSSColors(...colors);
```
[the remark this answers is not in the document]
Woldbye: I agree. I did this for 2 reasons:
1. I found it convenient to handle in the hook as it quickly gets verbose to handle in the subscribing components. 
2. Allowed me to use actual css-theme-variables as fallback
3. Best practise to use theme-css-variables, so might aswell enforce it in the API.

Anyways, I have moved it out of the hook as you suggested. I have also refactored it to always returns a non-null colorscale, by fallbacking the resolved css-colors to the given input, in case it fails to resolve it in the DOM.

dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/WeeklyHeatmapSection.tsx
```diff
@@ -127,30 +136,37 @@ type HeatmapGridProps = {
...
     // Monday-first sequence, matching data mapping (1..7 Mon..Sun)
     return Array.from({ length: 7 }, (_, i) => formatter.format(new Date(Date.UTC(1970, 0, 5 + i))));
   }, [locale]);
 
-  const effectiveMax = Math.max(1, maxValue);
+  const colors = useCSSColors([
+    'var(--graph-fill-low)',
+    'var(--graph-fill-medium)',
+    'var(--graph-fill-high)',
+  ] as const);
```
Comment (flopper123) [inline dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/WeeklyHeatmapSection.tsx]: This causes an infinite number of rerenders, as the array is re-instantiated on every rerender. It should be wrapped in a useMemo before being passed to the hook 😄

dashboard/src/utils/colorUtils.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+/**
+ * Get a CSS variable's computed color and convert OKLCH → hex.
+ * Works for both direct color strings and CSS variable references.
+ * @param variableName - The CSS variable name (e.g., '--my-color') or direct color string (e.g., '#ff0000', 'rgb(255, 0, 0)')
+ * @param element - The HTML element to get the CSS variable from (defaults to document.documentElement)
+ * @returns The resolved hex color string (e.g., '#rrggbbaa'), or the original input if not a CSS-variable.
+ */
+export function getResolvedHexColor(
```
flopper123: So I suggest we add constraint to the colors for tailwind for these. If we only allow hex values, and only allow this function to accept tailwind variables, it can be simplified down to:

```ts
export function getResolvedHexColor(
  variableName: `--${string}`
): string {
  if (typeof window === 'undefined') return variableName;

  return getComputedStyle(document.documentElement).getPropertyValue(variableName)
}
```

So I think it's worth it to add constraints for the simplicity of the solution here 😄

dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/WeeklyHeatmapSection.tsx
Comment (Woldbye) [reply dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/WeeklyHeatmapSection.tsx]: I actually noticed this myself, infinite lag spiral.. I pushed a fix that adds a `.join()` on the depedency of the array in the useCssColors hook. This ensures it never rerenders together with the `as const`. 

See [df4a922](https://github.com/betterlytics/betterlytics/pull/503/commits/df4a9223cdb9e171a4791cb38a10b6a5a54d0788)

Lindharden:
@Woldbye @flopper123 what's the state of this? Is it ready to be merged once conflicts are fixed, or were y'all investigating the use of chroma js?

Woldbye:
> @Woldbye @flopper123 what's the state of this? Is it ready to be merged once conflicts are fixed, or were y'all investigating the use of chroma js?

I have assigned @flopper123 to the PR, as he has taken over. He wanted to adjust the colors slightly before this gets merged. 

I have moved this to draft, while he finds the correct colors

PR #511, opened 2025-09-27
Setup custom events + fixed events table

flopper123:


Woldbye (dismissed):
Looks good. Only have one question about the name of one fnc you use

Lindharden (approved):
Looks good 😄

- We should add events to TOS dialog & Banners when both are merged.

I think you'll get some merge conflicts with the replay branch when it's merged, cause that branch also addresses the event time issues

PR #513, opened 2025-09-28
Added mid ellipsis string formatting

flopper123:
Added ellipsis to middle of text.

(Example just to show how it works, it will only do this by default on text above 50 characters)

image

### DIFF
--- a/dashboard/src/utils/formatters.ts
+++ b/dashboard/src/utils/formatters.ts
@@ -35,7 +35,7 @@ export function formatPercentage(num: number, decimalPlaces = 1): string {
 /**
  * Format a string, adds ellipsis to middle of string instead of end
  */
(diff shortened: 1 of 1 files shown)


Woldbye:
Good idea! 😄 glad to see a general solution we can reuse for other components.

It seems like there is still a lot of unused space in the example you provided. Could we make it more responsive with a dom watcher on the screen for the container of the string?

flopper123:
> Good idea! 😄 glad to see a general solution we can reuse for other components.
> 
> It seems like there is still a lot of unused space in the example you provided. Could we make it more responsive with a dom watcher on the screen for the container of the string?

It only looks stupid since I used a way smaller number of characters for the sake of the example 😄 
It'll only start doing it to text above 50 characters by default

Lindharden (approved):

PR #522, opened 2025-10-20
change error message

### LINKED ISSUE #442: Provide better feedback for OAuthAccountNotLinked response
If the user attempts to login via OAuth but an account already exists for the specified mail, an OAuthAccountNotLinked response is returned. This response simply states an error has occured, we should let the user know that an user already exists for this mail.

Image

evanyang1:
Closes #442 

Is this right? I would think that based on the error there should be multiple possible error messages.


evanyang1:
dang i deleted the branch, i'll start from scratch. It has also been a while

PR #526, opened 2025-10-22
166 add filter interactivity on tablescharts

### LINKED ISSUE #166: Add filter interactivity on tables/charts
Where appropriate, tables and charts should be clickable and apply a filter for interactive drilldown.

Lindharden:
Closes #166

### DIFF
--- a/dashboard/src/components/analytics/BrowserTable.tsx
+++ b/dashboard/src/components/analytics/BrowserTable.tsx
@@ -27,7 +27,7 @@ export default function BrowserTable({ data }: BrowserTableProps) {
         <Button
           variant='ghost'
           onClick={() => makeFilterClick('browser')(row.original.browser)}
-          className='cursor-pointer bg-transparent p-0 text-left text-sm font-medium'
+          className='cursor-pointer bg-transparent p-1 text-left text-sm font-medium'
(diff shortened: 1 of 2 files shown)


dashboard/src/hooks/use-filter-click.ts
```diff
@@ -0,0 +1,54 @@
+'use client';
+
+import { useCallback } from 'react';
+import { useQueryFiltersContext } from '@/contexts/QueryFiltersContextProvider';
+import { type FilterColumn, type FilterOperator } from '@/entities/filter';
+
+type Behavior = 'append' | 'replace-same-column' | 'toggle';
```
flopper123: The behaviors `append` and `toggle` does not seem to be used anywhere.
They're probably fine to have, as long as they've been tested.

flopper123 (changes requested):
UX:
Maybe there should be a slight padding-x on the button:
image

dashboard/src/hooks/use-filter-click.ts
Lindharden: I added these to handle cases like clicking a URL where the filter is already applied, the idea was to toggle it in that case. I’ve backed off from that for now until we see how this feature behaves in production.

PR #527, opened 2025-10-22
Env variable for disabling crawling

Lindharden:
This is to prevent our dev server from being crawled and indexed by google


dashboard/src/app/robots.ts
```diff
@@ -4,7 +4,7 @@ import { env } from '@/lib/env';
 export default function robots(): MetadataRoute.Robots {
   const baseUrl = env.PUBLIC_BASE_URL;
 
-  if (!env.IS_CLOUD) {
```
Woldbye: Are we sure it shouldn't be 

env.IS_CLOUD && env.ALLOW_CRAWLING
?

Lindharden: I don't think it changes too much, it's default false and I hope we migrate towards environment based instead of this. So eventually we'll remove this env variable and use !env.ENVIRONMENT = dev

PR #529, opened 2025-10-23
Added search for options for query filters

### LINKED ISSUE #523: Selectable Filter Values for Dashboard Filters
Currently, when applying filters in the dashboard (e.g., by URL, country, device type, etc.), users must manually type the filter value. This can lead to typos, incorrect filters, or uncertainty about which values exist.

Make each filter field (URL, Country, Device Type, etc.) selectable from a dynamic list of available values.

Acceptance Criteria:
- [ ] When selecting a filter type (e.g., Device Type), a dropdown or autocomplete list should appear.
- [ ] The list should show all available values for that dimension (e.g., Desktop, Mobile, Tablet).
- [ ] Users can either type to search or click to select.
- [ ] The list should be dynamically generated based on existing data in the database.

flopper123:
Closes #523 

image

- [x] Add infinite scroll / slice for top x options
- [x] Add arrow down / tab
- [x] Move the "+ "value"" as the final option element
- [ ] Fix values sometimes not being loaded
- [x] Add empty state for no options
- [x] Cursor pointer on options
- [x] Remove border around each individual option
- [x] Add mid-ellipsis to values


Lindharden (changes requested):
Looks pretty good, but I'm not a big fan of the ComboBox. I believe it makes much more sense to make a search input field with a dropdown, instead of having to trigger the dropdown before you can start searching. 

Also, is there a reason not to force filtering to be based on available values?

dashboard/src/components/filters/use-query-filter-search.ts
```diff
@@ -0,0 +1,84 @@
...
+
+type SearchMetadataResult = {
+  shouldUseServerSearch: boolean;
+};
+
+export function useQueryFilterSearch(filter: QueryFilter) {
+  const { startDate, endDate } = useTimeRangeContext();
+  const dashboardId = useDashboardId();
+
+  const [isDirty, setIsDirty] = useState(false);
```
Lindharden: What is meant by isDirty? 😄

dashboard/src/components/filters/QueryFilterInputRow.tsx
```diff
@@ -88,10 +107,22 @@ export function QueryFilterInputRow<TEntity>({
...
-        onChange={(evt) => onFilterUpdate({ ...filter, value: evt.target.value })}
+        onValueChange={(value) => onFilterUpdate({ ...filter, value })}
+        options={options}
+        searchQuery={isDirty ? search : search || filter.value}
+        onSearchChange={setSearch}
+        loading={isLoading}
+        enableSearch={true}
+        emptyState={
+          <div className='text-muted-foreground flex items-center gap-2 p-3 text-sm'>
+            <SearchX className='h-4 w-4' />
```
Lindharden: I think we should remove this icon and just have the text 😄

Lindharden (changes requested):
# Bug:
Seems to work quite nicely, but I found some **issues**:

If you select a value in another filter type and then swap to any other filer type, the value dropdown shows empty even though values should exist. See gif below for how to reproduce:

chrome_I0CBjwdtsw

# UI:
In addition, there's some issues with the **UI**. See the images below:

image

image

# UX:
In terms of **UX**, it would be nice if the tab function prioritizes the dropdown list instead of the delete filter icon. During tested, I already accidentally deleted the filter twice because I hit tab and enter in succession and it deleted my filter instead of selecting the first item in the list. Tab is the most common key to select an element in a list 🔑 

# Responsiveness:
Lastly, it seems to **not** be functional on mobile. I cannot scroll down the list on mobile view 😿

dashboard/src/components/ui/combobox.tsx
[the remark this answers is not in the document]
Lindharden: Neat! Perhaps it makes more sense to rename it BACombobox and move it out of the shadcn folder

PR #543, opened 2025-10-27
Changed to router push for applying url filter + added guard to preve…

### LINKED ISSUE #542: Allow back navigation to remove filters
Currently, when users apply filters on the page (e.g., by project, date, or status) and then navigate back using the browser’s back button, the filters remain active. This can be confusing since users expect the back navigation to restore the previous unfiltered state.

Expected behavior:
- [ ] When users apply filters and then navigate back, the page should restore the previous state (e.g., no filters applied or the prior filter set).
- [ ] The browser history should reflect filter changes.

Lindharden:
…nt url race condition

This PR allows the user to back/forward navigate applied url filters, so that if the user applies a filter, back navigation will remove the filter and vice versa. 

This need became evident when we added click handlers for applying filters to all our tables, as you would often click back in order to remove the filter again, but this would return you to the previous page instead.

Closes #542

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -62,6 +62,7 @@
     "leaflet": "^1.9.4",
     "lucide-react": "^0.503.0",
     "mailersend": "^2.6.0",
+    "moment-timezone": "^0.6.0",
     "motion": "^12.23.12",
     "next": "15.3.3",
     "next-auth": "^4.24.11",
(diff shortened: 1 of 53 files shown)


flopper123 (approved):
Works wonders! 😄

Lindharden:
> Code looks good - i am slightly concerned about the efficiency of the deep sort. Did you ever check how much time the deep keys sort add to the navigations ?

Should be fine. It is typically 0-10 items and flat, so it should not add any overhead at all. In addition, it's only use during encoding 😄

PR #545, opened 2025-10-29
544 use client timezone

flopper123:


dashboard/src/lib/ba-query.ts
```diff
@@ -50,23 +51,65 @@ const granularityIntervalMapper = {
...
+
+  // Note: the end timestamp is exclusive
+  // the BETWEEN keyword seems to be inclusive of the end timestamp
+  const range = safeSql`timestamp >= ${start} AND timestamp < ${end}`;
+
+  // Create the fill
+  const intervalFrom = safeSql`toStartOfInterval(${start}, ${interval}, ${SQL.String({ timezone })})`;
+  const intervalTo = safeSql`toStartOfInterval(${end}, ${interval}, ${SQL.String({ timezone })})`;
+
+  const fill = safeSql`FILL FROM ${intervalFrom} TO ${intervalTo} STEP ${interval}`;
```
Lindharden: I think this statement should include WITH FILL instead of just FILL, seems like FILL is an invalid sql statement, and it may confuse others that WITH needs to be prepended given it's not explicit. Prepending WITH here will complete this statement. Wdyt?

dashboard/src/lib/ba-timerange.ts
```diff
@@ -0,0 +1,409 @@
...
+
+function buildYearRange(
+  main: ResolvedRange,
+  granularity: GranularityRangeValues,
+  timezone: string,
+  compareAlignWeekdays?: boolean,
+) {
+  const unit = getGranularityUnit(granularity);
+
+  if (unit === 'day') {
```
Lindharden: We should probably force unit day here? It seems that I can modify the url to be interval 1y with granularity minute_1 forcing it to query clickhouse with these parameters 😄

dashboard/src/lib/ba-timerange.ts
```diff
@@ -0,0 +1,409 @@
...
+    return { start: alignedStart.toDate(), end: alignedEnd.toDate() };
+  }
+
+  return { start: startTz.toDate(), end: endTz.toDate() };
+}
+
+/**
+ * Map granularity → moment units
+ */
+function getGranularityUnit(g: GranularityRangeValues) {
```
Lindharden: Perhaps rename this function to getMomentGranularityUnit or convertToMomentGranularityUnit

dashboard/src/repositories/clickhouse/visitors.ts
```diff
@@ -90,74 +98,82 @@ export async function getSessionMetrics(
...
-      GROUP BY session_id
-    ),
-    included AS (
+  const queryResponse = timeWrapper(
+    safeSql`
+      WITH windowed_all AS (
+        SELECT session_id, timestamp, event_type
+        FROM analytics.events
+        WHERE site_id = {site_id:String}
+          AND timestamp BETWEEN {start:DateTime} - INTERVAL 30 MINUTE AND {end:DateTime} + INTERVAL 60 MINUTE
```
Lindharden: Now that DateTime is now inclusive in end, does this remain correct? 😄

dashboard/src/presenters/toAreaChart.ts
```diff
@@ -61,6 +62,26 @@ function dataToAreaChart<K extends string>({ dataKey, data, granularity, dateRan
   return chartData;
 }
 
+// New: relies on ClickHouse FILL; do not iterate buckets in JS
```
Lindharden:
```suggestion
```

dashboard/src/presenters/toAreaChart.ts
```diff
@@ -133,11 +148,61 @@ export function toAreaChart<K extends string>({
...
 
-  return {
-    data: dataSeries,
-    comparisonMap,
-    incomplete,
-  };
+  return { data: dataSeries, comparisonMap, incomplete };
+}
+
+// New: relies on ClickHouse-filled buckets, minimal mapping + same compare/incomplete logic
```
Lindharden:
```suggestion
```

dashboard/src/presenters/toAreaChart.ts
```diff
@@ -133,11 +148,61 @@ export function toAreaChart<K extends string>({
...
+export function toNewAreaChart<K extends string>({
```
Lindharden: Same here, should not be named New 😄

dashboard/src/entities/webVitals.ts
```diff
@@ -22,10 +22,10 @@ export type CoreWebVitalRow = z.infer<typeof CoreWebVitalRowSchema>;
...
   date: z.string(),
   name: CoreWebVitalNameSchema,
-  p50: z.number(),
-  p75: z.number(),
-  p90: z.number(),
-  p99: z.number(),
+  p50: z.number().or(z.null()),
+  p75: z.number().or(z.null()),
+  p90: z.number().or(z.null()),
+  p99: z.number().or(z.null()),
```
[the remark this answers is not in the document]
flopper123: When returning from ClickHouse, I think these:
```
        quantileTDigest(0.50)(value) AS p50,
        quantileTDigest(0.75)(value) AS p75,
        quantileTDigest(0.90)(value) AS p90,
        quantileTDigest(0.99)(value) AS p99
```

evaluate to `null` when the bucket is empty.
So without the "or null", the zod validation failed

dashboard/src/lib/ba-timerange.ts
Lindharden: I'm unsure if this was the case before, but I'm thinking that when timerange is 1 year, it should not be possible to execute queries with minute-based granularity 😄

PR #546, opened 2025-10-30
211 reject incoming events for non existent siteids

### LINKED ISSUE #211: Reject incoming events for non-existent siteIds
In the future, we should consider the rejecting events that we do not want to track. 

Some reasons:
Data hygiene: Prevents ClickHouse bloat from "zombie" sites still sending data
Ethically aligned: if a user says "delete me", we stop collecting data from their sites if they haven't removed their script
Trust: reassures privacy-conscious users that deletion means full deactivation
Legal clarity

Implementation ideas:
We can use an in-memory config on the tracking server(s) to store allowed siteIDs. Query postgres for valid siteIds on bootup. We can keep this in-sync through redis PubSub?

Lindharden:
This PR introduces IP blacklisting and domain origin enforcement for event validation, so events from blocked IPs are rejected and only events from the domain specified in the dashboard are accepted. 

{36277E57-993D-4459-AEB4-D7CD3BBB780D}

Key changes include:
- In-memory site-config cache with a TTL of 1 hour and a capacity of 50k, backed by Redis GET on miss with a 150 ms timeout.
- Negative caching for missing site-configs (TTL 60s) to prevent repeated Redis misses.
- Redis Pub/Sub listener on site_cfg_updates with version-aware invalidation and pre-warm refresh.
- Supervisor for the subscription to automatically reconnect and resubscribe if Redis goes down.

Each Redis call is wrapped in a Tokio timeout to prevent workers from being stuck, allowing events to continue flowing in case of Redis issues.

Tracked events are approved when a site-config exists and passes policy checks. If the site-config is missing or a fetch error occurs, events are not blocked but are tagged with a site_config_status (e.g., missing or fetch_error) for later analysis (this means w run a fail-open policy, where events are not rejected if redis is down). Events explicitly rejected due to policy violations, such as blacklisted IPs or invalid domains, continue to be tracked using the existing analytics_events_rejected_total{reason} metric.

Dashboard updates now write the updated site-config to Redis and publish an update event, prompting the Rust server to invalidate and refresh its in-memory cache.

New metrics added for production monitoring include:
- cache_lookups_total{cache="site_config", result="hit|miss|error"}
- site_config_pubsub_healthy (0/1)
- site_config_last_refresh_timestamp_seconds (Unix seconds)
- The existing analytics_events_rejected_total{reason} metric is retained.

This approach ensures resilience and graceful degradation: if Redis or site-config retrieval fails, events continue to be processed, and their status is tracked for observability and future analysis.

Closes #211

### DIFF
--- a/dashboard/messages/da.json
+++ b/dashboard/messages/da.json
@@ -746,8 +746,7 @@
         ]
       },
       "hero": {
-        "titleNonHighlight": "Et Bedre Alternativ til ",
-        "titleHighlight": "Google Analytics",
+        "title": "Et <highlight>Bedre Alternativ</highlight> til Google Analytics",
         "description": "Privatlivsvenlig, cookieløs webanalyse, der respekterer dine besøgende — ingen tracking, ingen cookies, ingen samtykkebannere.",
         "ctaPrimary": "Kom i Gang Gratis",
         "ctaGithub": "Se på GitHub",
--- a/dashboard/messages/en.json
+++ b/dashboard/messages/en.json
@@ -747,8 +747,7 @@
       },
 
       "hero": {
-        "titleNonHighlight": "A Better Alternative to ",
-        "titleHighlight": "Google Analytics",
+        "title": "A <highlight>Better Alternative</highlight> to Google Analytics",
         "description": "Privacy-first, cookieless analytics that respect your visitors — no tracking, no cookies, no consent banners.",
         "ctaPrimary": "Get Started for Free",
         "ctaGithub": "View on GitHub",
--- a/dashboard/messages/it.json
+++ b/dashboard/messages/it.json
@@ -746,8 +746,7 @@
         ]
       },
       "hero": {
-        "titleNonHighlight": "Una Migliore Alternativa a ",
-        "titleHighlight": "Google Analytics",
+        "title": "Una <highlight>Migliore Alternativa</highlight> a Google Analytics",
         "description": "Analytics privacy-first e senza cookie che rispettano i tuoi visitatori — nessun tracciamento, nessun cookie, nessun banner di consenso.",
         "ctaPrimary": "Inizia Gratis",
         "ctaGithub": "Vedi su GitHub",
--- a/dashboard/src/app/[locale]/(public)/(landing)/components/heroSection.tsx
+++ b/dashboard/src/app/[locale]/(public)/(landing)/components/heroSection.tsx
@@ -15,8 +15,9 @@ export async function HeroSection() {
       <div className='relative container mx-auto px-4 sm:px-6 lg:px-8'>
         <div className='text-center'>
           <h1 className='mb-6 text-4xl font-bold tracking-tight sm:text-6xl lg:text-7xl'>
-            {t('titleNonHighlight')}
-            <span className='text-blue-600 dark:text-blue-400'>{t('titleHighlight')}</span>
+            {t.rich('title', {
+              highlight: (chunks) => <span className='text-blue-600 dark:text-blue-400'>{chunks}</span>,
+            })}
           </h1>
           <p className='text-muted-foreground mx-auto mb-10 max-w-4xl text-lg sm:mt-8 sm:text-lg'>
             {t('description')}
--- a/dashboard/src/components/map/LeafletMap.tsx
+++ b/dashboard/src/components/map/LeafletMap.tsx
@@ -72,8 +72,12 @@ export default function LeafletMap({
 
   const worldBounds = useMemo(() => {
     if (!mapComponents?.L) return null;
-    return mapComponents.L.latLngBounds(mapComponents.L.latLng(-100, -220), mapComponents.L.latLng(100, 220));
-  }, [mapComponents]);
+    const hasAntarctica = visitorData.some((d) => d.country_code === 'AQ' && d.visitors);
+    return mapComponents.L.latLngBounds(
+      mapComponents.L.latLng(hasAntarctica ? -100 : -60, -220),
+      mapComponents.L.latLng(100, 220),
+    );
+  }, [mapComponents, visitorData]);
 
   if (isPending || !mapComponents || !worldGeoJson) {
     return (
--- a/dashboard/src/components/map/MapCountryGeoJSON.tsx
+++ b/dashboard/src/components/map/MapCountryGeoJSON.tsx
@@ -46,6 +46,10 @@ export default function MapCountryGeoJSON({
       if (!country_code) return;
 
       let geoVisitor = visitorData.find((d) => d.country_code === country_code);
+      if (!geoVisitor?.visitors && country_code === 'AQ') {
+        layer.setStyle({ opacity: 0, fillOpacity: 0 });
+        return;
+      }
       if (!geoVisitor) {
         geoVisitor = { country_code, visitors: 0 };
       }


backend/src/site_config_cache/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+        }
+
+        debug!(site_id = %site_id, "site-config cache miss; fetching from Redis");
+        let key = format!("{}{}", CONFIG_KEY_PREFIX, site_id);
+
+        let redis = self.redis.read().await.clone();
+        let Some(redis) = redis else {
+            if let Some(m) = &self.metrics { m.increment_cache_lookup(CACHE_NAME, "miss"); }
+            return Ok(None);
+        };
```
flopper123: Does this lookup if Redis is setup?
If so, it should probably handle that case further out to not disrupt the code flow.

backend/src/site_config_cache/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+                        }
+                    };
+
+                    if !self.should_refresh(&site_id, msg_version) {
+                        debug!(site_id = %site_id, "skipping refresh; not a newer version");
+                        continue;
+                    }
+
+                    debug!(site_id = %site_id, "received config update; invalidating cache");
+                    self.invalidate(&site_id);
```
flopper123: Should this also invalidate the negative cache?

backend/src/validation/mod.rs
```diff
@@ -327,6 +335,118 @@ impl EventValidator {
...
+/// Provides a mapping to the ClickHouse Enum8 values
+impl From<SiteConfigStatus> for i8 {
+    fn from(value: SiteConfigStatus) -> Self {
+        match value {
+            SiteConfigStatus::Approved => 1,
+            SiteConfigStatus::Missing => 2,
+            SiteConfigStatus::FetchError => 3,
+        }
+    }
+}
```
flopper123: This mapping should be moved to `backend\src\db\models.rs`, where we're doing the rest of the enum mappings

backend/src/validation/mod.rs
```diff
@@ -327,6 +335,118 @@ impl EventValidator {
...
+    allow_subdomains: bool,
+) -> Result<(), ValidationError> {
+    if let Ok(url) = Url::parse(event_url) {
+        let event_domain = url.host_str().unwrap_or("");
+        let expected_lc = expected_domain.to_ascii_lowercase();
+        let event_lc = event_domain.to_ascii_lowercase();
+
+        let allowed = if allow_subdomains {
+            event_lc == expected_lc || event_lc.ends_with(&format!(".{}", expected_lc))
+        } else {
```
flopper123: I'm probably misunderstading this a little, but if you have `allow_subdomains`, will this `ends_with` allow, say:
`example.io` to equal `betterlytics.io`, as they both ends with `io`?

dashboard/src/entities/dashboardConfig.ts
```diff
@@ -0,0 +1,36 @@
...
+export type DashboardConfig = z.infer<typeof DashboardConfigSchema>;
+export type DashboardConfigCreate = z.infer<typeof DashboardConfigCreateSchema>;
+export type DashboardConfigUpdate = z.infer<typeof DashboardConfigUpdateSchema>;
+
+/**
+ * Applies Zod defaults for DashboardConfigCreate
+ */
+export function withDashboardConfigCreateDefaults(dashboardId: string): DashboardConfigCreate {
+  return DashboardConfigCreateSchema.parse({ dashboardId });
+}
```
flopper123: Why have this? The other times we have a function starting with `with`, it's to signal a wrapper function for our actions.
I would probably just replace the implementing with `DashboardConfigCreateSchema.parse({ dashboardId })` directly.

dashboard/src/repositories/postgres/dashboardConfig.ts
```diff
@@ -0,0 +1,71 @@
...
+    throw new Error('Failed to find dashboard config');
+  }
+}
+
+export async function createDashboardConfig(input: DashboardConfigCreate): Promise<DashboardConfig> {
+  try {
+    const { dashboardId, ...data } = input;
+    const cfg = await prisma.dashboardConfig.create({
+      data: { dashboardId, ...data },
+    });
```
flopper123: Why split them up, just to merge them together?
```suggestion
    const cfg = await prisma.dashboardConfig.create({
      data: input,
    });
```

backend/src/site_config_cache/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+            Some(json) => {
+                match serde_json::from_str::<SiteConfig>(&json) {
+                    Ok(cfg_parsed) => {
+                        let cfg = Arc::new(cfg_parsed);
+                        self.cache.insert(site_id.to_string(), cfg.clone());
+                        self.negative_cache.invalidate(site_id);
+                        if let Some(m) = &self.metrics { m.increment_cache_lookup(CACHE_NAME, "hit"); }
+                        debug!(
+                            site_id = %site_id,
+                            domain = %cfg.domain.as_deref().unwrap_or(""),
```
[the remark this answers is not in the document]
Lindharden: It was because domain was initially not part of the config. I've made it non-optional now. If it, for some reason, isn't to be found in Redis, it'll throw a deserialization error and it'll be marked as a fetch error 😄

backend/src/site_config_cache/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+
+#[derive(Clone)]
+pub struct SiteConfigCache {
+    cache: Cache<String, Arc<SiteConfig>>, // keyed by site_id
+    negative_cache: Cache<String, ()>, // tracks recently missing site_ids
+    redis: Arc<RwLock<Option<Arc<Redis>>>>,
+    metrics: Option<Arc<MetricsCollector>>,
+}
+
+impl SiteConfigCache {
```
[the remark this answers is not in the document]
Lindharden: What do you propose? Doesn't it make sense that this site config cache module instantiates, maintains and updates its own cache? The listener is also strictly related to the site-config channel

dashboard/src/app/actions/siteConfig.ts
```diff
@@ -0,0 +1,17 @@
+'use server';
+
+import { withDashboardAuthContext, withDashboardMutationAuthContext } from '@/auth/auth-actions';
+import { type AuthContext } from '@/entities/authContext';
+import { SiteConfigUpdateSchema, type SiteConfigUpdate } from '@/entities/siteConfig';
+import { getSiteConfig, saveSiteConfig } from '@/services/siteConfig';
+
+export const getSiteConfigAction = withDashboardAuthContext(async (ctx: AuthContext) => {
+  return await getSiteConfig(ctx.dashboardId);
+});
```
flopper123: Question: Should this be blocked from "public" access through demo pages for instance? Since it could theoretically expose things such as blacklisted IPs, if someone from a demo page is able to invoke this action? 🤔

flopper123 (dismissed):

dashboard/src/app/actions/siteConfig.ts
Lindharden: Yes I agree this might be a good idea to block from public 👍

flopper123 (approved):

PR #547, opened 2025-11-01
Demo dashboard

Lindharden:
This PR introduces the demo dashboard and lays the foundation for allowing users to make their own dashboards public for sharing. Supporting user-shared dashboards will only require minimal changes going forward.

Key Changes:
- Demo Dashboard Support, accessible at /public/demo_dashboard_id
- The auth context now short-circuits when viewing a demo dashboard.
- In demo mode, all server actions are wrapped with unstable_cache to prevent excessive database load from high volumes of anonymous visitors.
- To further reduce the number of possible cache keys, several filter and date-range options have been disabled.
- Added visual feedback for disabled options so users understand why certain filters and controls are unavailable in the demo.
- Introduced a dedicated mutation auth context to enforce strict authorization on all mutation actions.

This ensures only authorized users can perform write operations.

These changes make the demo dashboard efficient, safe under heavy traffic, and ready for future public-sharing functionality.

Note: it was not possible to make the public page served using ISR, as filters & timeranges are applied as URL parameters, and these do not create a distinct cached version for ISR. This is why I had to introduce an unstable_cache middleware to reduce public load.

### DIFF
--- a/dashboard/src/app/(protected)/dashboard/[dashboardId]/TrackingScript.tsx
+++ b/dashboard/src/app/(protected)/dashboard/[dashboardId]/TrackingScript.tsx
@@ -17,7 +17,10 @@ export function TrackingScript({ siteId }: TrackingScriptProps) {
     script.setAttribute('data-site-id', siteId);
     script.setAttribute('data-server-url', `${PUBLIC_TRACKING_SERVER_ENDPOINT}/track`);
     script.setAttribute('data-scripts-base-url', PUBLIC_ANALYTICS_BASE_URL);
-    script.setAttribute('data-dynamic-urls', '/dashboard/*/funnels/*,/dashboard/*,/share/*/funnels/*,/share/*');
+    script.setAttribute(
+      'data-dynamic-urls',
+      '/dashboard/*/funnels/*,/dashboard/*,/share/*/funnels/*,/*/share/*/funnels/*,/share/*,/*/share/*',
+    );
     script.setAttribute('data-outbound-links', 'full');
     script.setAttribute('data-web-vitals', 'true');
     script.setAttribute('data-replay', 'true');
--- a/dashboard/src/app/layout.tsx
+++ b/dashboard/src/app/layout.tsx
@@ -51,7 +51,7 @@ export default async function RootLayout({
             src={`${env.PUBLIC_ANALYTICS_BASE_URL}/analytics.js`}
             data-site-id={env.APP_TRACKING_SITE_ID}
             data-server-url={`${env.PUBLIC_TRACKING_SERVER_ENDPOINT}/track`}
-            data-dynamic-urls='/dashboard/*/funnels/*,/dashboard/*,/share/*/funnels/*,/share/*'
+            data-dynamic-urls='/dashboard/*/funnels/*,/dashboard/*,/share/*/funnels/*,/*/share/*/funnels/*,/share/*,/*/share/*'
             data-web-vitals='true'
           />
         )}


dashboard/src/components/TimeRange/CompareRangePicker.tsx
```diff
@@ -81,34 +91,48 @@ export function CompareRangePicker({ className = '' }: { className?: string }) {
...
-        />
-      </div>
+      <DisabledTooltip disabled={isDemo} message={tDemo('notAvailable')}>
+        {(isDisabled) => (
+          <div
+            role='button'
+            tabIndex={0}
+            className='flex w-full items-center justify-between rounded-sm p-2 text-left'
+            onClick={() => {
+              if (isDisabled) return;
```
flopper123: This should also be server validated

dashboard/src/auth/auth-actions.ts
```diff
@@ -23,20 +51,83 @@ export async function requireAuth(): Promise<Session> {
   return session;
 }
 
-export async function requireDashboardAuth() {
-  const session = await requireAuth();
+type ActionRequiringAuthContext<Args extends Array<unknown>, Ret> = (context: AuthContext, ...args: Args) => Ret;
 
-  return session;
+async function resolveDashboardContext(dashboardId: string): Promise<AuthContext> {
```
flopper123: This should probably be cleaner separated

dashboard/src/components/ui/DisabledTooltip.tsx
```diff
@@ -0,0 +1,37 @@
+'use client';
+
+import React from 'react';
+import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/ui/tooltip';
+
+type DisabledTooltipProps = {
+  disabled: boolean;
```
Woldbye: I couldn't find this on shadcn, but its inside the shadcn ui folder. Shouldn't this be moved??

dashboard/src/components/TimeRange/QuickSelectSection.tsx
```diff
@@ -28,15 +31,21 @@ export function QuickSelectSection({ selectedRange, onRangeSelect }: QuickSelect
   const renderButton = (value: Exclude<TimeRangeValue, 'custom'>) => {
     const preset = byValue.get(value);
     if (!preset) return null;
+    const isAllowed = allowedValues ? allowedValues.includes(value) : true;
```
Woldbye:
```suggestion
    const isAllowed = !allowedValues || allowedValues.includes(value);
```

intention is a bit clearer if you write it like this

dashboard/src/components/TimeRange/PrimaryRangePicker.tsx
```diff
@@ -17,13 +17,16 @@ import { DateRangeSection } from '@/components/TimeRange/DateRangeSection';
 import { getAllowedGranularities } from '@/utils/granularityRanges';
 import { formatPrimaryRangeLabel } from '@/utils/formatPrimaryRangeLabel';
 import { LiveIndicator } from '@/components/live-indicator';
+import { useDemoMode } from '@/contexts/DemoModeContextProvider';
 
 export function PrimaryRangePicker({ className = '' }: { className?: string }) {
   const [open, setOpen] = useState(false);
   const isMobile = useIsMobile();
   const t = useTranslations('components.timeRange');
```
Woldbye:
```suggestion
  const t = useTranslations('components');
```

I personally think its cleaner to just import the shared root path, and is more consistent with the rest of the code base.

dashboard/src/components/TimeRange/GranularitySection.tsx
```diff
@@ -28,24 +31,29 @@ export function GranularitySection({
...
+                <Button
+                  type='button'
+                  variant='ghost'
+                  size='sm'
+                  className={cn(
+                    'h-7 w-full cursor-pointer rounded-[4px] border border-transparent px-2 text-[10px] font-medium',
+                    !isAllowed && 'cursor-not-allowed opacity-50 hover:opacity-50',
+                    !isActive && 'text-muted-foreground hover:bg-accent',
+                    isActive &&
+                      'bg-input text-foreground border-border hover:bg-input hover:dark:bg-input shadow-sm',
```
Woldbye:
```suggestion
                    isActive ? 'bg-input text-foreground border-border hover:bg-input hover:dark:bg-input shadow-sm' : 'text-muted-foreground hover:bg-accent'
```

Code reads easier if you use a ternary here

dashboard/src/auth/auth-actions.ts
```diff
@@ -5,8 +5,36 @@ import { authOptions } from '@/lib/auth';
...
+const actionSignatureMap = new WeakMap<AnyFn, string>();
+function hashString(input: string): string {
+  // Simple DJB2 hash, stable and fast for short strings
+  let hash = 5381;
+  for (let i = 0; i < input.length; i++) {
+    hash = (hash << 5) + hash + input.charCodeAt(i);
+    hash |= 0; // force 32-bit
+  }
+  return Math.abs(hash).toString(36);
+}
```
Woldbye: Can we not use the exported one from deviceTypes, instead of adding a new djb2 hash?

dashboard/src/app/[locale]/public/[dashboardId]/campaign/page.tsx
```diff
@@ -0,0 +1 @@
+export { default } from '@/app/(protected)/dashboard/[dashboardId]/campaign/page';
```
Comment (flopper123) [inline dashboard/src/app/[locale]/public/[dashboardId]/campaign/page.tsx]: When re-exporting the pages like this the links in our dashboard will break, which is currently solved by sending the base path to the sidebar, etc., but there may also be some "deeper" issues where we (in the future) may forget to add a page, or so.
I just investigated a little and found this NextJS feature ["rewrites"](https://nextjs.org/docs/app/api-reference/config/next-config-js/rewrites), which seems to be able to potentially solve this, without having to actually re-export the pages, by just routing `/[locale]/public/[dashboardId]` --> `/dashboard/[dashboardId]`
There may of course be some drawbacks from doing this, however, it might be worth looking into if you haven't already 😄

dashboard/src/app/(protected)/dashboard/[dashboardId]/layout.tsx
```diff
@@ -63,37 +57,30 @@ export default async function DashboardLayout({ children, params }: DashboardLay
...
-                <Suspense>
-                  <IntegrationBanner />
-                </Suspense>
-                <div className='flex w-full justify-center'>{children}</div>
-                {mustAcceptTerms && <TermsRequiredModal isOpen={true} />}
-              </BannerProvider>
-            </main>
-            {/* Conditionally render tracking script based on server-side feature flag */}
-            {shouldEnableTracking && siteId && <TrackingScript siteId={siteId} />}
+        <DashboardLayoutShell
```
Comment (flopper123) [inline dashboard/src/app/(protected)/dashboard/[dashboardId]/layout.tsx]: Shouldn't this one also be wrapped in the `DemoModeProvider` but with `isDemo={false}`? 🤔

dashboard/src/auth/auth-actions.ts
```diff
@@ -23,20 +51,83 @@ export async function requireAuth(): Promise<Session> {
...
   action: ActionRequiringAuthContext<Args, Ret>,
 ) {
   return async function (dashboardId: string, ...args: Args): Promise<Awaited<Ret>> {
-    const session = await requireDashboardAuth();
-    const context = await authorizeUserDashboard(session.user.id, dashboardId);
+    const context = await resolveDashboardContext(dashboardId);
+
+    try {
+      // For demo/public dashboards, cache reads to reduce load
+      if (context.userId === 'demo') {
```
flopper123: This should maybe be guarded in a different manner. I'm slightly worried we're discriminating on the `userId` for whether we're in a demo context or not.
We could potentially expand the `AuthContext` with a separate property to use, like `isDemo` for instance.

This is also a design choice we probably have to make. Say you're the owner of a demo dashboard, and you view the dashboard on `/[locale]/public/[dashboardId]`. Do you want to be able to make modifications on that dashboard or not? 🤔 
Because in case you're logged in correctly, this check would fail, even though it's probably a nice idea to let that case be handled here, if that makes sense 😄

dashboard/src/auth/auth-actions.ts
```diff
@@ -23,20 +51,83 @@ export async function requireAuth(): Promise<Session> {
...
+        const actionId = getActionSignature(action as AnyFn);
+        const serializedArgs = JSON.stringify(args, (_key, value) => {
+          if (value instanceof Date) return { __date: value.toISOString() };
+          return value;
+        });
+
+        const cacheKey = ['demo:v4', actionId, context.dashboardId, context.siteId, serializedArgs].join('|');
+
+        return await unstable_cache(async () => action(context, ...args), [cacheKey], { revalidate: 300 })();
```
flopper123: Should the actual `action` call be validated in itself? I'm assuming that there will be some action with some arguments that we don't want to be called by demo users. Right now the mutations are of course blocked, but most of the "blocked" filters / time ranges are currently only validated/blocked client side for instance.

dashboard/src/repositories/postgres/dashboard.ts
```diff
@@ -42,6 +41,20 @@ export async function findUserDashboard(data: DashboardFindByUserData): Promise<
...
+  const rel = await prisma.userDashboard.findFirst({
+    where: { dashboardId: data.dashboardId, userId: data.userId },
+    include: { dashboard: true },
+  });
+
+  if (rel === null) return null;
+
+  return {
+    userDashboard: DashboardUserSchema.parse(rel),
+    dashboard: DashboardSchema.parse(rel.dashboard),
```
flopper123: The naming seems a little off to me, a `userDashboard` sounds like a dashboard to me, but there's also a `dashboard`. Should it maybe be:
```suggestion
    dashboardUser: DashboardUserSchema.parse(rel),
    dashboard: DashboardSchema.parse(rel.dashboard),
```

dashboard/src/services/auth.service.ts
```diff
@@ -112,16 +114,22 @@ export async function registerNewUser(registrationData: RegisterUserData): Promi
   return UserSchema.parse(newUser);
 }
 
-export async function authorizeUserDashboard(userId: string, dashboardId: string): Promise<AuthContext> {
-  const userDashboard = await findUserDashboard({ userId, dashboardId });
-  const dashboard = await findDashboardById(userDashboard.dashboardId);
+export async function getAuthorizedDashboardContextOrNull(data: DashboardFindByUserData) {
```
flopper123: Do we no longer need the `authorizeUserDashboard` for places where we definitely want authorized users, or will this one fully replace the other?

flopper123 (changes requested):
Comments from testing:

1) Some of the links are still pointing to `/dashboard/...`, such as these:
image
It seems like there's a potential security flaw with that, if you're logged in as a different user, and click on it, you actually go to the dashboard, and can click around. I don't think you should be allowed to visit the dashboard though `/dashboard/...` even though it's public.

2) You can add as many filters as you like, by clicking on the different tables:
image

3) You can force the compare to be last year by adding `?compare=year` to the page
image

4) If you're logged in as a user who doesn't own the public dashboard, the public dashboard does not use the language of the user. This is also the case for things like user settings. The user `test@betterlytics.io` here has locale Italia, yet the dropdown/user settings on the public page is still in English:
image

5) Another small language issue is, if I go to `/da/public/...`, and try to then remove the `/da/...`, it'll force me back to `/da/...`. So there should probably be a way to manually change the language of the public dashboard

6) I get this issue in the terminal when refreshing on a public dashboard:
image

dashboard/src/auth/auth-actions.ts
Lindharden: Are you referring to the method body? 😄

dashboard/src/auth/auth-actions.ts
```diff
@@ -23,20 +51,83 @@ export async function requireAuth(): Promise<Session> {
...
+    try {
+      // For demo/public dashboards, cache reads to reduce load
+      if (context.userId === 'demo') {
+        const actionId = getActionSignature(action as AnyFn);
+        const serializedArgs = JSON.stringify(args, (_key, value) => {
+          if (value instanceof Date) return { __date: value.toISOString() };
+          return value;
+        });
+
+        const cacheKey = ['demo:v4', actionId, context.dashboardId, context.siteId, serializedArgs].join('|');
```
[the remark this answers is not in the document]
Lindharden: demo:v4 was due to internal conflicts during development. I've reverted it to v1 😄 

I have hashed serializedArgs now as well, but I don't see why we would want to dash dashboardId and siteId 👍

dashboard/src/app/[locale]/public/[dashboardId]/layout.tsx
```diff
@@ -0,0 +1,49 @@
...
+  return (
+    <PublicEnvironmentVariablesProvider publicEnvironmentVariables={publicEnvironmentVariables}>
+      <DashboardProvider>
+        <DemoModeProvider isDemo={true}>
+          <DashboardLayoutShell
+            dashboardId={dashboardId}
+            isDemo={true}
+            basePath={`/${locale}/public`}
+            trackingSiteId={shouldEnableTracking && siteId ? siteId : null}
+            includeIntegrationManager={false}
```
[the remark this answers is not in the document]
Comment (Lindharden) [reply dashboard/src/app/[locale]/public/[dashboardId]/layout.tsx]: How do you propose we do this, when layout is a server-component and contexts are consumed in client-components? 😄

dashboard/src/app/(protected)/dashboard/[dashboardId]/layout.tsx
Comment (Lindharden) [reply dashboard/src/app/(protected)/dashboard/[dashboardId]/layout.tsx]: The context default is already false, so every consumser will see it as **not demo**. There's no need to wrap 👍

dashboard/src/auth/auth-actions.ts
Lindharden: I already made this design choice when I made this ticket, it is not possible to make modifications on a public dashboard and it should not be possible to do so. The public dashboard is solely for display purposes.

Good suggestion expanding the AuthContext with a property for isDemo. I've done this now 😄

dashboard/src/auth/auth-actions.ts
Lindharden: Care to elaborate on intent and what you propose?

dashboard/src/contexts/DemoModeContextProvider.tsx
```diff
@@ -0,0 +1,19 @@
...
+};
+
+export function DemoModeProvider({ isDemo, children }: DemoModeProviderProps) {
+  const value = useMemo(() => isDemo, [isDemo]);
+  return <DemoModeContext.Provider value={value}>{children}</DemoModeContext.Provider>;
+}
+
+export function useDemoMode(): boolean {
+  return useContext(DemoModeContext);
+}
```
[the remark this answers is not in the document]
Lindharden: createContext<boolean>(false) makes the default value false, so useDemoMode() will never return null. If a component isn’t wrapped by DemoModeProvider, it just receives false, which is the safe “non-demo” fallback. Hence there’s no risk of demo-only UI leaking; adding an extra check for null wouldn’t change behavior 😄

dashboard/src/repositories/postgres/dashboard.ts
Lindharden: Agree. I have made changes to the code, but the schema is named userDashboard, so this entails adding a migration for this rename. Should not be part of this issue.

dashboard/src/services/auth.service.ts
Lindharden: Care to elaborate what you mean and where you're referencing? 😄

Lindharden:
> Comments from testing:
> 
> 1. Some of the links are still pointing to `/dashboard/...`, such as these:
> It seems like there's a potential security flaw with that, if you're logged in as a different user, and click on it, you actually go to the dashboard, and can click around. I don't think you should be allowed to visit the dashboard though `/dashboard/...` even though it's public.

I have created a navigation provider to solve this in a more robust manner. All links now properly resolve to either /public or /dashboard. 

I did not prevent access to /dashboard because you cannot perform any actions, so it's the same as /public, and it's only for our dashboard. But I'll see if we can efficiently restrict it - shouldn't matter either way.

> 2. You can add as many filters as you like, by clicking on the different tables:

Good find. I have limited this now and added a feedback toast for UX.

> 3. You can force the compare to be last year by adding `?compare=year` to the page

As we discussed in person as well, this isn't something we really want to bother fixing as this requires adversaries to deliberately modify the url. The limitation of these filters & time ranges is to prevent the main flood of non-authenticated users on the landing page from triggering too many distinct db queries.

> 4. If you're logged in as a user who doesn't own the public dashboard, the public dashboard does not use the language of the user. This is also the case for things like user settings. The user `test@betterlytics.io` here has locale Italia, yet the dropdown/user settings on the public page is still in English:

Dashboard uses public locale, so that's expected 👍

> 5. Another small language issue is, if I go to `/da/public/...`, and try to then remove the `/da/...`, it'll force me back to `/da/...`. So there should probably be a way to manually change the language of the public dashboard

Does not relate to this PR. It's intended and has always been like this. If you want to explicitly force it to english this way, you have to write /en in the url

> 6. I get this issue in the terminal when refreshing on a public dashboard:

I have added localized SEO configs and Robots policy for the public page now - will fix alongside this 👍

flopper123 (dismissed):

PR #550, opened 2025-11-07
Geography Map Tooltip/Popup with comparison data

### LINKED ISSUE #269: Add comparison to world map tooltip
When hovering a country in the world map, we should be able to display number of visitors for the compared range

Woldbye:
Closes #269 

Improves geography map interactions by differentiating hover tooltips from click popups in style, adding comparison data visualization, and fixing
tooltip bugs.

### Changes
  - Large maps now show rich popup with comparison data and time ranges when clicking countries
  - Small maps continue using simple tooltip for both hover and click
  - Fixed initial tooltip visibility and hover conflicts with popups
  - **New**: MapPopupContent component for large map popups with enhanced styling and comparison data
  - **Updated**: Documentation to reflect changes to geography selection
  - **Updated**: use-leaflet-style to add rounded corners to tooltips and fix pointer events
  - **Updated**: Tooltip components to accept onMouseEnter callback for clearing hover state
  - **Fixed**: Tooltip appearing before user interaction
  - **Fixed**: Hover tooltip not hiding when hovering over click popup
  - **Optimization**: Split selection context into two to reduce rerenders. This optimization alone prevents the geojson from rerendering when a new country is hovered.

### Demo
msedge_WeyF93JwtS

msedge_RZLNDe7KmG

image

### DIFF
--- a/dashboard/src/components/map/tooltip/MapPopupContent.tsx
+++ b/dashboard/src/components/map/tooltip/MapPopupContent.tsx
@@ -2,7 +2,6 @@
 
 import { FlagIconProps } from '@/components/icons';
 import { CountryDisplay } from '@/components/language/CountryDisplay';
-import { TrendIndicator } from '@/components/TrendIndicator';
 import { TrendPercentage } from '@/components/TrendPercentage';
 import { SupportedLanguages } from '@/constants/i18n';
 import type { GeoVisitorWithCompare } from '@/entities/geography';
@@ -87,8 +86,6 @@ function MapPopupContentComponent({
         <span className='text-muted-foreground'>{t('geography.visitors')}</span>
         {geoVisitor.compareVisitors !== undefined && percentageChange !== undefined ? (
           <TrendPercentage percentage={percentageChange} withParenthesis withIcon />
-        ) : geoVisitor.compareVisitors === 0 && geoVisitor.visitors > 0 ? (
-          <TrendIndicator percentage={1} />
         ) : (
           <span className='text-foreground ml-1'>{formatNumber(geoVisitor.visitors)}</span>
         )}
--- a/dashboard/src/components/map/tooltip/MapTooltipContent.tsx
+++ b/dashboard/src/components/map/tooltip/MapTooltipContent.tsx
@@ -1,6 +1,5 @@
 import { FlagIconProps } from '@/components/icons';
 import { CountryDisplay } from '@/components/language/CountryDisplay';
-import { TrendIndicator } from '@/components/TrendIndicator';
 import { TrendPercentage } from '@/components/TrendPercentage';
 import { SupportedLanguages } from '@/constants/i18n';
 import type { GeoVisitorWithCompare } from '@/entities/geography';
@@ -46,11 +45,9 @@ function MapTooltipContent({ geoVisitor, size, className, label, locale, onMouse
           <span className='text-muted-foreground'>{label}:</span>
           <div className='text-foreground flex flex-row gap-1'>
             <span>{formatNumber(geoVisitor.visitors)}</span>
-            {geoVisitor.compareVisitors !== undefined && percentageChange !== undefined ? (
+            {geoVisitor.compareVisitors !== undefined && percentageChange !== undefined && (
               <TrendPercentage percentage={percentageChange} withParenthesis={true} withIcon={true} />
-            ) : geoVisitor.compareVisitors === 0 && geoVisitor.visitors > 0 ? (
-              <TrendIndicator percentage={1} />
-            ) : null}
+            )}
           </div>
         </div>
       </div>


dashboard/src/app/actions/geography.ts
```diff
@@ -35,11 +35,19 @@ export const getWorldMapDataAlpha2 = withDashboardAuthContext(
...
-      return worldMapResponseSchema.parse(dataToWorldMap(geoVisitors, CountryCodeFormat.Original));
+
+      const compareGeoVisitors =
+        compareStartDate &&
+        compareEndDate &&
+        (await fetchVisitorsByGeography(ctx.siteId, compareStartDate, compareEndDate, queryFilters));
+
+      return worldMapResponseSchema.parse(
+        dataToWorldMap(geoVisitors, compareGeoVisitors ?? [], CountryCodeFormat.Original),
+      );
```
Lindharden:
```suggestion
      return dataToWorldMap(geoVisitors, compareGeoVisitors ?? [], CountryCodeFormat.Original);
```

The presenter layer already outputs the data as a WorldMapResponse, so there's no need to parse it here 😄

dashboard/src/components/map/popup/MapPopupContent.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { useTranslations } from 'next-intl';
+import React from 'react';
+
+export type MapPopupContentProps = {
+  geoVisitor?: GeoVisitorWithCompare;
+  className?: string;
+  locale: SupportedLanguages;
+  size?: 'sm' | 'lg';
+  t: ReturnType<typeof useTranslations<'components'>>;
+  timeRangeCtx: TimeRangeContextProps;
```
Lindharden: Can't we just consume the time range context instead of passing it as a prop? This seems to slightly deviate from the purpose of Contexts, which is designed to eliminate "prop drilling". 😄

dashboard/src/app/actions/geography.ts
Woldbye: Great, I removed the parse call and the schema import

dashboard/src/components/map/popup/MapPopupContent.tsx
```diff
@@ -0,0 +1,120 @@
...
+    value: React.ReactNode;
+    tooltip?: string;
+    muted?: boolean;
+  }) => (
+    <div className='flex items-center justify-between gap-2'>
+      <div className='flex items-center gap-2' title={tooltip}>
+        {color && <div className={`${color} h-2 w-2 rounded-full`} />}
+        <span className={muted ? 'text-muted-foreground' : ''}>{label}</span>
+      </div>
+      <span className={cn('font-medium', muted && 'text-muted-foreground')}>{value}</span>
```
flopper123: Quibble: Use `cn` for all the `classNames`
```suggestion
        {color && <div className={cn(color, 'h-2 w-2 rounded-full')} />}
        <span className={cn({ 'text-muted-foreground': muted })}>{label}</span>
      </div>
      <span className={cn('font-medium', muted && 'text-muted-foreground')}>{value}</span>
```

dashboard/src/components/map/popup/MapPopupContent.tsx
```diff
@@ -0,0 +1,120 @@
...
+                    interval: timeRangeCtx.interval,
+                    offset: timeRangeCtx.offset,
+                    startDate: timeRangeCtx.startDate,
+                    endDate: timeRangeCtx.endDate,
+                  })
+            }
+            tooltip={`${timeRangeCtx.startDate.toLocaleString()} - ${timeRangeCtx.endDate.toLocaleString()}`}
+            value={formatNumber(geoVisitor.visitors)}
+          />
+          {timeRangeCtx.compareStartDate && (
```
flopper123: Quibble: Potentially also check that `timeRangeCtx.compareEndDate` is defined 😄

dashboard/src/presenters/toWorldMap.ts
```diff
@@ -42,8 +47,19 @@ export function dataToWorldMap(data: GeoVisitor[], format: CountryCodeFormat): W
...
 
+  const processedCompareData = compareData.map((visitor) => {
+    const transformedData = transformerFunction(visitor.country_code);
+    return transformedData
+      ? {
+          ...visitor,
+          country_code: transformedData,
+        }
+      : visitor;
+  });
```
flopper123: I'm a little confused by this 😄 Why can the `transformerFunction` return something nullish? And doesn't this result in an array of data with formats of country codes?

dashboard/src/entities/geography.ts
```diff
@@ -7,8 +7,21 @@ export const GeoVisitorSchema = z.object({
...
+  compareData: z.array(GeoVisitorSchema),
   maxVisitors: z.number(),
 });
 
 export type GeoVisitor = z.infer<typeof GeoVisitorSchema>;
 export type WorldMapResponse = z.infer<typeof worldMapResponseSchema>;
+
+export type GeoVisitorComparison = {
+  compareVisitors?: number;
+  compareDate?: Date;
```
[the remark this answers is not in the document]
Woldbye: I removed it. Its legacy code from the map timeseries I imported this code from :D

dashboard/src/components/map/tooltip/MapStickyTooltip.tsx
```diff
@@ -24,26 +24,38 @@ export default function MapStickyTooltip({ size = 'sm' }: MapStickyTooltip) {
...
 
     return () => {
-      mapContainer.removeEventListener('mousemove', onMouseMove);
+      mapContainer.removeEventListener('mousemove', updateMousePosition);
     };
-  }, [map, selectedFeature]);
+  }, [map, selectedFeature, hoveredFeature]);
+
+  const hasValidPosition =
+    hoveredFeature?.mousePosition || latestMouseRef.current.x !== 0 || latestMouseRef.current.y !== 0;
```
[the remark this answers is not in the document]
Woldbye: The position logic handles the edge case where the user's mouse is already hovering over a country when the page loads (before any mousemove event fires). The latestMouseRef check prevents the tooltip from briefly flashing at the top-left corner {0, 0} when a new country is hovered

dashboard/src/presenters/toWorldMap.ts
flopper123: Perfect, then it's all good 😎👌

Woldbye:
@flopper123 

1. Fixed the root cause so it works correctly everywhere (See [37ca223](https://github.com/betterlytics/betterlytics/pull/550/commits/37ca223dfff3f7ea95fa607977826068d356dbff))
2.   Copied the trend behavior from other tooltips for consistency:
  * previous = 0, current > 0: No trend shown
  * previous = 0, current = 0: Shows `-`
  * previous > 0, current = 0: Shows percentage decrease
  * previous > 0, current > 0: Shows percentage change

PR #564, opened 2025-11-21
Bug Report functionality

Lindharden:


dashboard/src/repositories/postgres/bugReports.ts
```diff
@@ -0,0 +1,26 @@
...
+} from '@/entities/bugReport';
+
+export async function createBugReport(data: BugReportCreate): Promise<BugReport> {
+  try {
+    const payload = BugReportCreateSchema.parse(data);
+
+    const report = await prisma.bugReport.create({
+      data: {
+        userId: payload.userId,
+        dashboardId: payload.dashboardId ?? null,
```
flopper123: Since this one comes from the auth context, is there ever a case where it can be nullish? 😄

flopper123 (approved):

PR #566, opened 2025-11-22
Changelog + Changelog modal

Lindharden:
This PR introduces a "What's New" modal popup accessible from the topbar. It is non-intrusive and shows a "New" pill when a new release of the modal is available and the user hasn’t seen it yet.

Additionally, this PR adds a public /changelog endpoint exposing our full changelog in all supported locales.

To manage translations more efficiently and prevent an excessive number of language keys in the main language files, each changelog-entry now has its own dedicated TSX file per locale. This approach simplifies maintenance across languages, and existing translations were auto-generated with ChatGPT.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -21,6 +21,11 @@ POSTGRES_DB=dashboard
 POSTGRES_USER=user
 POSTGRES_PASSWORD=password
 
+# Postgres database - SiteConfig table read-only user
+SITE_CONFIG_DATABASE_URL="postgresql://siteconfig_ro:CHANGE_ME_STRONG@localhost:5432/dashboard"
+POSTGRES_SITECONFIG_RO_PASSWORD=CHANGE_ME_STRONG
+
+
 
 
 #####################################
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -39,10 +39,27 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a23eb6b1614318a8071c9b2521f36b424b2c83db5eb3a0fead4a6c0809af6e61"
 
 [[package]]
+name = "arc-swap"
+version = "1.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "69f7f8c3906b62b754cd5326047894316021dcfe5a194c8ea52bdd94934a3457"
+
+[[package]]
 name = "arraydeque"
 version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7d902e3d592a523def97af8f317b08ce16b7ab854c1985a0c671e6f15cebc236"
+
+[[package]]
+name = "async-lock"
+version = "3.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5fd03604047cee9b6ce9de9f70c6cd540a0520c813cbd49bae61f33ab80ed1dc"
+dependencies = [
+ "event-listener",
+ "event-listener-strategy",
+ "pin-project-lite",
+]
 
 [[package]]
 name = "async-trait"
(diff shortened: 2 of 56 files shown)


dashboard/src/components/whats-new/WhatsNewModal.tsx
```diff
@@ -0,0 +1,149 @@
...
+                'text-foreground hover:bg-muted/60 relative h-10 w-10 rounded-full transition-colors',
+                isUnread && 'bg-primary/10 hover:bg-primary/20',
+              )}
+              aria-label={`View what's new in version ${metadata.version}`}
+              onClick={() => setIsOpen(true)}
+            >
+              <Sparkles className='text-foreground h-4 w-4' />
+              {isUnread && (
+                <span className='bg-primary pointer-events-none absolute -top-1.5 right-0 flex translate-x-1/2 items-center rounded-full px-1.5 py-0.5 text-[0.55rem] font-semibold tracking-[0.25em] text-white uppercase shadow-lg'>
+                  New
```
Woldbye: I would suggest localizing this! :)

dashboard/src/components/whats-new/WhatsNewModal.tsx
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  return <WhatsNewModalWithDisplay currentWhatsNewModalDisplay={currentWhatsNewModalDisplay} />;
+}
+
+type WhatsNewModalWithDisplayProps = {
+  currentWhatsNewModalDisplay: WhatsNewEntry;
+};
+
+function WhatsNewModalWithDisplay({ currentWhatsNewModalDisplay }: WhatsNewModalWithDisplayProps) {
```
Woldbye: General note:

I think the code readability would increase a lot if you just called this a change-log. I am sure you considered this, so I am curious as to why you haven't ?

dashboard/prisma/migrations/20251122110008_whats_new_versioning/migration.sql
```diff
@@ -0,0 +1,3 @@
+-- AlterTable
+ALTER TABLE "User" ALTER COLUMN "changelogVersionSeen" SET DEFAULT 'v0',
```
Woldbye: If default is v0, wouldn't this mean everyone has the initial change log seen?

dashboard/src/components/changelog/ChangelogEntryCard.tsx
```diff
@@ -0,0 +1,39 @@
...
+export function ChangelogEntryCard({ entry, locale }: ChangelogEntryCardProps) {
+  const releaseDate = new Date(entry.releasedAt);
+  const isValidDate = !Number.isNaN(releaseDate.getTime());
+  const formattedReleaseDate = isValidDate
+    ? new Intl.DateTimeFormat(locale, { month: 'short', day: 'numeric', year: 'numeric' }).format(releaseDate)
+    : entry.releasedAt;
+
+  return (
+    <section className='border-border/60 border-t py-10 first:border-t-0 first:pt-8 last:border-b'>
+      <div className='flex flex-col gap-6 md:grid md:grid-cols-[160px_1fr] md:items-start md:gap-12'>
```
Woldbye: In my experience the "fr" grid utility behave inconsistently  across mobile browsers :/ we had issues at work where it didn't apply at all, or applied occasionally...

Maybe we can find another way to declare this ?

dashboard/src/components/changelog/ChangelogFeed.tsx
Woldbye: Clean component, but I would love to see the intersection logic moved out and generalized. Trust me, you are not the last developer that need this 

Something like this comes to mind:

```ts
export function useInView<T extends HTMLElement>({
  rootMargin = '0px'
}: { rootMargin?: string } = {}) {
  const [ref, setRef] = useState<T | null>(null);
  const [inView, setInView] = useState(false);

  useEffect(() => {
    if (!ref) return;

    const observer = new IntersectionObserver(
      ([entry]) => setInView(entry.isIntersecting),
      { rootMargin }
    );

    observer.observe(ref);
    return () => observer.disconnect();
  }, [ref, rootMargin]);

  return { ref: setRef, inView };
}

```

Woldbye (changes requested):
Nice feature. I think general code quality is high.

Regarding the localization of the change logs:

I would highly suggest finding a different solution that leans on our existing i18n setup. You essentially replace “too many language keys” with:
•	duplicated components
•	duplicated bound strings
•	duplicated JSX
•	duplicated layout
•	duplicated metadata
•	duplicated markup

Instead we could use modular i18n imports. I know it's not simple to implement correctly, but I think it's worth the effort here.. If this causes issues for you, I would be happy to help implement it

dashboard/src/components/changelogModal/ChangelogModal.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+            <DialogFooter className='border-border/60 bg-muted/30 text-muted-foreground flex flex-col gap-3 rounded-b-[24px] border-t px-5 py-5 text-left text-[0.7rem] sm:flex-row sm:items-center sm:justify-between md:px-6'>
+              <p className='text-muted-foreground/80 max-w-sm text-[0.7rem] leading-snug sm:max-w-md'>
+                {t('haveFeedback')}
+              </p>
+              <div className='flex flex-col gap-2 sm:flex-row'>
+                <Button variant='ghost' asChild className='w-full cursor-pointer text-sm sm:w-auto'>
+                  <Link href='/changelog'>{t('viewAll')}</Link>
+                </Button>
+                <Button className='w-full cursor-pointer text-sm sm:w-auto' onClick={() => setIsOpen(false)}>
```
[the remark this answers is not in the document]
Lindharden: The setIsOpen is pure, so should be fine 😄

Issue #498, opened 2025-09-15
Respect user’s preferred language on public pages when session exists


Lindharden:
Currently, public pages don’t automatically reflect the user’s preferred language, even if they have an active session. Update the logic so that when a logged-in user visits a public page, the interface language is set according to their saved preferences. This ensures a consistent multilingual experience across both private and public sections.

PR #575, opened 2025-11-29
Change authorization to use database sessions

Woldbye:
Should not be merged until #498 is merged

Implements database sessions for NextAuth, replacing the previous JWT-based session strategy. This provides:
* Can instant session revocate
* Can query active sessions per user
* Sessions validated against database on each request

### Changes

* Switch session strategy from jwt to database
* Add workaround for NextAuth's credentials provider limitation (doesn't support database sessions out-of-box)
* Manually create sessions in signIn callback for credentials login
* Override jwt.encode/jwt.decode to handle session tokens
* Add clean-up to automatically remove expired session from DB

### Technical notes

NextAuth's CredentialsProvider doesn't natively support database sessions. We work around this by:
1. Creating the session record manually in the signIn callback
2. Setting the session cookie ourselves
3. Overriding JWT encode/decode to pass through the session token

This causes additional database queries for each distinct session resolution, which means we need to be thoughtful about resolving session. The PR addresses this implication as well.


dashboard/src/repositories/postgres/session.ts
```diff
@@ -0,0 +1,23 @@
...
+      where: {
+        expires: {
+          lt: new Date(),
+        },
+      },
+    });
+  } catch (error) {
+    console.error('Error deleting expired sessions:', error);
+  }
+}
```
Lindharden: I don't think we need this yet. Its' only usecase seems to be to reduce DB bloat for expired sessions, but there'll be a minimal number of expired sessions because it takes 30 days of inactivity to be considered expired.

I think we're better off removing this now and reintroducing when we see bloat stack up - and adding it to a background job instead 😄

dashboard/src/lib/auth.ts
```diff
@@ -77,102 +87,75 @@ export const authOptions: NextAuthOptions = {
...
+        token.uid = user.id;
       }
+      return token;
+    },
+
+    async session({ session, user }) {
+      const freshUser = await findUserByEmail(user.email!);
+      if (!freshUser || !session.user) return session;
+
+      const settings = await getUserSettings(freshUser.id);
```
Lindharden: I think we should add an option to make findUsreByEmail include settings in a single join query - otherwise this introduces 2 distinct query roundtrips every time we retrieve the session. We need to be careful about this when using database strategy, simply because every time we request the user session (which we do a lot) it does 2 queries

Lindharden:
I have further moved session-related business logic into a dedicated session service file. I removed the cleanup of stale sessions from the login function, as this added a lot of unnecessary overhead during each login. Stale sessions adds bloat **very** slowly in the database, given that each session only becomes stale after 30 days of inactivity. 

When we introduce a persistent background-job container, we'll move session cleanup into here 😄

PR #579, opened 2025-11-30
Resolve bug with settings changing instantly without save

### LINKED ISSUE #438: Changing settings in the settings modal changes settings instantly
When changing settings in the settings modal, it changes the settings of the app without having to save.  However, on refresh it turns back to your previously saved settings.

Image

Woldbye:
Closes #438 

The core issue that causes the bug is that the Avatar and Theme settings has side-effects in the handle method, that applies it instnatly when its changed.

To resolve the issue, I have added a `SettingsEffectsContext`, which ensures the actual effect of changing the setting is deferred until user clicks the `Save` button.  

### Demo
msedge_UeBfoPWPhP

### DIFF
--- a/dashboard/src/components/userSettings/UserSettingsDialog.tsx
+++ b/dashboard/src/components/userSettings/UserSettingsDialog.tsx
@@ -2,7 +2,6 @@
 
 import { useEffect, useMemo, useState } from 'react';
 import { Dialog, DialogContent, DialogDescription, DialogHeader, DialogTitle } from '@/components/ui/dialog';
-import { VisuallyHidden } from '@radix-ui/react-visually-hidden';
 import { Button } from '@/components/ui/button';
 import { Tabs, TabsContent, TabsList, TabsTrigger } from '@/components/ui/tabs';
 import { Settings, Shield, AlertTriangle, Loader2, Save, BarChart3, Receipt, User } from 'lucide-react';
@@ -49,9 +48,7 @@ export default function UserSettingsDialog({ open, onOpenChange }: UserSettingsD
     return (
       <Dialog open={open} onOpenChange={onOpenChange}>
         <DialogContent className='sm:max-w-[800px] md:min-w-[700px] lg:min-w-[900px]'>
-          <VisuallyHidden asChild>
-            <DialogTitle>{tDialog('title')}</DialogTitle>
-          </VisuallyHidden>
+          <DialogTitle>{tDialog('title')}</DialogTitle>
           <div className='flex flex-col items-center justify-center space-y-3 py-16'>
             <Spinner />
             <p className='text-muted-foreground text-sm'>{tDialog('loading')}</p>
@@ -65,9 +62,7 @@ export default function UserSettingsDialog({ open, onOpenChange }: UserSettingsD
     return (
       <Dialog open={open} onOpenChange={onOpenChange}>
         <DialogContent className='sm:max-w-[800px] md:min-w-[700px] lg:min-w-[900px]'>
-          <VisuallyHidden asChild>
-            <DialogTitle>{tDialog('title')}</DialogTitle>
-          </VisuallyHidden>
+          <DialogTitle>{tDialog('title')}</DialogTitle>
           <div className='flex flex-col items-center justify-center space-y-3 py-16'>
             <AlertTriangle className='text-destructive h-8 w-8' />
             <div className='text-center'>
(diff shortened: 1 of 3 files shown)


dashboard/src/contexts/SettingsEffectsContext.tsx
```diff
@@ -0,0 +1,52 @@
+'use client';
+
+import React, { createContext, useContext, useCallback, useRef } from 'react';
+import type { UserSettingsUpdate } from '@/entities/userSettings';
+
+type SettingsEffect = {
+  apply: (settings: UserSettingsUpdate) => void;
+};
```
flopper123: Usually with event emitters they are defined as dispatchers, so I suggest making the type of SettingsEffect a `Dispatch<SettingsEffect>`.

```suggestion
type SettingsEffect = Dispatch<SettingsEffect>;
```

flopper123 (changes requested):
Works very nicely. Had some code comments though 😄

Woldbye:
> @Woldbye it would be nice if it's possible to make certain settings take effect temporarily. That is, if the user changes language or theme, I'd expect them to take effect immediately. But if the user then clicks cancel, I'd expect it to return back to what it was

Sadly doesn't seem feasible to implement this smoothly for language, as it requires a full page refresh. I recommend keeping language out of this for now.

I have implemented a general way to register preview effects (such as Theme). I have ensured to roll the preview back, if the modal is unexpectedly closed without saving, or user navigates.

dashboard/src/components/userSettings/UserSettingsDialog.tsx
```diff
@@ -40,6 +41,78 @@ interface UserSettingsTabConfig {
...
+    return (
+      <Dialog open={open} onOpenChange={onOpenChange}>
+        <DialogContent className='sm:max-w-[800px] md:min-w-[700px] lg:min-w-[900px]'>
+          <div className='flex items-center justify-center py-8'>
+            <span>{tDialog('noSettings')}</span>
+          </div>
+        </DialogContent>
+      </Dialog>
+    );
+  }
```
flopper123: I get this error from these:
image

I think they just need to have a:
`<DialogTitle>`

As the message says, you can wrap them in a `VisuallyHidden`, such that they're not visible 😄

dashboard/src/components/userSettings/UserSettingsDialog.tsx
```diff
@@ -86,23 +159,28 @@ export default function UserSettingsDialog({ open, onOpenChange }: UserSettingsD
...
-    if (settings) {
+    if (open) {
       setFormData({ ...settings });
     }
-  }, [settings]);
+  }, [settings, open]);
+
+  useEffect(() => {
+    return () => restore();
+  }, [restore]);
```
flopper123: This is an anti-pattern 😄 
I just tested it having removed this, and it seems to fully work as expected. I think it's due to also being in the `handleCloseDialog`
```suggestion
```

dashboard/src/contexts/UserSettingsPreviewContext.tsx
```diff
@@ -0,0 +1,54 @@
...
+    handlersRef.current.delete(key);
+  }, []);
+
+  const preview = useCallback((settings: UserSettingsUpdate) => {
+    handlersRef.current.forEach((handler) => handler(settings));
+  }, []);
+
+  const restore = useCallback(() => {
+    handlersRef.current.forEach((handler) => handler(originalSettingsRef.current));
+  }, []);
```
flopper123: Instead of having the `originalSettingsRef` at all, you could just here do this 😄 
```suggestion
  const restore = useCallback(() => {
    handlersRef.current.forEach((handler) => handler(originalSettings));
  }, [originalSettings]);
```

flopper123 (changes requested):
Overall very nice! 🔥 

Had a few code comments.

I found a bug during testing:
1) Open the settings dialog
2) Change the theme
3) Click on the `Usage` tab
4) Click `Cancel`

This closes the dialog, however, the theme change remains.

If you then reopen the dialog, click on the `Preferences` tab, then click `Cancel` (wthout changing anything), then the theme reverts.

dashboard/src/contexts/UserSettingsPreviewContext.tsx
Woldbye: Good point.. I changed it!

dashboard/src/contexts/UserSettingsPreviewContext.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+type UserSettingsPreviewProviderProps = {
+  children: React.ReactNode;
+  originalSettings: UserSettingsUpdate;
+};
+
+export function UserSettingsPreviewProvider({ children, originalSettings }: UserSettingsPreviewProviderProps) {
+  const handlersRef = useRef<Map<string, Dispatch<UserSettingsUpdate>>>(new Map());
+  const originalSettingsRef = useRef(originalSettings);
+  originalSettingsRef.current = originalSettings;
```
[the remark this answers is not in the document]
Woldbye: No the pattern you mention below in the other comment is a much better solution

PR #583, opened 2025-12-02
582 prevent race condition when creating subscriptions for the same user

### LINKED ISSUE #582: Prevent race condition when creating subscriptions for the same user
We occasionally encounter this exception in production:

Image

This happens due to a race condition where multiple requests attempt to create a subscription for the same user simultaneously. We should resolve this by either using an atomic upsert operation or ensuring that a subscription is created deterministically during user creation.

Lindharden:
This PR prevents a race condition that could occur when creating the default subscription and initial user settings for new users. Previously, concurrent page fetches could trigger multiple subscription creation attempts.

Default subscriptions and user settings are now created atomically when using the CredentialsProvider, as well as through the PrismaAdapter event handler. If the adapter event handler fails, the system remains safe because the getters for both subscription and user settings use upserts, ensuring they are created if missing.

Closes #582

### DIFF
--- a/dashboard/src/repositories/postgres/subscription.ts
+++ b/dashboard/src/repositories/postgres/subscription.ts
@@ -1,24 +1,21 @@
 import prisma from '@/lib/postgres';
 import {
-  STARTER_SUBSCRIPTION_STATIC,
   Subscription,
   SubscriptionSchema,
   UpsertSubscriptionData,
   UpsertSubscriptionSchema,
-  buildStarterSubscriptionWindow,
+  buildStarterSubscription,
 } from '@/entities/billing';
 
 export async function getUserSubscription(userId: string): Promise<Subscription | null> {
   try {
-    const { currentPeriodStart, currentPeriodEnd } = buildStarterSubscriptionWindow();
+    const starterSubscription = buildStarterSubscription();
 
     const subscription = await prisma.subscription.upsert({
       where: { userId },
       create: {
         user: { connect: { id: userId } },
-        ...STARTER_SUBSCRIPTION_STATIC,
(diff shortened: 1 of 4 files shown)


dashboard/src/repositories/postgres/subscription.ts
```diff
@@ -1,45 +1,35 @@
...
+    const subscription = await prisma.subscription.upsert({
       where: { userId },
+      create: {
+        user: { connect: { id: userId } },
+        ...STARTER_SUBSCRIPTION_STATIC,
+        currentPeriodStart,
+        currentPeriodEnd,
+      },
+      update: {},
     });
```
flopper123: This seems incorrect. It now runs `upsert` on the users `get` subscription function.

flopper123 (changes requested):
Works nicely locally.
I have a general question/concern. It seems like it will now simply override any pre-existing subscription, instead of failing (which of course fixes the race-condition). For example:
Could a user, who has a (paid) subscription, accidentally hit one of these paths where it `upserts`? In that case it would overwrite their subscription to the "starter" subscription.

Also had some code comments.

dashboard/src/repositories/postgres/subscription.ts
Lindharden: Yes, that is intended. We want to ensure that if the user subscription is ever fetched, then it should exist. If not, it should be created. This is in case the chained event listening for user creation from the PrismaAdapter fails to create subscription, we don't want the user to exist without a subscription.

The update bracket is empty, so it doesn't update the entry. It only creates if it doesn't exist 👍

PR #589, opened 2025-12-03
Introduces request boundary cache for auth authentication

Lindharden:
This PR introduces request boundary cache for our server-action auth wrappers.

Switched all session/auth lookups to use these caches:
- `getAuthSession` now calls `getCachedSession()` instead of `getServerSession(authOptions)`.
- `tryGetAuthorizedContext` uses `getCachedAuthorizedContext(...)`.
- `resolveDemoDashboardContext` and `resolvePrivateDashboardContext` use `getCachedSession()` and `tryGetAuthorizedContext`.
- `requireDashboardAuth` now uses `getCachedAuthorizedContext(session.user.id, dashboardId)`.

This means, that all parallel calls to our wrapped actions share the same getServerSession and getAuthorizedDashboardContextOrNull results. This reduces the number of repeated auth hits per request with no real downside. For instance, the overview page does 10+ server actions in parallel during one RSC render - this currently results in 10+ distinct dashboard authentication queries. With these changes, a single query covers all actions in the RSC render.

This is a prerequisite for swapping our auth to a DB session strategy, to prevent additional repeated session queries when triggering actions in parallel.

Nothing is shared across requests. Ecah request still does its own fresh auth, so there's no staleness or cross-user leakage.

### DIFF
--- a/dashboard/src/auth/auth-actions.ts
+++ b/dashboard/src/auth/auth-actions.ts
@@ -13,6 +13,7 @@ import { unstable_cache } from 'next/cache';
 import { DashboardFindByUserSchema } from '@/entities/dashboard';
 import { stableStringify } from '@/utils/stableStringify';
 import { trace, SpanStatusCode } from '@opentelemetry/api';
+import { cache } from 'react';
 
 // Stable per-action signature to avoid cache key collisions (alternatively we provide each function an explicit name)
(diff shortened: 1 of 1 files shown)


flopper123 (approved):
Awesome! 🔥

PR #639, opened 2025-12-22
Added handling for OAuthAccountNotLinked

### LINKED ISSUE #442: Provide better feedback for OAuthAccountNotLinked response

Lindharden:
Closes #442


flopper123 (approved):
Very nice 😄

Woldbye:
Is the OAuthNotLinked error only send if the account already exists ?? By the name it seems like a general oAuth error to me

PR #640, opened 2025-12-22
Fixed label mismatch for campaign entries

### LINKED ISSUE #619: Sparkline and total sessions on Campaigns differs, but their label is the same
## Description:
Campaign page shows two different metrics both labeled "sessions". The sparkline tooltip displays total sessions, while the count under the campaign title displays unique visitors (sessions). This causes confusion as both use the same label.

## How to reproduce:
  1. Run simulate-events script to generate campaigns data
  2. Navigate to Campaigns page
  3. Hover over sparkline to see tooltip value
  4. Compare with the sessions count shown under campaign title
  5. Notice the values differ (e.g., 389 vs 313) despite same label

## Screenshot:
Image

Lindharden:
Closes #619

### DIFF
--- a/dashboard/messages/da.json
+++ b/dashboard/messages/da.json
@@ -1378,7 +1378,7 @@
       },
       "loading": "Indlæser kampagner...",
       "campaignRow": {
-        "sessions": "{count, plural, one {# session} other {# sessioner}}",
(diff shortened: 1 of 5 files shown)


flopper123 (approved):
Very nice 😄

PR #643, opened 2025-12-23
generalize colors

### LINKED ISSUE #470: Reuse color calculations
We have color utilities in multiple places (Pie charts, graphs, etc.) that calculates a color based on a value.

We could/should probably abstract/reuse these around the app.

evanyang1:
closes #470


flopper123 (approved):
Awesome cleanup! 👏😄

PR #646, opened 2025-12-31
Dedicated pricing page with plan comparison

Lindharden:


flopper123 (changes requested):
Overall looks super nice! I have two small comments:

1. The pills on light mode are a little invisible, at least on my screen. Maybe they should have a border of some kind
image

2. The "Pricing" link in the public top bar should probably link to the new page:
image

3. I get an error saying there's are duplicate keys:
image

Lindharden:
> Overall looks super nice! I have two small comments:
> 
> 1. The pills on light mode are a little invisible, at least on my screen. Maybe they should have a border of some kind

That is delibarate. Border does not look nice. They intentionally appear if almost just placed directly on the background :)

> 2. The "Pricing" link in the public top bar should probably link to the new page:

Nice catch, fixed 👍 

> 3. I get an error saying there's are duplicate keys:

Fixed 👍

flopper123 (approved):
Awesome! 🤩

PR #647, opened 2025-12-31
Updated versions

flopper123:
What I did:
- Updated the dependencies that could be updated manually
- Ran `pnpm audit -P --fix` to fix all sub-dependencies used in prod
- Tested locally that things could still build, and everything seems to work as expected


Woldbye (approved):
Still compiles and runs for me 👍

Lindharden (approved):

PR #648, opened 2026-01-01
Frontend component cleanup

flopper123:


dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/WeeklyHeatmapSection.tsx
```diff
@@ -191,15 +191,13 @@ function HeatmapGrid({ data, maxValue, metricLabel, metric }: HeatmapGridProps)
...
                 >
-                  <div>
-                    <div className='text-popover-foreground font-medium'>
-                      {`${dayLabels[dayIndex]} ${String(hourIndex).padStart(2, '0')}:00 - ${String((hourIndex + 1) % 24).padStart(2, '0')}:00`}
-                    </div>
-                    <div className='text-popover-foreground/90'>
-                      {`${formatHeatmapMetricValue(metric, value)} ${metricLabel.toLowerCase()}`}
-                    </div>
+                  <div className='text-popover-foreground font-medium'>
+                    {`${dayLabels[dayIndex]} ${String(hourIndex).padStart(2, '0')}:00 - ${String((hourIndex + 1) % 24).padStart(2, '0')}:00`}
```
Comment (Lindharden) [inline dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/WeeklyHeatmapSection.tsx]: Should this also be wrapped in the new Text component?


```suggestion
                    <Text
                      variant='body'
                      tone='muted'
                    >{`${dayLabels[dayIndex]} ${String(hourIndex).padStart(2, '0')}:00 - ${String((hourIndex + 1) % 24).padStart(2, '0')}:00`}</Text>
```

dashboard/src/components/dashboard/SummaryCardsSection.tsx
```diff
@@ -21,7 +21,7 @@ type SummaryCardsSectionProps = {
 
 export default function SummaryCardsSection({ cards, className }: SummaryCardsSectionProps) {
   return (
-    <div className={`grid grid-cols-2 gap-2 sm:gap-4 md:grid-cols-2 xl:grid-cols-4 ${className}`}>
+    <div className={`sm:gap-layout-xl gap-layout-md grid grid-cols-2 md:grid-cols-2 xl:grid-cols-4 ${className}`}>
```
Lindharden: This shouldnt be utilizing the new Grid component?

Lindharden (changes requested):
Pretty neat. I left a few comments.

I'm also wondering, since there's more to cleanup, such as creating custom button variants and reusing this throughout the dashboard, removing redundant cursor-pointer stylings, and more: do you think it would be beneficial to create subtasks for this or do you already have a plan for a complete set of custom component wrappers?

Woldbye (commented):
Nice work! Good clean up to standardize layouts. 

I agree with Lind that it would be nice to update our issues with the full clean up plan, but I don't think the full clean up should be added to this PR.

One question: Did you consider using 'cva' for variant definitions across all files? It would make the API a lot more flexible, and match the already established patterns by shadcn.
I especially think it would be nice for the `<Text>`-component, as it would make it easier to support working with consistently styled text on a character level.

PR #665, opened 2026-01-05
Improved design of the about page

Lindharden:


flopper123 (changes requested):
Awesome!
I have one small thing. On white mode where the there's no card around the `PrimaryContentCards`, the layout looks a bit "off", since they don't align with the `SecondaryContactItems`. Should they maybe align on white mode, or maybe remove the card from dark mode and also align them?

image


image

PR #669, opened 2026-01-06
Removed plan-details from pricing documentation

Lindharden:
With the introduction of the dedicated pricing page with a plan-feature comparison table, we should no longer have plan details in the pricing docs. 

This PR simplifies the plan overview documentation to a more high-level perspective and corrects links.


Lindharden:
> Looks good to me.
> 
> What do you think of adding a mention of the "free" (no credit card req) plan somewhere, as this is what most indie developers would be searching for.

We have this in the pricing page and landing page 😄

PR #670, opened 2026-01-06
Invalidates sessions on password changes

Lindharden:
- Invalidates all sessions, except current, when changing password.
- Invalidates all sessions when changing password using the forgot-password form.


Woldbye (dismissed):
Nice to see a small surgical change. Code looks good!! 

A lot of websites offer users an option to revoke all sessions or specific sessions through the user panel. What do you think of using what you have created here to add something like that?? (We can open a seperate issue, it doesn't have to be in this pr)

flopper123 (dismissed):
Awesome! Works wonders 🤩

PR #728, opened 2026-01-27
Added applied filter count to filter selector trigger

Lindharden:
An user did not initially notice that he had accidentally applied filters (by clicking on table entries) and he couldn't figure out why data was not showing.

This PR adds an "applied filter count" to the filter selector trigger in an attempt to make it slightly more visible that filters are selected.

### DIFF
--- a/dashboard/src/components/filters/QueryFiltersSelector.tsx
+++ b/dashboard/src/components/filters/QueryFiltersSelector.tsx
@@ -79,6 +79,8 @@ export default function QueryFiltersSelector() {
 
   const { data: isSavedFiltersLimitReached } = useSavedFiltersLimitReached();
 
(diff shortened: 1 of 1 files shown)


Woldbye:
Good idea!!

I have two suggestion for improvements:

1. Add some glow styling to make it pop up a bit more. 

2. Use the number-flow component i introduce on PR#730 to animate the counter.

Wdyt ?

Lindharden:
> I have two suggestion for improvements:
> 
> 1. Add some glow styling to make it pop up a bit more.
> 2. Use the number-flow component i introduce on PR#730 to animate the counter.

I tried adding a few different variant of glows and I couldn't make it flow nicely with the design :( However, the filters should be further accentuated when they're properly inlined.

For the number-flow, I don't think that belongs for this one, because it's not very interactive and it doesn't change while the user sees it 😄

PR #730, opened 2026-01-29
Use @number-flow/react to animate numbers

Woldbye:
- Replaced custom animation implementation with number-flow library
  - Added willChange prop for GPU acceleration on rapid updates
  - Used ICU plural format for localization (unit keys)
  - Small layout changes to accomodate the new animation 

## Demo
msedge_uBnVOfjMeE

### DIFF
--- a/dashboard/src/components/pricing/PricingSlider.tsx
+++ b/dashboard/src/components/pricing/PricingSlider.tsx
@@ -2,7 +2,7 @@
 
 import { EVENT_RANGES, EventRange } from '@/lib/billing/plans';
 import { cn } from '@/lib/utils';
-import { useLocale, useTranslations } from 'next-intl';
+import { useTranslations } from 'next-intl';
 import NumberFlow from '@number-flow/react';
 
 interface PricingSliderProps {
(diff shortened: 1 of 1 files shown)


dashboard/src/components/inputs/LabeledSlider.tsx
```diff
@@ -74,9 +78,24 @@ export function LabeledSlider({
...
+          style={{ gridTemplateColumns: isRecommended ? 'auto 1fr' : 'auto 0fr' }}
+        >
+          {valueParts ? (
+            <NumberFlow
+              className='tabular-nums'
+              value={valueParts.value}
+              locales={locale}
+              format={valueParts.intlUnit ? { style: 'unit', unit: valueParts.intlUnit, unitDisplay: 'narrow' } : undefined}
+              suffix={valueParts.suffix ? ` ${valueParts.suffix}` : undefined}
+              willChange
```
flopper123: I'm unsure if we want the `willChange` on every `NumberFlow`. Based on the documentation it seems like an escape hatch:
Image

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/will-change

flopper123 (changes requested):
Awesome! 🤩 
Only a few comments:

1)
The height if the pricing cards changes quite a bit between the "Free" --> "NumberRoll" and the "NumberRoll" --> "Contact Us".

These cards should optimally not change height 😄

Image

PR #731, opened 2026-01-30
Add Gauge component and integrate with Core Web Vitals

Woldbye:
Depends on #730 , as it uses it to animate the values on the public page.

## Changes
- Add reusable Gauge component with SVG arc segments and needle
- Add math-utils.ts for polar/cartesian coordinate helpers
- Replace web-vitals bar visualization with gauge visualization
- Each CWV metric now displays with an interactive gauge showing:
  - Good/Fair/Poor threshold segments
  - Animated progress arc
  - Needle pointing to current value
- Needle is optional - What do you guys think, does it work or better without?
- Replaced the old gauges on the landing place with the new animated gauges.
- Update amount of red zone slightly from the previous 8%, as I think it looks visually more pleasing.

## Demo
chrome_LvGTosbxhC

chrome_bf1ecZomfz

### DIFF
--- a/dashboard/src/components/pricing/PricingSlider.tsx
+++ b/dashboard/src/components/pricing/PricingSlider.tsx
@@ -20,6 +20,8 @@ export function PricingSlider({
 }: PricingSliderProps) {
   const t = useTranslations('pricingSlider');
 
+  const isUnlimited = currentRange.value > 10_000_000;
+
   return (
     <div className={className}>
       <div className='mb-4 flex h-16 flex-col items-center justify-end'>
@@ -29,7 +31,7 @@ export function PricingSlider({
           value={currentRange.value}
           locales={'en-US'}
           format={{ notation: 'compact' }}
-          suffix={currentRange.value > 10_000_000 ? '+' : undefined}
+          suffix={isUnlimited ? '+' : undefined}
           willChange
         />
       </div>


flopper123 (changes requested):
Very nice! 🤩 

----

A few UI/UX feedback:

1) The new gauges on landing page causes the heights to be extended. Optimally this should remain the same as before:

Image

2) There seems to be a lot of "red" / "bad numbers", I would dial it back so that most remain in green, some in yellow, and then maybe one in red at a time:

Image

3) The text / description is a little small, and somewhat hard to read. I'd suggest making the gauge-widths smaller, more like currently on prod, such that the text can be larger, without taking up more space 😄 

Image

4) The colors should probably also align with our other colors. Right now on light mode, the colors are very prominent, instead of being a little more laid back like the rest of our colors:

image

----

5) The web vitals chart is taking up more height now, so I'd suggest removing a little space between the cards and the explanation text:

Image

6) I'd suggest removing the needles, as it makes it look like a speedometer, and can overlap with the text:

Image

7) There seems to be a few flexing/layout issues:

Image

8) The formatting of the values have changes. On prod, there's a space between the value and the units. Now there's no space:

image

Lindharden (approved):
This is very much approved from my side.

There's a few minor tweaks that I would make if time allows, but it's approved nonetheless:

1) 

I would increase spacing between the title and number inside the gauge by a tiny bit. They are very cramped together, almost overlapping:

Image

Image

2)

There's a bit of empty space above and below the gauges in the landing page feature card. I would move the top row a tiny bit upwards, and the bottom row a tiny bit upwards. So that there's slightly more spacing between the two gauge rows, and less empty space in the card itself. (only by a tiny bit so they don't look disconnected)

Image

Woldbye:
> There's a bit of empty space above and below the gauges in the landing page feature card. I would move the top row a tiny bit upwards, and the bottom row a tiny bit upwards. So that there's slightly more spacing between the two gauge rows, and less empty space in the card itself. (only by a tiny bit so they don't look disconnected)


Good suggestions, it looks a bit better now. I have resolved both issues with latest commit [6cdd9ac](https://github.com/betterlytics/betterlytics/pull/731/commits/6cdd9acde253f84f533ce6ba2b9d05e05228511d)

Lindharden (approved):
Looks amazing!

PR #733, opened 2026-01-31
Email reports

Lindharden:
User-requested feature.

This PR introduces weekly and monthly email reports showcasing top-performing pages, top referrers, and aggregated statistics for the selected period.

It also fixes an issue where the email preview page was inaccessible because it had not been moved into the /[locale] folder following the introduction of localization.

Due to email expenses and limitations, the feature is gated behind PRO-plan.

### DIFF
--- a/dashboard/src/services/email/template/weekly-report-mail.ts
+++ b/dashboard/src/services/email/template/weekly-report-mail.ts
@@ -10,19 +10,22 @@ export interface EmailReportData extends EmailData {
   dashboardUrl: string;
 }
 
-function formatChange(change: number): string {
+function formatChange(change: number | null): string {
+  if (change === null) return '—';
   if (change === 0) return 'No change';
   const sign = change > 0 ? '+' : '';
   return `${sign}${change}%`;
 }
 
-function getTrendIcon(change: number): string {
+function getTrendIcon(change: number | null): string {
+  if (change === null) return '';
   if (change > 0) return '↑';
   if (change < 0) return '↓';
   return '→';
 }
 
-function getTrendColor(change: number): string {
+function getTrendColor(change: number | null): string {
+  if (change === null) return emailColors.neutral;
   if (change > 0) return emailColors.positive;
   if (change < 0) return emailColors.negative;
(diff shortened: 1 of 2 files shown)


dashboard/src/instrumentation.ts
```diff
@@ -1,5 +1,18 @@
 export async function register() {
   registerOpenTelemetry();
+  await registerBackgroundJobs();
```
flopper123: `registerOpenTelemetry` is not awaited, though it's also async. Should we maybe also remove the await for `registerBackgroundJobs`? 😄
It probably won't change much though. So just leave it as is if you feel like it

dashboard/src/services/email/template/weekly-report-mail.ts
```diff
@@ -0,0 +1,268 @@
...
+  if (change === 0) return 'No change';
+  const sign = change > 0 ? '+' : '';
+  return `${sign}${change}%`;
+}
+
+function getTrendIcon(change: number): string {
+  if (change > 0) return '↑';
+  if (change < 0) return '↓';
+  return '→';
+}
```
flopper123: Having both the trend and the change seems to do the same thing:

Image

Should we maybe only show one of them for these? 😄

Lindharden: Good catch. I've updated it to only display the arrows 😄

dashboard/src/services/reports/report-data.service.ts
```diff
@@ -0,0 +1,164 @@
...
+    end: Date;
+  };
+  metrics: ReportMetrics;
+  topPages: TopPage[];
+  topSources: TopSource[];
+}
+
+function calculatePercentChange(current: number, previous: number): number {
+  if (previous === 0) {
+    return current > 0 ? 100 : 0;
```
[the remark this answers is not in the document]
Lindharden: I've adjusted it to display a "—" when there is no comparison data 👍

flopper123 (approved):

PR #735, opened 2026-01-31
Unknown locale safeguards

Lindharden:
This PR addresses the “unknown ogLocale” issue observed in the logs. It leverages the native hasLocale utility from next-intl to validate locales and adds a locale check in the root locale layout, in line with the approach recommended in the next-intl documentation.


flopper123 (approved):
Very nice!

PR #762, opened 2026-02-14
Keyword monitoring

### LINKED ISSUE #756: Support keyword/content checks in uptime monitor
Add the ability to validate a specific keyword or string in the HTTP response body. This ensures alerts trigger even if a page returns 200 but contains broken or missing content.

Lindharden:
Adds support for keyword checks for uptime monitors

Closes #756

### DIFF
--- a/backend/src/monitor/probe.rs
+++ b/backend/src/monitor/probe.rs
@@ -466,7 +466,7 @@ fn check_keyword_match(expected: &Option<String>, body: &Option<Vec<u8>>) -> Opt
     };
 
     let haystack = String::from_utf8_lossy(body);
-    if haystack.to_lowercase().contains(&keyword.to_lowercase()) {
+    if haystack.contains(keyword) {
         None
     } else {
         Some(ReasonCode::KeywordNotFound)
--- a/docs/src/content/dashboard/monitoring.mdx
+++ b/docs/src/content/dashboard/monitoring.mdx
@@ -152,7 +152,7 @@ This catches scenarios where a page returns `200 OK` but serves broken content, 
 **How it works:**
 
 - The keyword is matched against the **raw HTML source code** (first 32 KB of the response)
-- Matching is **case-insensitive**
+- Matching is **case-sensitive**
 - The response is assumed to be **UTF-8 encoded**. Non-UTF-8 content may not match as expected
 - Maximum keyword length: 256 characters
 


dashboard/src/lib/billing/validators/monitoring.validators.ts
```diff
@@ -51,6 +51,13 @@ class MonitoringCapabilityValidator extends CapabilityValidator {
     }
     return this;
   }
+
+  expectedKeyword(keyword: string | null | undefined, httpMethod: string | undefined): this {
+    if (keyword != null && keyword.trim() !== '' && httpMethod !== 'HEAD') {
```
flopper123: Would probably make more sense to say that this is only valid on GET requests, maybe? 🤔 
```suggestion
    if (keyword != null && keyword.trim() !== '' && httpMethod === 'GET') {
```

docs/src/content/dashboard/monitoring.mdx
Woldbye: Add a note that `UTF-8` encoding is assumed, due to the `utf8_lossy` conversion the backend uses.

dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/monitoring/shared/components/AdvancedSettingsSection.tsx
```diff
@@ -104,6 +104,28 @@ export function AdvancedSettingsSection({
                 </Tabs>
               )}
             </CapabilityGate>
+
+            <div className='space-y-2 pt-1'>
```
Comment (Woldbye) [inline dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/monitoring/shared/components/AdvancedSettingsSection.tsx]: Should this be wrapped in a `<CapabilityGate>` like the other settings?

backend/src/monitor/probe.rs
```diff
@@ -427,6 +457,22 @@ fn ensure_crypto_provider() {
...
 
+/// Returns `Some(ReasonCode::KeywordNotFound)` if a keyword was expected but not found in the body.
+fn check_keyword_match(expected: &Option<String>, body: &Option<Vec<u8>>) -> Option<ReasonCode> {
+    let keyword = expected.as_deref().filter(|k| !k.is_empty())?;
+
+    let Some(body) = body.as_deref() else {
+        return Some(ReasonCode::KeywordNotFound);
+    };
+
+    let haystack = String::from_utf8_lossy(body);
```
Woldbye: `keyword.to_lowercase()` is recomputed on every probe execution. The keyword is static between config changes. Shoud we not precomputed this once when the `MonitorCheck` is loaded from the DB for performance?

Woldbye (commented):
Nice feature! Looks good overall. Had a few minor suggestions and questions

backend/src/monitor/probe.rs
Lindharden: the to_lowercase of the keyword should be neglible because it's only at most 256 chars. If we precompute this in the repository/cache then it wont be possible to reconstruct/reuse it elsewhere in the future, so I think we should just leave it as is. Lmk if you disagree 😄

PR #763, opened 2026-02-14
Added selfhost docs

flopper123:


Lindharden (changes requested):
Looks good 👍

Lindharden:
On another note, we should probably add a link to our github self-host repository.

PR #764, opened 2026-02-14
Added dev widget for DX

Lindharden:
Adds a floating developer widget in the bottom right corner to allow changing subscription, color theme & language more easily.

it is gated behind env.IS_DEVELOPMENT

### DIFF
--- a/dashboard/src/components/dev/DevWidget.tsx
+++ b/dashboard/src/components/dev/DevWidget.tsx
@@ -2,7 +2,6 @@
 
 import { useState, useTransition } from 'react';
 import { useTheme } from 'next-themes';
-import { useRouter, usePathname } from '@/i18n/navigation';
 import { useLocale } from 'next-intl';
 import { Wrench } from 'lucide-react';
 import { toast } from 'sonner';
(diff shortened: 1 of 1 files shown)


dashboard/src/components/dev/DevWidget.tsx
```diff
@@ -0,0 +1,125 @@
...
+      } catch {
+        toast.error('Failed to change subscription');
+      }
+    });
+  }
+
+  function handleLanguageChange(lang: string) {
+    router.push(pathname, { locale: lang as SupportedLanguages });
+    router.refresh();
+  }
```
flopper123: This only works on public pages. Doing this on a dashboard page results in `/da/dashboard/{id}/` for instance.

dashboard/src/components/dev/DevWidget.tsx
```diff
@@ -0,0 +1,125 @@
...
+}));
+
+const THEMES = [
+  { value: 'light', label: 'Light' },
+  { value: 'dark', label: 'Dark' },
+  { value: 'system', label: 'System' },
+];
+
+export function DevWidget({ initialTier }: DevWidgetProps) {
+  const { theme, setTheme } = useTheme();
```
flopper123: This only updates the theme until refresh, should this change be persisted? 🤔

flopper123 (changes requested):
Overall very nice 🔥 

Should we maybe also add a "simulate events" button? 🤔 I don't know if it can call the script directly, but would it make sense to add?

Lindharden:
> Should we maybe also add a "simulate events" button? 🤔 I don't know if it can call the script directly, but would it make sense to add?

I think this is better suited for CLI instead of a dev tool, considering it's typically a one-time seed. But we can always look into adding this if the need arises 😄

PR #776, opened 2026-02-19
Remove dead code across the dashboard

Woldbye:
Removes unused code, dead server action chains, and unused locale keys from the dashboard.

  - Deleted unused hooks, components, and utilities that have been replaced by newer implementations
  - Removed server actions (and their service/repo chains) that are no longer called by any component
  - Cleaned up 45 unused translation keys from all 3 locale files (en, da, it)
  - Removed commented-out Stripe `automatic_tax` block


flopper123 (approved):
Nice, looks correct to me 👍

PR #779, opened 2026-02-21
Pass single query object

flopper123:
This is the ground work for the MCP server

### DIFF
--- a/dashboard/src/middleware.ts
+++ b/dashboard/src/middleware.ts
@@ -1,7 +1,17 @@
 import createMiddleware from 'next-intl/middleware';
 import { routing } from '@/i18n/routing';
+import type { NextRequest } from 'next/server';
 
-export default createMiddleware(routing);
+const intlMiddleware = createMiddleware(routing);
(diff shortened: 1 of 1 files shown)


dashboard/src/entities/analytics/analyticsQuery.entities.ts
```diff
@@ -0,0 +1,31 @@
+import type { GranularityRangeValues } from '@/utils/granularityRanges';
+import type { QueryFilter } from '@/entities/analytics/filter.entities';
+import type { TimeRangeValue } from '@/utils/timeRanges';
+import type { CompareMode } from '@/utils/compareRanges';
+
+export type BAAnalyticsQuery = {
```
Woldbye: Did you consider adding a Zod schema `BAAnalyticsQuerySchema` for runtime-validation at the action-boundary??

flopper123: Good idea, will add 🫡

PR #780, opened 2026-02-21
Push notifications integrations 

### LINKED ISSUE #755: Pushover push notifications for uptime monitoring
Investigate the use of Pushover API for uptime alerts. If appropriate, allow uptime alerts to be sent to Pushover via their API, enabling users to receive push notifications alongside existing alert channels.

We should consider creating an abstract "alert channel" approach to better support additional alert channels in the future.

### LINKED ISSUE #786: Integration page for dashboard settings
In order to support integrations (push notificiations), a integration-configuration page needs to be added to the dashboard settings page.

Lindharden:
This PR adds prerequisites for integration push notification support:
- Integration settings page
- Integration schemas
- Pushover & Discord configurations
- Validation of configurations
- Encryption of config at rest

I have added a temporary env to enable/disable this page. This will be removed once the backend notification engine is implemented.

Closes #786 
Closes #755


dashboard/src/app/(protected)/dashboard/[dashboardId]/(settings)/settings/integrations/PushoverConfigDialog.tsx
```diff
@@ -0,0 +1,163 @@
...
+                if (error) setError(undefined);
+              }}
+              className={error ? 'border-destructive' : ''}
+              disabled={isPending}
+            />
+            {error && <p className='text-destructive text-sm'>{error}</p>}
+            <p className='text-muted-foreground text-xs'>
+              {t.rich('credentialsHint', {
+                link: (chunks) => (
+                  <a
```
Comment (Woldbye) [inline dashboard/src/app/(protected)/dashboard/[dashboardId]/(settings)/settings/integrations/PushoverConfigDialog.tsx]: Consider using the `ExternalLink` component here to stay consistent with the rest of the codebase.

.env.example
```diff
@@ -111,4 +111,8 @@ GITHUB_SECRET=
 GOOGLE_CLIENT_ID=
 GOOGLE_CLIENT_SECRET=
 
-BACKGROUND_JOBS_ENABLED=false
\ No newline at end of file
+BACKGROUND_JOBS_ENABLED=false
+
+# Integrations - Register an app at https://pushover.net/apps to get an API token
+PUSHOVER_APP_TOKEN=
```
Woldbye: Should probably be added to `.env.production` also?

.env.example
```diff
@@ -111,4 +111,8 @@ GITHUB_SECRET=
...
+PUBLIC_ENABLE_INTEGRATIONS=false
```
Woldbye: The `PUBLIC` prefix feels a bit misleading - Its not client-exposed.

dashboard/src/services/dashboard/integration.service.ts
```diff
@@ -0,0 +1,184 @@
...
+import {
+  Integration,
+  IntegrationType,
+  IntegrationConfigSchemas,
+} from '@/entities/dashboard/integration.entities';
+import * as IntegrationRepository from '@/repositories/postgres/integration.repository';
+import { symmetricEncrypt, symmetricDecrypt } from '@/lib/crypto';
+import { env } from '@/lib/env';
+
+const ENCRYPTION_KEY = env.NEXTAUTH_SECRET;
```
Woldbye: The existing TOTP service encrypts/decrypts user 2FA secrets using a dedicated key. It was my understand that this were done to address the specific security concern - that `env.NEXTAUTH_SECRET` would mask shorter than 32 character keys by padding them with null-bytes.

To resolve this, I think you need a dedicated `env.INTEGRATION_SECRET_ENCRYPTION_KEY` here, that you zod validate to exactly 32 chars ( `z.string().length(32)` )

dashboard/src/entities/dashboard/integration.entities.ts
```diff
@@ -0,0 +1,58 @@
+import { z } from 'zod';
+
+export const INTEGRATION_TYPES = ['pushover', 'discord'] as const;
+export type IntegrationType = (typeof INTEGRATION_TYPES)[number];
```
Woldbye:
```suggestion
const integration_types = ['pushover', 'discord'] as const;
export type IntegrationType = (typeof INTEGRATION_TYPES)[number];
export const INTEGRATION_TYPES = integration_types as [IntegrationType, ...IntegrationType[]];
```

Exposing it like this, gives clean typing on the `INTEGRATION_TYPES` for type interference, which elminates any future casts in any data-driven usage. I'd heavily suggest adopting this pattern for PushoverPriorities aswell, and any future enums we add to the project.

dashboard/src/app/(protected)/dashboard/[dashboardId]/(settings)/settings/integrations/IntegrationsSettings.tsx
```diff
@@ -0,0 +1,185 @@
...
+  useEffect(() => {
+    startTransition(async () => {
+      try {
+        const [types, existing] = await Promise.all([
+          getAvailableIntegrationTypesAction(dashboardId),
+          getIntegrationsAction(dashboardId),
+        ]);
+        setAvailableTypes(types);
+        setIntegrations(existing);
+      } catch {}
```
Comment (Woldbye) [inline dashboard/src/app/(protected)/dashboard/[dashboardId]/(settings)/settings/integrations/IntegrationsSettings.tsx]: I think we should send the toast.error here?
```suggestion
      } catch {
        toast.error(t('toast.error'));
      }
```

dashboard/src/app/(protected)/dashboard/[dashboardId]/(settings)/settings/integrations/IntegrationsSettings.tsx
```diff
@@ -0,0 +1,185 @@
...
+    });
+  }, [dashboardId]);
+
+  const getIntegrationForType = (type: IntegrationType): Integration | undefined => {
+    return integrations.find((i) => i.type === type);
+  };
```
Comment (Woldbye) [inline dashboard/src/app/(protected)/dashboard/[dashboardId]/(settings)/settings/integrations/IntegrationsSettings.tsx]: This seems like a symptom of modeling this as an array. The data is keyed by type everywhere, and the code keeps fighting by searching arrays. If you change the type of integrations to `Record<IntegrationType, Integration>` it would simplify a lot of the code here.

dashboard/src/app/(protected)/dashboard/[dashboardId]/(settings)/settings/integrations/IntegrationsSettings.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  const handleSave = (type: IntegrationType, config: Record<string, unknown>) => {
```
Comment (Woldbye) [inline dashboard/src/app/(protected)/dashboard/[dashboardId]/(settings)/settings/integrations/IntegrationsSettings.tsx]: You could type this as `const handleSave = (type: IntegrationType, config: PushoverConfig | DiscordConfig)` and leverage the Zod scheme

dashboard/src/services/dashboard/integration.service.ts
```diff
@@ -0,0 +1,184 @@
...
+  } catch (error) {
+    console.error('Error getting integration:', error);
+    throw new Error('Failed to get integration');
+  }
+}
+
+export async function saveIntegration(
+  dashboardId: string,
+  type: IntegrationType,
+  config: Record<string, unknown>,
```
Woldbye:
```suggestion
  config: IntegrationConfig,
```

dashboard/src/services/dashboard/integration.service.ts
```diff
@@ -0,0 +1,184 @@
...
+  discord: async (config) => {
+    if (typeof config.webhookUrl !== 'string') return 'invalid_discord_webhook';
+    const isValid = await validateDiscordWebhookUrl(config.webhookUrl);
+    return isValid ? null : 'invalid_discord_webhook';
+  },
+};
+
+export async function validateIntegrationConfig(
+  type: IntegrationType,
+  config: Record<string, unknown>,
```
Woldbye:
```suggestion
  config: IntegrationConfig,
```

dashboard/src/services/dashboard/integration.service.ts
```diff
@@ -0,0 +1,184 @@
...
+  IntegrationType,
+  IntegrationConfigSchemas,
+} from '@/entities/dashboard/integration.entities';
+import * as IntegrationRepository from '@/repositories/postgres/integration.repository';
+import { symmetricEncrypt, symmetricDecrypt } from '@/lib/crypto';
+import { env } from '@/lib/env';
+
+const ENCRYPTION_KEY = env.NEXTAUTH_SECRET;
+
+function encryptConfig(config: Record<string, unknown>): Record<string, unknown> {
```
Woldbye:
```suggestion
function encryptConfig(config: IntegrationConfig): { encrypted: string } {
```

dashboard/src/services/dashboard/integration.service.ts
```diff
@@ -0,0 +1,184 @@
...
+  const encrypted = symmetricEncrypt(JSON.stringify(config), ENCRYPTION_KEY);
+  return { encrypted };
+}
+
+function decryptConfig(config: Record<string, unknown>): Record<string, unknown> {
```
Woldbye:
```suggestion
function decryptConfig(config: { encrypted: string }):  IntegrationConfig {
```

Woldbye (commented):
Nice work 👍 

Overall really nice! :D I think its the right move to create a dedicated integration settings dashboard. This allows us to give users concrete control over each integration. For example, we could in the future let users decide if Pushover downtime warnings as critical push notifications, which are emitted repeatedly on the phone until acknowledged.

Visually it also looks pleasing, though I had a few concrete security questions I flagged in the PR, and some suggestions for more tight typing. 

1. It seems like the code type-safety and readability would benefit greatly from declaring `type IntegrationConfig = DiscordConfig |  PushoverConfig` once and replacing many of the `Record<string, unknown>` with this.  Many of my comments assume you have already done this (when I suggest IntegrationConfig in the comments).

2. The whole point of encrypting configs  (which I agree is absoultely correct, as you do it here) is so that a database breach doesn't expose the secrets. But then `getIntegrationsAction` decrypts them and sends them to the browser on every page load - not just when editing. If there's an XSS vulnerability anywhere in the dashboard, an attacker can call `getIntegrationsAction(dashboardId)` and get every integration secret in plaintext, thereby defeating the purpose of the encryption.
As far as I understand, the reason you are doing it, is to prepopulate the settings for the UX. But is this really worth it for the security concern?? We could maybe populate it with 'xxxx' or something to show its set, but not run into the security concern.

3. Change the `<a>` links to the use the `ExternalLink`, to stay consistent with the rest of the codebase.

4. Are we legally allowed to modify the .svg's?? I had a few concrete code comments on the .svg's, but I'm not sure we are legally allowed to change them, so I didnt add them.

5. WDYT of opening a subissue connected to #780, and connect it to the PR for better logging of our progress on the task?

dashboard/messages/en.json
```diff
@@ -3505,5 +3506,75 @@
...
+    },
+    "actions": {
+      "configure": "Configure",
+      "add": "Add",
+      "edit": "Edit",
+      "disconnect": "Disconnect"
+    },
+    "disconnect": {
+      "title": "Disconnect {name}",
+      "description": "Are you sure you want to disconnect this integration? This will remove the configuration and stop all notifications.",
```
flopper123: Quibble: Maybe state that it'll only stop all notification from this source, and not globally. Ignore if it doesn't matter
```suggestion
      "description": "Are you sure you want to disconnect this integration? This will remove the configuration and stop all notifications from this source.",
```

dashboard/src/app/(protected)/dashboard/[dashboardId]/(settings)/settings/integrations/DiscordConfigDialog.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const [webhookUrl, setWebhookUrl] = useState(existingConfig?.webhookUrl ?? '');
+  const [error, setError] = useState<string>();
+
+  useEffect(() => {
+    if (open) {
+      const config = integration?.config as Partial<DiscordConfig> | undefined;
+      setWebhookUrl(config?.webhookUrl ?? '');
+      setError(undefined);
+    }
+  }, [open, integration]);
```
Comment (flopper123) [inline dashboard/src/app/(protected)/dashboard/[dashboardId]/(settings)/settings/integrations/DiscordConfigDialog.tsx]: This seems to be already handled by the `const [webhookUrl, setWebhookUrl] = useState(existingConfig?.webhookUrl ?? '');`, so is this `useEffect` necessary? 😄

flopper123 (commented):
Visually looks very nice 😄 

Some small UI/UX comments:

1. Maybe we should reword the "Contact us and we'll add it." as it feels like we'll definitely add it:
Image

2. This feature is currently activateable for free users as well. Should this be for Pro+?

3. For input, would it make sense to save with Enter?

Image

4. If you disable the integration, it still says "Connected". Should it maybe be grayed out, or something like that? 🤔 

Image

5. (NOT SURE) The disable should maybe have a small label, just such that the user can quickly see what it actually does?

Image

.env.example
Lindharden: Yeah this is to have it available in the client 😄 

But nonetheless, as I describe in the pull request description, this will be removed when integration is fully functional :)

dashboard/src/services/dashboard/integration.service.ts
```diff
@@ -0,0 +1,184 @@
...
+    throw new Error('Failed to toggle integration');
+  }
+}
+
+export async function validateDiscordWebhookUrl(webhookUrl: string): Promise<boolean> {
+  if (!/^https:\/\/discord\.com\/api\/webhooks\//.test(webhookUrl)) return false;
+
+  try {
+    const response = await fetch(webhookUrl, {
+      method: 'POST',
```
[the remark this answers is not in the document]
Lindharden: I've separated to GET for validation, and POST for initial creation 👍

Lindharden:
> 2. This feature is currently activateable for free users as well. Should this be for Pro+?

I think Pushover should be gated behind Pro in the future, but for now it seems a bit reserved to only allow discord as sole push notifications 🤔 

> 5. (NOT SURE) The disable should maybe have a small label, just such that the user can quickly see what it actually does?

I personally think it's already intuitive that the toggle is associated with toggling the integration, i think the label will add some clutter considering there's already a status badge. Wyt? 😄 

Nonetheless, I have addressed your comments 👍

dashboard/src/app/actions/dashboard/integrations.action.ts
```diff
@@ -0,0 +1,71 @@
...
+import { withDashboardAuthContext, withDashboardMutationAuthContext } from '@/auth/auth-actions';
+import { AuthContext } from '@/entities/auth/authContext.entities';
+import * as IntegrationService from '@/services/dashboard/integration.service';
+import { revalidatePath } from 'next/cache';
+import { env } from '@/lib/env';
+
+const integrationAvailability: Record<IntegrationType, () => boolean> = {
+  pushover: () => !!env.PUSHOVER_APP_TOKEN,
+  discord: () => true
+};
```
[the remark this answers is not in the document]
Woldbye: I agree with the record approach. These are keyed by integration type - an array would be misleading.

Woldbye (approved):
Approved!! Looks spot on to me now. I was especially happy to see you adopt the data driven approach.

flopper123 (approved):

PR #808, opened 2026-03-01
Notifications log

Lindharden:
Adds notification log to the integrations page:

image


dashboard/src/app/(protected)/dashboard/[dashboardId]/(settings)/settings/integrations/NotificationHistoryDialog.tsx
```diff
@@ -0,0 +1,102 @@
...
+                    <TableHead>{t('history.columns.error')}</TableHead>
+                  </TableRow>
+                </TableHeader>
+                <TableBody>
+                  {rows.map((row, i) => (
+                    <TableRow key={i}>
+                      <TableCell className='text-muted-foreground text-xs'>
+                        {formatRelativeTimeFromNow(row.ts)}
+                      </TableCell>
+                      <TableCell>{row.integrationType}</TableCell>
```
Comment (Woldbye) [inline dashboard/src/app/(protected)/dashboard/[dashboardId]/(settings)/settings/integrations/NotificationHistoryDialog.tsx]: Can we use the proper integration icon here? It looks a bit bleak that it appears like this:

Image

When the proper icon is right behind it

flopper123 (changes requested):
It doesn't seem to catch the "Error", or is this Error only used for when the actual webhook fails?

Image

Should the initial connection be counted as a notification? I got a Discord notification when connecting as usual, however, that one is not in the log.
Image

PR #811, opened 2026-03-06
Client error tracking

Lindharden:
- Client-side JavaScript error tracking with fingerprinting and grouping
- Errors are fingerprinted in the Rust backend using stack frames (V8 + Firefox format support) with SHA-256 hashing
- Error overview page with sortable/filterable table, sparkline volume charts, status management (resolve/ignore/unresolve), bulk actions, and search
- Shared PaginationControls component used across errors and campaigns for consistency
- Empty state with table preview illustration when no errors have been recorded

Note, it is intentional that translations are not included. This is the first part of client-side error tracking and will not be released until fully complete.

### DIFF
--- a/dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/errors/ErrorList.tsx
+++ b/dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/errors/ErrorList.tsx
@@ -304,15 +304,17 @@ export function ErrorTable({ errorGroups, initialVolumeMap, dashboardId }: Error
       },
       {
         id: 'last_seen',
-        accessorFn: (row) => row.last_seen.getTime(),
+        accessorFn: (row) => row.last_seen?.getTime() ?? 0,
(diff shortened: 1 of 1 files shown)


backend/src/processing/mod.rs
```diff
@@ -202,6 +212,18 @@ impl EventProcessor {
...
+            let list = processed.event.raw.exception_list.clone().unwrap_or_default();
+            if let Ok(arr) = serde_json::from_str::<serde_json::Value>(&list) {
+                processed.error_type = arr[0]["type"].as_str().unwrap_or("").to_string();
+                processed.error_message = arr[0]["value"].as_str().unwrap_or("").to_string();
+            }
+            processed.error_fingerprint = generate_error_fingerprint(
+                &processed.error_type,
+                &list,
+            );
+            processed.exception_list = list;
```
flopper123: Suggestion: Move some of this to a "process_js_error", as it does a little more than just mapping.

PR #813, opened 2026-03-08
Improved referrer parsing and grouping

Lindharden:
This PR removes the use of Snowplow referrer parsing for referrer names.

Referrer names are now always stored as the domain (e.g., reddit.com) instead of a normalized label like "Reddit". This ensures consistent grouping and avoids the ongoing maintenance required to classify new or changing sites.

Snowplow is now only used for referrer classification (e.g., social, direct, etc.). The long-term plan is to fully deprecate this dependency and move the classification logic client-side.

The referrer name column is currently unused but has been retained for safety 😄 

Additionally, this PR removes the "Sources" tab and now shows a combined "Channel" and "Sources" tab.

### DIFF
--- a/dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/(overview)/TrafficSourcesSection.tsx
+++ b/dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/(overview)/TrafficSourcesSection.tsx
@@ -22,12 +22,11 @@ export default function TrafficSourcesSection({ trafficSourcesCombinedPromise }:
       if (item.children) return;
       return makeFilterClick('referrer_url')(item.label);
     }
-    if (tabKey === 'sources') return makeFilterClick('referrer_source')(item.label);
     if (tabKey === 'channels') return makeFilterClick('referrer_source')(item.label);
   };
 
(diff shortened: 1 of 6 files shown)


dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/(overview)/TrafficSourcesSection.tsx
```diff
@@ -17,14 +17,16 @@ export default function TrafficSourcesSection({ trafficSourcesCombinedPromise }:
   const t = useTranslations('dashboard');
   const { makeFilterClick } = useFilterClick({ behavior: 'replace-same-column' });
 
-  const onItemClick = (tabKey: string, item: { label: string }) => {
-    if (tabKey === 'referrers') return makeFilterClick('referrer_url')(item.label);
-    if (tabKey === 'sources') return makeFilterClick('referrer_source')(item.label);
+  const onItemClick = (tabKey: string, item: { label: string; children?: unknown[] }) => {
+    if (tabKey === 'referrers') {
+      if (item.children) return;
```
Comment (flopper123) [inline dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/(overview)/TrafficSourcesSection.tsx]: This should probably still be clickable, like devices

PR #818, opened 2026-03-12
Performance update

flopper123:
- Updated table ordering
- Added sampling capability for large sites
- Added temporary concurrency limit for large sites to prevent potential crashes

### DIFF
--- a/dashboard/src/components/topbar/PublicTopBar.tsx
+++ b/dashboard/src/components/topbar/PublicTopBar.tsx
@@ -7,6 +7,7 @@ import Logo from '@/components/logo';
 import { useState } from 'react';
 import { Menu, X } from 'lucide-react';
 import ExternalLink from '@/components/ExternalLink';
+import { GitHubIcon } from '@/components/icons/SocialIcons';
 import { useTranslations } from 'next-intl';
 import NextLink from 'next/link';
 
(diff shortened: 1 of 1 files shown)


migrations/21_optimize_events_primary_key.sql
```diff
@@ -0,0 +1,110 @@
+-- Migration 21: Optimize events primary key
+-- Changes ORDER BY from (site_id, date, visitor_id, session_id, timestamp)
+-- to (site_id, event_type, toDate(timestamp), visitor_id, timestamp)
+
+CREATE TABLE IF NOT EXISTS analytics.events_new (
+    site_id String,
+    visitor_id String,
+    session_id String,
+    domain String DEFAULT '',
```
Lindharden: Should be LowCardinality

dashboard/src/observability/clickhouse-concurrency.ts
```diff
@@ -0,0 +1,69 @@
...
+async function withConcurrencyLimit<T>(siteId: string, fn: () => Promise<T>): Promise<T> {
+  const limiter = limiters.get(siteId);
+  if (!limiter) return fn();
+
+  if (limiter.running < limiter.limit) {
+    limiter.running++;
+  } else {
+    await new Promise<void>((resolve) => {
+      limiter.queue.push({ resolve });
+    });
```
Woldbye: What's the point of wrapping this in a promise?? Isnt it a sync push operation

dashboard/src/observability/clickhouse-concurrency.ts
```diff
@@ -0,0 +1,69 @@
...
+  try {
+    return await fn();
+  } finally {
+    limiter.running--;
+    const next = limiter.queue.shift();
+    if (next) {
+      limiter.running++;
+      next.resolve();
+    }
+  }
```
Woldbye: Should we be catching `fn` promise errors here?

dashboard/src/observability/clickhouse-instrumented.ts
```diff
@@ -9,6 +10,19 @@ function sanitizeStatement(statement: string, maxLength: number = 2000): string
   return condensed.length > maxLength ? condensed.slice(0, maxLength) : condensed;
 }
 
+function normalizeQuery(sql: string): { hash: string; normalized: string } {
+  const normalized = sql
+    .replace(/\s+/g, ' ')
+    .replace(/'(?:[^'\\]|\\.)*'/g, '?')
+    .replace(/\b\d+(\.\d+)?\b/g, '?')
+    .replace(/\(\s*\?(?:\s*,\s*\?)+\s*\)/g, '(?)')
+    .trim();
```
Woldbye: Quibble: To my understanding this gets called on large sql strings often.

Therefore it might make sense to squash replaces into a single replace to avoid copying the string each time.

PR #831, opened 2026-03-28
Add dot-grid animation to landing page's framework compatibility carousel

Woldbye:
Add Canvas dot-grid animation to the landing page's framework compatibility carousel.

**Notes:** 
* I can't use tailwindcss variables directly for the grid colors, as they don't work in the canvas. I could call `getComputedStyle` within the canvas, and resolve them before applying. But I think its fragile, and only two of the frameworks need the theme color, so I instead implemented it with hardcoded colors per theme, and the `useTheme` hook. 
* Could simplify the code substantially if I used framer.motion for the animations. Worth looking into in the future if we pull framer.motion in on the landing page.

## Demo
msedge_sqeXNA969A
msedge_ndVnhla62k

### DIFF
--- a/dashboard/src/app/[locale]/(public)/(landing)/components/FrameworkCard.tsx
+++ b/dashboard/src/app/[locale]/(public)/(landing)/components/FrameworkCard.tsx
@@ -2,22 +2,21 @@
 
 import { useState } from 'react';
 import Image from 'next/image';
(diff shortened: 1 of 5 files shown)


Woldbye:
Moved back to draft, as I think it will substantially cleaner to add this once the DotGridCarousel has been released to shadcn

PR #835, opened 2026-03-28
Error-tracking feature card

Lindharden:


flopper123 (changes requested):
Some minor UI feedback:
1. The text is a little too small / too gray to read on a white background
2. The text / icons should maybe be slightly larger

Image

PR #840, opened 2026-03-31
Session materialized view

flopper123:


migrations/32_update_sessions_settings.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE analytics.sessions
+    MODIFY SETTING min_rows_for_wide_part = 0, min_bytes_for_wide_part = 0;
```
Lindharden: We should try to avoid adjusting internal ClickHouse configurations needlessly. This will cause large overhead of write partitions for no benefit.

Lindharden (approved):

Lindharden (approved):

PR #841, opened 2026-04-01
japanese translations in progress

### LINKED ISSUE #842: Add Japanese as a supported language
Add Japanese as a supported language following the steps outlined in the parent issue.

Currently, all the existing locals are `LTR` text direction. So as part of adding Japanese, we need to ensure to also add support to dynamically switch the text direction per locale to support the Japanese `RTL` text direction.

evanyang1:


Woldbye:
I have made a new sub-issue for Japanese and assigned the PR to you. As I mention in the issue, its important that we also implement infrastructure to dynamically toggle text-direction per locale, as Japanese is `RTL`.

Let me know if you need some inspiration on how to solve this, or have any additional questions! :)

evanyang1:
Japanese is written right to left in books; however especially online, Japanese can also be written left to right and top to bottom just like in English

Issue #839, opened 2026-03-31
Global Event Properties


Woldbye:
Allow users to attach persistent key-value metadata to all analytics events - including built-in events like pageviews, scroll depth, web vitals, and error tracking.

Currently, custom properties can only be attached to individual custom events via `betterlytics.event()`. Users who want to segment analytics by dimensions like subscription plan, app version, or environment have no way to attach that context to their broader event stream aswell as events emitted from the script (page-view e.g.).

PR #846, opened 2026-04-03
Global Properties filter

### LINKED ISSUE #847: Global properties: backend ingestion and dashboard filter support
Accept global properties from the JS client, validate and store them per event, and expose them as a filterable dimension in the dashboard filter dropdown.

Users should be able to select a global property key from the filter menu and filter all dashboard views by its values - exactly how they filter by country, browser, or UTM parameters today.

Should include an update to documentation explaining the global properties usage and betterlytics client usage to our users.

Woldbye:
Closes #847 

First step towards the full Global Property feature #839. This PR adds global event properties support on backend, updates our analytics.js script to store them in-memory (and expose API to communicate), and extends our dashboard filters to support global properties. Further, it updates our docs on the new API for the client.

## Changes

  - JS client: `setGlobalProperties` / `getGlobalProperties` API with sessionStorage persistence across page loads
  - Backend (Rust): Accepts, validates (type, size, control chars), and stores `global_properties_json` as a String column — consistent with `custom_event_json`
  - Dashboard filtering: Global properties appear in the filter column dropdown via a drill-down submenu, with fully parameterized `JSONExtractString` / `JSONExtractKeys` queries
  - Change filter selector component to a Dropdown, and add focus handling for accessability
  - Active filter UX: Selected global property filters show a muted category label above the trigger and in active filter badges

## Demo

image


backend/src/validation/mod.rs
```diff
@@ -322,6 +334,53 @@ impl EventValidator {
         }
     }
 
+    fn validate_global_properties(&self, value: &serde_json::Value) -> Result<(), ValidationError> {
+        let obj = value.as_object().ok_or_else(|| {
+            ValidationError::InvalidGlobalProperties("Must be a JSON object".to_string())
+        })?;
```
Lindharden, line 340: We should ensure we also block empty strings, since they're currently valid and will go through. May also consider adding an explicit number range limit, since serde json number allows values like 1e308 or extremely long decimals. Dunno if it's a concern, but worth a thought 😄

backend/src/validation/mod.rs
```diff
@@ -30,6 +33,9 @@ impl Default for ValidationConfig {
             max_site_id_length: 100,                  // Site ID is usually short, but we should keep leeway for extra long domain names
             max_user_agent_length: 8 * 1024,          // 8192 bytes - same limit that apache uses (https://httpd.apache.org/docs/2.2/mod/core.html#limitrequestfieldsize)
             max_error_exceptions_size: 16 * 1024,     // 16KB - client caps stack at 10KB + type/value/mechanism overhead
+            max_global_properties_keys: 30,
+            max_global_property_key_length: 64,
+            max_global_property_value_length: 500,
```
Lindharden: I think we can significantly reduce this. 500 chars is a lot for a global property and will inflate all events stored by a significant amount, potentially exploitable by adversaries.

We could consider 128~ or even less than that

dashboard/src/components/filters/QueryFilterInputRow.tsx
```diff
@@ -54,49 +74,96 @@ export function QueryFilterInputRow<TEntity>({
...
-  const filterColumnRef = useRef<string>(filter.column);
+  const dashboardId = useDashboardId();
+  const analyticsQuery = useAnalyticsQuery();
+
+  const { data: globalPropertyKeys = [], isLoading: isLoadingPropertyKeys } = useQuery({
+    queryKey: ['global-property-keys', analyticsQuery.startDate?.toString(), analyticsQuery.endDate?.toString()],
+    queryFn: () => getGlobalPropertyKeysAction(dashboardId, analyticsQuery, { limit: 50 }),
+    staleTime: 5 * 60 * 1000,
+    gcTime: 5 * 60 * 1000,
+  });
```
Lindharden: This should be fetched once in QueryFiltersSelector and passed down as a prop.

dashboard/src/entities/analytics/filter.entities.ts
```diff
@@ -29,6 +30,7 @@ export const QueryFilterSchema = z.object({
   column: z.enum(FILTER_COLUMNS),
   operator: z.enum(FILTER_OPERATORS),
   values: z.array(z.string()),
+  propertyKey: z.string().optional(),
```
Lindharden: Would be nice to enforce propertyKey when column is 'global_property' and absent otherwise (at type level). Could potentially be solved relatively simple with a refine() on the current schema.

dashboard/src/services/analytics/filters.service.ts
```diff
@@ -1,14 +1,29 @@
...
   search?: string,
   limit?: number,
+  propertyKey?: string,
 ) {
+  if (column === 'global_property' && propertyKey) {
+    return getGlobalPropertyValues(siteQuery, propertyKey, search?.trim(), limit);
+  }
+  if (column === 'global_property') {
+    return [];
+  }
```
Lindharden:
```suggestion
if (column === 'global_property') {
  return propertyKey ? getGlobalPropertyValues(siteQuery, propertyKey, search?.trim(), limit) : [];
}
```

dashboard/src/utils/formatters.ts
```diff
@@ -60,6 +60,25 @@ export function capitalizeFirstLetter(value: string): string {
   return value.charAt(0).toUpperCase() + value.slice(1);
 }
 
+import type { FilterColumn } from '@/entities/analytics/filter.entities';
```
Lindharden: Move import to top of file 😄

dashboard/src/utils/formatters.ts
```diff
@@ -60,6 +60,25 @@ export function capitalizeFirstLetter(value: string): string {
...
+
+function formatNumericString(value: string, locale: SupportedLanguages): string {
+  const num = Number(value);
+  if (!isNaN(num) && value.trim() !== '') return formatNumber(num, locale, { maximumFractionDigits: 5 });
+  return value;
+}
+
+const FILTER_VALUE_FORMATTERS: Partial<Record<FilterColumn, FilterValueFormatter>> = {
+  global_property: formatNumericString,
+};
```
Lindharden: This will result in 1.000.000 becoming 1M, but the users property value is 1.000.000 not 1M. This is misleading since we don't know exactly what the user is storing in their global property and whether it should be formatted compactly.

static/analytics.js
```diff
@@ -124,6 +127,22 @@
...
+      if (props == null || typeof props !== "object" || Array.isArray(props)) {
+        return console.error("Betterlytics: setGlobalProperties requires a flat object");
+      }
+      globalProperties = {};
+      var keys = Object.keys(props);
+      for (var i = 0; i < keys.length; i++) {
+        var val = props[keys[i]];
+        if (["string", "number", "boolean"].includes(typeof val)) {
+          globalProperties[keys[i]] = val;
+        }
```
Lindharden: Might be worth to add a console.warn when the value is not a string, number or boolean to give them some indicator, instead of making them debug why it's not properly registered 😄

static/analytics.js
```diff
@@ -124,6 +127,22 @@
         is_custom_event: true,
         properties: JSON.stringify(eventProps),
       }),
+    setGlobalProperties: function (props) {
```
Lindharden, line 130: I know it's a design choice that a second setGlobalProperties call would overwrite existing properties, but it's not obvious from the name. Might consider renaming this, or alternatively providing a separate clearGlobalProperties() and making setGlobalProperties merge would be more ergonomic.

static/analytics.js
```diff
@@ -107,6 +109,7 @@
         user_agent: userAgent,
         screen_resolution: screenResolution,
         timestamp: Math.floor(Date.now() / 1000),
+        global_properties: Object.assign({}, globalProperties),
```
Lindharden: If we make this `global_properties: Object.keys(globalProperties).length ? Object.assign({}, globalProperties) : undefined`, then we would omit the property entirely when it's not set. This will save bandwith and reduce the payload that needs processing in general 😄

Lindharden (changes requested):
I left a few comments on your PR. Seems like cursor-pointer was also lost on the filter rows for some reason 😄 

Seems to work nicely 😄

static/analytics.js
Woldbye: Could do that but it would force us to handle undefined everywhere in the use chain. Do you Think its worth it to save a few bytes?

dashboard/src/entities/analytics/filter.entities.ts
Woldbye: Remove the property like we discussed instead

backend/src/validation/mod.rs
Woldbye: Agreed. I have ensured it blocks empty strings, and added explicit limits matching of `2^53-1` ([`Number.MAX_SAFE_INTEGER`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER)).

backend/src/validation/mod.rs
Woldbye: I changed it to 128. I think its a good default.

dashboard/src/components/filters/QueryFilterInputRow.tsx
Woldbye: Done! Though note that `useQuery` already ensured only one query were fired anyway - So this is strictly for cleaner architecture.

dashboard/src/services/analytics/filters.service.ts
Woldbye: Removed the `propertyKey` all together.

dashboard/src/utils/formatters.ts
Woldbye: Nice catch. I think you are right that its bad practise to use compact. I changed it to use "standard" instead which is a better fit if you ask me.

static/analytics.js
Woldbye: Good idea! Done exactly as you requested.

static/analytics.js
Woldbye: I did what you suggested, the serde validation seemed to do the trick already

Woldbye:
> I left a few comments on your PR. Seems like cursor-pointer was also lost on the filter rows for some reason 😄
> 
> Seems to work nicely 😄

I readded the cursor-pointer, and attempted to resolve the items you listed! Please have a look when you can find the time for a new review.

static/analytics.js
Woldbye: I made a clearGlobalProperties and merge as you requested. Hope its clearer to our users :)

dashboard/src/components/filters/QueryFiltersSelector.tsx
```diff
@@ -26,6 +30,15 @@ export default function QueryFiltersSelector() {
   const [isSavedFiltersOpen, setIsSavedFiltersOpen] = useState(false);
   const isMobile = useIsMobile();
   const t = useTranslations('components.filters');
+  const dashboardId = useDashboardId();
+  const analyticsQuery = useAnalyticsQuery();
+
+  const { data: globalPropertyKeys = [], isLoading: isLoadingPropertyKeys } = useQuery({
+    queryKey: ['global-property-keys', analyticsQuery.startDate?.toString(), analyticsQuery.endDate?.toString()],
```
Lindharden: The dashboardId is absent from the cache key. Navigating between two dashboards with different date ranges but the same dates could serve stale property keys from a different site

Lindharden (dismissed):
Looks good. Only one minor comment left 😄

dashboard/src/components/filters/QueryFiltersSelector.tsx
Woldbye: You're absolutely right. I have done as you suggested! :)

flopper123 (approved):
Awesome 😄

PR #856, opened 2026-04-09
Global Properties - Dashboard drilldown on Overview

### LINKED ISSUE #848: Global properties: overview tab on dashboard
Add a dedicated section (or alternatively a tab to the existing event section) to the dashboard overview that surfaces the most common global property keys and their value distributions. This gives users a quick summary of how their traffic breaks down by the dimensions they've defined, without needing to manually add filters.

The value distributions should be clickable, whereas a click should select that specific filter (like how our other section cards work on the dashboard).

Woldbye:
Adds a "Top Global Properties" tab to the overview dashboard, showing the most used global property keys with expandable value breakdowns. Clicking a property or value applies the corresponding gp.* filter.

## Changes

  1. New ClickHouse repository (`globalProperties.repository.ts`) querying key-value counts via ARRAY JOIN on `global_properties_json`
  2. __IMPORTANT__ The nested items on the multitable drilldowns are now limited to show max 5 items in view, but up to N. I have used the radix scroll section for the UI. This also affects the browser and devices tabs on the overview.
  3. Service layer (`globalProperties.service.ts`) groups flat rows into a key -> values hierarchy with configurable limits
  4. Presenter (`toGlobalPropertiesDataTable.ts`) pairs current/compare periods and computes percent change
  5. `CustomEventsSection` extended with a second tab fetching events and properties in a single parallel action -- Would love to know if you agree its the right move to fetch in a single parallel action.
  6. `MultiProgressTable` children sections wrapped in a scrollable area for long value lists
  7. `isFilterColumn` type guard added to `filter.entities.ts`, replacing ad-hoc casts when applying filters from item clicks

## Demo
image


dashboard/messages/en.json
```diff
@@ -1617,7 +1617,7 @@
       "geography": "Geography",
       "devicesBreakdown": "Devices",
       "trafficSources": "Traffic Sources",
-      "customEvents": "Custom Events",
+      "customEvents": "Events",
```
Lindharden: I'm uncertain about this change, can you argue why you did this? 😄

dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/(overview)/CustomEventsSection.tsx
```diff
@@ -1,44 +1,84 @@
...
+      label: t('tabs.events'),
+      loading: eventsState.loading,
+      data: (eventsQuery.data ?? []).map((event) => ({
+        label: event.event_name,
+        value: event.current.count,
+        trendPercentage: event.change?.count,
+        comparisonValue: event.compare?.count,
+      })),
+    },
+    ...(properties.length > 0
```
Comment (Lindharden) [inline dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/(overview)/CustomEventsSection.tsx]: We should always display the tab, even if there are no events. Similarly to how we always display the custom events tab 😄

dashboard/src/presenters/toGlobalPropertiesDataTable.ts
```diff
@@ -0,0 +1,62 @@
...
+      ? new Map(compareProp.values.map((v) => [v.value, v]))
+      : undefined;
+
+    const children: MatchedPropertyValue[] = prop.values.map((v) => {
+      const cv = compareValueMap?.get(v.value);
+      return {
+        value: v.value,
+        current: { count: v.count },
+        compare: cv ? { count: cv.count } : undefined,
+        change: computeChange ? { count: percentChange(v.count, cv?.count ?? 0) } : undefined,
```
Lindharden: The field's name is misleading, it holds a percentage change and not a count 😄

dashboard/src/services/analytics/globalProperties.service.ts
```diff
@@ -0,0 +1,46 @@
+'server-only';
+
+import {
+  getTopGlobalPropertyKeys,
+  getTopGlobalPropertyValuesForKeys,
+} from '@/repositories/clickhouse/globalProperties.repository';
+import { BASiteQuery } from '@/entities/analytics/analyticsQuery.entities';
+
+const DEFAULT_KEY_LIMIT = 10;
+const DEFAULT_VALUE_LIMIT = 20;
```
Lindharden: Might make sense to move to the API router and pass it from there 😄

Lindharden (changes requested):

dashboard/src/repositories/clickhouse/globalProperties.repository.ts
Lindharden, line 1: I think these should be filtered to either unique visitors or sessions, so that the count of each property is not inflated by internal events such as scroll_depth, CWV, outbound links and so on 😄 

We should therefore decide whether this should answer as an user-level property, "how mayn users have plan=pro", which is typically the most natural response to expect. Or whether it should be session-based, which becomes usesful as properties can change between sessions.

Lindharden (changes requested):

dashboard/messages/en.json
Woldbye: The name of the custom events tab stays the same - this is the name for the shared multi table of custom events and the global properties table. The old value were moved to the "events" key, as this is now used as the general name for the table. You could argue that it's cleaner to inverse the i18n keys, but I did it this way to reduce the actual code diff

dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/(overview)/CustomEventsSection.tsx
Comment (Woldbye) [reply dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/(overview)/CustomEventsSection.tsx]: I feel that the global properties is more of an opt in feature so it made sense to conditionally toggle it. Can easily change the behavior

dashboard/src/services/analytics/globalProperties.service.ts
Woldbye: Agree :) let's do that, but won't do anything till we have resolved what we actually want to display here :D (your final comment on the pr)

dashboard/src/repositories/clickhouse/globalProperties.repository.ts
Woldbye: > I think these should be filtered to either unique visitors or sessions, so that the count of each property is not inflated by internal events such as scroll_depth, CWV, outbound links and so on 😄 
> 
> We should therefore decide whether this should answer as an user-level property, "how mayn users have plan=pro", which is typically the most natural response to expect. Or whether it should be session-based, which becomes usesful as properties can change between sessions.

This is the right question!! I opened this pr precisely to raise this question as I felt myself that the events model were suboptimal.

Maybe the best solution is to create an entirely new global properties section instead of nesting it inside the events section. Then we can give it three tabs with global property overviews of: user, events, sessions. Wdyt of this ?

Lindharden: Just a quick update from our in-person discussion. We'll count global properties for unique visitors because this includes value changes and neither overcounts nor undercounts segmented visitors 😄

dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/(overview)/CustomEventsSection.tsx
```diff
@@ -1,44 +1,80 @@
...
 import { trpc } from '@/trpc/client';
 import { useQueryState } from '@/hooks/use-query-state';
+import { isFilterColumn } from '@/entities/analytics/filter.entities';
 
 export default function CustomEventsSection() {
   const { input, options } = useBAQueryParams();
-  const query = trpc.events.customEventsOverview.useQuery(input, options);
-  const eventState = useQueryState(query);
+  const eventsQuery = trpc.events.customEventsOverview.useQuery(input, options);
+  const propertiesQuery = trpc.events.globalPropertiesOverview.useQuery(input, options);
```
Comment (flopper123) [inline dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/(overview)/CustomEventsSection.tsx]: These tabs should be Lazy Loaded like the rest, you can draw inspiration from `dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/(overview)/DevicesSection.tsx`

dashboard/src/components/MultiProgressTable.tsx
```diff
@@ -162,9 +166,15 @@ function MultiProgressTable<T extends ProgressBarData>({
...
-                  </div>
+                  <ScrollAreaPrimitive.Root
+                    className='mt-2 ml-4 max-h-46 border-l'
+                    onClick={(e) => e.stopPropagation()}
+                  >
+                    <ScrollAreaPrimitive.Viewport className='max-h-46'>
+                      {renderProgressList(children as T[], tabKey, level + 1)}
+                    </ScrollAreaPrimitive.Viewport>
+                    <ScrollBar />
+                  </ScrollAreaPrimitive.Root>
```
flopper123: Question: What are these changes for?

dashboard/src/repositories/clickhouse/globalProperties.repository.ts
```diff
@@ -0,0 +1,107 @@
...
+export type GlobalPropertyKeyCountRow = {
+  property_key: string;
+  visitors: number;
+};
+
+export type GlobalPropertyKeyValueRow = {
+  property_key: string;
+  value: string;
+  visitors: number;
+};
```
flopper123: These should come from the Zod schema 😄

dashboard/src/repositories/clickhouse/globalProperties.repository.ts
```diff
@@ -0,0 +1,107 @@
...
+        end: endDateTime,
+        key_limit: keyLimit,
+      },
+    })
+    .toPromise()) as GlobalPropertyKeyCountRow[];
+
+  return rows.map((row) => ({
+    property_key: row.property_key,
+    visitors: Number(row.visitors),
+  }));
```
flopper123: Missing Zod validation on both queries

dashboard/src/repositories/clickhouse/globalProperties.repository.ts
```diff
@@ -0,0 +1,107 @@
...
+      uniq(visitor_id) AS visitors
+    FROM analytics.events
+    ARRAY JOIN global_properties_keys AS property_key
+    WHERE site_id = {site_id:String}
+      AND timestamp BETWEEN {start:DateTime} AND {end:DateTime}
+      AND length(global_properties_keys) > 0
+      AND ${SQL.AND(filters)}
+    GROUP BY property_key
+    ORDER BY visitors DESC
+    LIMIT {key_limit:UInt32}
```
flopper123, line 31: Question: Should these be done as rollup queries like `getOperatingSystemRollup` in `dashboard/src/repositories/clickhouse/devices.repository.ts`? That way it'll be a single query instead of being two sequential queries 😄

dashboard/src/services/analytics/globalProperties.service.ts
```diff
@@ -0,0 +1,43 @@
...
+    const list = valuesByKey.get(row.property_key) ?? [];
+    list.push({ value: row.value, visitors: row.visitors });
+    valuesByKey.set(row.property_key, list);
+  }
+
+  return topKeys.map(({ property_key, visitors }) => ({
+    property_key,
+    visitors,
+    values: valuesByKey.get(property_key) ?? [],
+  }));
```
flopper123: Quibble:
```suggestion
  return topKeys.map((topKey) => ({
    ...topKey,
    values: valuesByKey.get(topKey.property_key) ?? [],
  }));
```

flopper123 (changes requested):
Works very nice 😄
Only a few code comments

dashboard/src/components/MultiProgressTable.tsx
Woldbye: reverted, I can see main added a scroll-overflow-y on each multi-tab-table, so this is no longer necessary

dashboard/src/repositories/clickhouse/globalProperties.repository.ts
Woldbye: probably :D Feel free to change them, I am not sure how to do it cleanly

Lindharden (dismissed):
Awesome 😄

flopper123 (dismissed):

Lindharden (approved):

PR #859, opened 2026-04-12
Bug geo tooltip

### LINKED ISSUE #858: Map tooltip on the overview section stays open while hovering the legend
The geography map tooltip on the overview section stays open while hovering the legend.

When the legend is hovered the hover tooltip should close.

https://github.com/user-attachments/assets/95f1c725-73e9-4ba2-8940-c388b5ed030b

Woldbye:
Closes #858 

Fix 2 bugs with geography tooltip causing it to stay open:

1. on fast mouse-movements or scrolls it would stay stuck on the screen even though mouse is outside map -- I resolved this by adding a listener on the window. I couldnt find a more light weight way to do it sadly, but it only runs when map is mounted.
2. on hover of legend, it would stay open -- I resolved this by adding a listener to the legend.


flopper123 (approved):
very nice

PR #861, opened 2026-04-19
Sanitize incoming events for global properties

flopper123:
Added backend sanitization module instead of dropping all events in invalid global properties

### DIFF
--- a/docs/src/content/integration/global-properties.mdx
+++ b/docs/src/content/integration/global-properties.mdx
@@ -44,7 +44,7 @@ betterlytics.setGlobalProperties({
 });
 ```
 
-The following values are rejected by the client with a console warning:
+The following values are silently dropped on the server:
 
 ```javascript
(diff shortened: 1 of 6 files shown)


docs/src/content/integration/global-properties.mdx
```diff
@@ -44,7 +44,7 @@ betterlytics.setGlobalProperties({
 });
 ```
 
-The following values are rejected by the client with a console warning:
+The following values are silently dropped on the server:
```
Lindharden: "on the server" is not relevant for the user to know. They just need to know it's excluded or dropped somewhere, not where 😄

```suggestion
Global properties with the following values are ignored and not included in events:
```

Lindharden (changes requested):
I think we should add sanitization of empty string keys, simply because they provide no value and would otherwise clutter the ClickHouse index and unions 😄

Lindharden (approved):

PR #869, opened 2026-04-20
Added missing backfill statement for daily usage MV & Updated pricing…

Lindharden:
… docs to reflect usage measurements


flopper123 (approved):
Awesome

PR #878, opened 2026-04-26
Add global properties to funnel

### LINKED ISSUE #855: Global Properties: Add global property filters to the Funnels
The funnels UI filter is seperate from the FilterQuery on the rest of the dashboards. The Funnels should also get the new global-property filtering available for the funnel steps.

To support global property in funnel filters, I suggest that we add a reusable base that both the FunnelFilter UI and dashboard can share,

Woldbye:
Adds Global properties to our funnels. To implement the refactor cleanly I have unified funnels and the ordinary filter to use the same base component. I have also adjusted the filter layout to handle our 10 saved filter limit + multiple rows, abstracted the old filter to reusable components and fixed bugs and accesability issues on the filter component.

So when testing this:
1. Create 10 saved filters, so you can see the ui difference
2. Add rows until it overflows
3. Try the new layout on multiple screen sizes


flopper123 (changes requested):
UX feedback:
1. When creating your filters, if you accidentally miss-click such that it closes the pop-over, it clears all your filters. Previously, it didn't clear on close.

2. The popover layout on laptop sized screens shifts when selecting a filter colum:Not open:

Image

Open:
Image

3. The line for the saved filters is now shown, even when no saved filters exists:

Image

4. Gaps / paddings have been greatly increased on the filters:

Image

5. The open state of the saved filters are no longer remembered when reopening the popover:

Image

6. By adding the new icon to the input, as since the global properties have their own icon, that makes "Global property" text above the field obsolete:
Image

 7. The cancel button is highlighted and clickable, even when there's nothing to cancel

Image

8. Accidentally closing the Create Funnel modal while setting it up, then clicking Create Funnel again now clears whatever you were working on.

9. On hover, the filter doesn't specify the "Global Property" as everywhere else:
image

Woldbye:
Thank you for the review. I do apologize for regressions and have attempted to resolve them.

> 1. When creating your filters, if you accidentally miss-click such that it closes the pop-over, it clears all your filters. Previously, it didn't clear on close.

Sorry I misunderstood the actual UX, I thought it was a bug not a feature 👎 
I have restored the persistence behavior of main on close


> 2. The popover layout on laptop sized screens shifts when selecting a filter colum:Not open:

This is due to the `collisionPadding: 16`. I restored the `collisionPadding: 0` of main (default of the popover component).


> 3. The line for the saved filters is now shown, even when no saved filters exists:

Nice catch! I have reverted this, and conditionally hidden it - exactly like on main


> 4. Gaps / paddings have been greatly increased on the filters:

I believe I have restored the exact same gaps / paddings / margins as on main on the filters now.


> 5. The open state of the saved filters are no longer remembered when reopening the popover:

I am not sure how this differs from 1? As mentioned above I restored the persistence behavior of main.


> 6. By adding the new icon to the input, as since the global properties have their own icon, that makes "Global property" text above the field obsolete:

Great idea!! I have nuked the invasive global property subtitle's completely.


> 7. The cancel button is highlighted and clickable, even when there's nothing to cancel

The disabled/active state of the apply button already conveys whether there's changes to be applied. So I think it makes a lot more sense to always show the cancel button, as its not clear why it would be disabled. Though I have reverted this, so it matches the state of main (disabled state follows the disabled state of active, i.e. whenever its not dirty disable)

> 8. Accidentally closing the Create Funnel modal while setting it up, then clicking Create Funnel again now clears whatever you were working on.

Sorry I misunderstood the actual UX, like with 1). I have restored the persistence behavior from main so it behaves the same.


> 9. On hover, the filter doesn't specify the "Global Property" as everywhere else:

I agree this is not optimal, but is it acceptable for now? I think the correct play here is to create a universal solution that incorporates the filter Icon's everywhere. As such, I would prefer to handle this in a follow up PR.

flopper123 (changes requested):
UX feedback:
1. Empty value row is longer than when values are there now: 

Image

Whereas with values they then re-align:

Image

flopper123 (approved):

Lindharden (approved):

PR #880, opened 2026-04-27
Page table optimizations

flopper123:


backend/src/processing/mod.rs
```diff
@@ -67,6 +67,8 @@ pub struct ProcessedEvent {
     pub error_fingerprint: String,
     pub global_properties_keys: Vec<String>,
     pub global_properties_values: Vec<String>,
+    /// Duration for engagement events
+    pub duration_seconds: u32,
```
Woldbye: Suggestion: engagement_duration_seconds seems like a nice selfexplanatory variable name

static/analytics.js
```diff
@@ -290,40 +295,38 @@
...
   }
 
   window.addEventListener("scroll", () => updateScrollDepth(), {
     passive: true,
   });
 
-  document.addEventListener(
-    "visibilitychange",
-    () => document.visibilityState === "hidden" && flushScrollDepth(),
-  );
```
Woldbye, line 325: Removing the visibilitychange event listener will break this for mobile users. I highly advice still flushing scroll-depth on visibilitychange, so it doesnt break for SPA, native or mobile web users.

NOTE: I think you removed this to avoid double flushing (i.e. counting duration twice). If you reintroduce the visibilitychange event-listener this is a real concern. But the mitigation should not be removing the visibilitychange listener, but instead making flushEngagement idempotent. Main's implementation previously had lastSentScrollDepthPx exactly to make it idempotent.

migrations/33_engagement_events.sql
Woldbye, line 1: This migration is irreversible and unguarded. If it partially fails and reruns, we will get duplicate engagement_events. Not sure how best to resolve this, but I do think it needs to be resolved.

dashboard/src/services/analytics/pages.service.ts
```diff
@@ -67,19 +62,22 @@ export async function getExitPageAnalyticsForSite(siteQuery: BASiteQuery): Promi
...
+  const [pageviewsChartData, dailyAvgTimeData, dailyBounceRateData, sessionMetricsData] = await Promise.all([
+    getTotalPageViewsForSite(siteQuery),
+    getDailyAverageTimeOnPageForSite(siteQuery),
+    getDailyBounceRateForSite(siteQuery),
+    getSessionMetrics(siteQuery),
+  ]);
+
+  const totalPageviews = pageviewsChartData.reduce((sum, row) => sum + row.views, 0);
+  const activeDays = dailyAvgTimeData.filter((row) => row.avgTime > 0).length;
+  const avgTimeOnPage =
```
Woldbye, line 81: The new avgTimeOnPage metric is different from the one on main, meaning users history and displayed values will change. 

We need to either document this, or keep the behavior of main

dashboard/src/services/analytics/pages.service.ts
```diff
@@ -67,19 +62,22 @@ export async function getExitPageAnalyticsForSite(siteQuery: BASiteQuery): Promi
...
+    activeDays > 0 ? dailyAvgTimeData.reduce((sum, row) => sum + row.avgTime, 0) / activeDays : 0;
+  const activeBounceRateDays = dailyBounceRateData.filter((row) => row.bounceRate > 0).length;
+  const avgBounceRate =
```
Woldbye: The avgBounceRate metric is different from the one on main, meaning users history and displayed values will change.

We need to either document this, or keep the behavior of main

static/analytics.js
```diff
@@ -77,10 +78,10 @@
 
   var globalProperties = {};
 
-  // Scroll depth tracking state
+  // Engagement tracking state (duration + scroll depth)
+  var pageStartTime = Date.now();
```
Woldbye: I did some local testing, and I definintly think we should change to `performance.now()` (monotonic clock) instead of `Date.now()` (system clock). The system clock can drift or even be changed from user actions. I manually changed it in my tests to an earlier time of day, which made the events produce a negative duration, though the guard `duration > 0` caught this, and it was simply rejected. Important to note, that users can also roll up their clock, causing the opposite effect, and the guard wont catch this. Sleep/Wake on some windows systems are also infamous for rolling up the clock, so I do think this is important, and not just some rare edge case.

WDYT, is this worth opening an issue about to fix in a follow up?

Woldbye (commented):
PAGES PAGE OPTIMIZATIONS YAY. 

Had a chance to go through the code, the changes are not really my expertise, but had a few things I thought was worth flagging to double check.

**Additional notes:**
- It seems like the `backend/src/validation/mod.rs` only applies the validation for `event_name == "scroll-depth"`. The new client sends `event_name: "engagement"`. I am pretty sure this means the Rust validation is skipped. Meaning users, can send whatever they want polluting the data.
- *Quibble* - the `avg_scroll_depth` MCP metric still queries `scroll_depth_percentage` from all events. Should probably update it as part of this PR.

migrations/33_engagement_events.sql
Lindharden: If the migration fail, the data and the schema will be rolled back to a pre-migration state, so it should be fine 😄

static/analytics.js
Lindharden: Good suggestion 👍

static/analytics.js
Lindharden: Awesome. I've investigated and we can safely implement this 😄

dashboard/src/services/analytics/pages.service.ts
flopper123: Nice catch! It's now properly weighing the different days compared to the number of engagement events 😄

dashboard/src/services/analytics/pages.service.ts
flopper123: Nice catch! It's now properly weighing the different days compared to the number of sessions 😄

Woldbye (dismissed):
The new queries look solid, and seems to me like the math now holds statistically!

Lindharden (approved):

PR #888, opened 2026-05-01
Consolidated scheduler into pg boss embedded worker

Lindharden:


dashboard/src/worker/workerEnv.ts
```diff
@@ -3,14 +3,19 @@ import { z } from 'zod';
 const schema = z.object({
   POSTGRES_URL: z.string().min(1),
   CLICKHOUSE_URL: z.string().url(),
-  WORKER_CLICKHOUSE_WRITE_USER: z.string().min(1),
-  WORKER_CLICKHOUSE_WRITE_PASSWORD: z.string().min(1),
+  WORKER_CLICKHOUSE_WRITE_USER: z.string().optional().default(''),
```
Woldbye, line 6: what motivated the change to default('')? My intuition says its a bad idea, as it will throw weird errors run-time, instead of fail fast and hard on build (if the environments are missing)

dashboard/src/instrumentation.ts
```diff
@@ -4,14 +4,15 @@ export async function register() {
 }
 
 async function registerBackgroundJobs() {
-  if (process.env.NEXT_RUNTIME === 'nodejs') {
+  if (process.env.NEXT_RUNTIME == 'nodejs') {
```
Woldbye:
```suggestion
  if (process.env.NEXT_RUNTIME === 'nodejs') {
```

I think you may have made a typo changing to weak comparison here

Woldbye (dismissed):
Looks good!! I found a single small code change I would like you to double check (change to weak comparison), rest looks great!

dashboard/src/worker/workerEnv.ts
Lindharden: The motivation was to let to let the new embedded pg-boss worker boot without requiring WORKER_CLICKHOUSE_WRITE_USER / WORKER_CLICKHOUSE_WRITE_PASSWORD because self-hosters will not be running background jobs with clickhouse write access 😄

flopper123 (approved):
Awesome!

PR #891, opened 2026-05-02
Add radix scrollbar to multi progress table

Woldbye:
Also solves the skeleton scrollbar bug by design

msedge_YebXHCbxel


Woldbye:
**Question**: Should we apply the radix styling universally for all scroll-bar consumers??

Lindharden (changes requested):
Scrollbar seems nice, but it causes scrollbar to mismatch with all other scrollbars used throughout the site 😿 

I'd rather we extend the scroll-area for consistensy 😄 This way we may be able to avoid hooking into radix primitives directly

Woldbye:
@Lindharden 

I have went through your comments.
1. I have removed the custom styling all together, and now use the ScrollArea directly. Instead we can add the styling in a follow-up if we want it :)

2. Unify the shrink-0, so all branches of OSIcon and BrowserIcon get it.

flopper123 (approved):
Nice 😄

Lindharden (approved):
Looks good 😄

PR #892, opened 2026-05-02
Fix firefox text-selection bug on tables

### LINKED ISSUE #860: Cannot copy or export data
I am trying to make use of the data shown on the Referrers page, in the lower module Referrer Details. However, I can't find a way to export the data.
When trying to just copy what appears to be a table with the data, all I get is the following:

`Referrer Details`
`other | 12vs 0 | 3% | 8.3%vs 0% | 1m 12svs 0s |  `
`other | 5vs 0 | 1.3% | 0%vs 0% | 4m 42svs 0s |  `
`other | 4vs 0 | 1% | 0%vs 0% | 7m 20svs 0s |  `
`other | 4vs 0 | 1% | 0%vs 0% | 2m 23svs 0s | `

As you can see, the column for Source is not included.

Woldbye:
Closes #860

Resolve the bug with text-selections of table columns in Firefox.

The first column in these tables wraps its label in a `<Button variant='ghost'>` to make it click-to-filter. Firefox's user-agent stylesheet sets `-moz-user-select: none` on `<button>` elements by default. Chromium-based browsers and Safari don't apply this rule, which is why the bug were only reproduceable in Firefox.

## Fix

Added the select-text Tailwind class (user-select: text) to the click-to-filter Button in each of the 5 affected DataTable cells.


Lindharden:
Awesome! Works like a charm!

Lindharden:
I have fixed merge conflicts 👍

Lindharden (approved):

PR #895, opened 2026-05-03
SuperAdmin dashboard

Lindharden:


dashboard/src/actions/superadmin/users.action.ts
```diff
@@ -0,0 +1,25 @@
...
+  'user',
+  async (userId: string) => ({ targetId: userId, payload: { deletedUserId: userId } }),
+  async (ctx: SuperAdminCtx, userId: string) => {
+    if (userId === ctx.actorUserId) {
+      throw new UserException('Super admins cannot delete themselves');
+    }
+    if (isSuperAdmin(userId)) {
+      throw new UserException('Super admins cannot delete other super admins');
+    }
+    await removeUser(userId);
```
flopper123, line 23: When a user deletes their own account, we also delete their owned dashboards and pending invitations.
Should this do the same for consistency?

dashboard/src/app/robots.ts
```diff
@@ -20,7 +20,7 @@ export default function robots(): MetadataRoute.Robots {
       {
         userAgent: '*',
         allow: '/',
-        disallow: ['/dashboard/*', '/dashboards/*', '/billing/*', '/api/*'],
+        disallow: ['/dashboard/*', '/dashboards/*', '/billing/*', '/api/*', '/admin', '/admin/*'],
```
flopper123: The admin page is on the "locale" path, so this disallow does not disallow `/da/admin` for instance.

Maybe we should move the admin page outside of locale?

dashboard/src/repositories/clickhouse/sessions.repository.ts
```diff
@@ -1,4 +1,4 @@
-'server only';
+import 'server-only';
```
flopper123, line 1: This should probably just be `'server-only'` yeah? 🤔 

Same goes for the other repositories

flopper123 (changes requested):

dashboard/src/actions/superadmin/users.action.ts
Woldbye: I think it MUST do the same, since juts deleting the account will create orphaned dashboards if the deleted user is an owner of a dashboard shared by multiple tenants.

dashboard/src/auth/superAdmin-auth.ts
```diff
@@ -0,0 +1,19 @@
...
+export function isSuperAdmin(userId: string | undefined): boolean {
+  return !!userId && env.SUPER_ADMIN_USER_IDS.has(userId);
+}
+
+export function assertSuperAdmin(session: Session | null): asserts session is Session {
+  if (!session?.user?.id || !isSuperAdmin(session.user.id)) {
+    throw new ForbiddenError();
+  }
+
+  if (!env.IS_DEVELOPMENT && !session.user.totpEnabled) {
```
Woldbye, line 16: I am fine with this, but it should be noted that the `session.user.totpEnabled` guard doesn't stop user's who have TOTP enabled on their account, but has logged in with OIDC in the current session (i.e. without 2FA). 

The `totpEnabled` is set once, when a user completes the 2FA setup. Its never restamped on OIDC logins, I think this means that:

1. User signs in via github oauth flow
2. User enables 2FA in their settings
3. User gets promoted to SUPER USER (added to SUPER_ADMIN_IDS)
4. User now has `session.user.totpEnabled` even when he logs in with an OAuth provider (i.e. without password or 2FA)

We control who becomes super-users, so maybe this is not a real issue, but worth highlighting the logic problem atleast.

dashboard/src/lib/env.ts
```diff
@@ -68,7 +68,12 @@ const envSchema = z.object({
   INTEGRATION_ENCRYPTION_KEY: z.string().length(32),
   SAMPLING_TRAFFIC_THRESHOLD: z.coerce.number().optional().default(100_000),
   SAMPLING_FACTOR: z.coerce.number().min(0).max(1).optional().default(0.25),
-  HIGH_TRAFFIC_CONCURRENCY_LIMIT: z.coerce.number().optional().default(20)
+  HIGH_TRAFFIC_CONCURRENCY_LIMIT: z.coerce.number().optional().default(20),
+  SUPER_ADMIN_USER_IDS: z
```
Woldbye, line 65: I'm not so sure its a good idea to hoist the super_admin_user_ids into the .env?? Doesn't this mean they are parsed once at module load => We would have to redeploy to update the super_admin_user_ids

dashboard/src/middlewares/withSuperAdminAction.ts
```diff
@@ -0,0 +1,105 @@
...
+    try {
+      const session = await getCachedSession();
+      assertSuperAdmin(session);
+
+      ctx = { actorUserId: session.user.id };
+      const resolvedAudit = await audit(...args);
+      auditInfo = resolvedAudit;
+
+      const result = await fn(ctx, ...args);
+      await recordAuditResult(ctx.actorUserId, action, targetType, resolvedAudit, 'success');
```
Woldbye: If `await recordAuditResult(ctx.actorUserId, action, targetType, resolvedAudit, 'success');` throws but the `recordAuditResult` succeeds, we would falsely be auditing the result as a failure, even though the `await audit(...args)` succeeded. The fix is probably as easy as lifting the success `resolvedAuditResult` outside this try { ... } catch { ... }

dashboard/src/repositories/clickhouse/sessions.repository.ts
Lindharden: It needs to be import 😄

dashboard/src/lib/env.ts
Lindharden: Yes it requires a redeploy, but we don't expect to be adding any super admins 😄

Woldbye (dismissed):

flopper123 (dismissed):

flopper123 (approved):

PR #897, opened 2026-05-03
Improve correctness & reliability of referrer classification

### LINKED ISSUE #882: Improvements to referrer channels based on domains
Currently, the following domains are listed as "Other" in my dashboard:

linkedin.com
com.linkedin.android
ecosia.org
com.google.android.gm
com.google.android.googlequicksearchbox
facebook.com

May I request the following channel categorisation instead:

SOCIAL
linkedin.com
com.linkedin.android
facebook.com

SEARCH
ecosia.org
com.google.android.gm
com.google.android.googlequicksearchbox

Lindharden:
This PR improves referrer channel classification by seeding a ClickHouse dictionary from three sources:

- Snowplow's referrer database
- GA4's maintained source category list
- A small Betterlytics custom override list

The dictionary is synced on backend startup, which lets us update classifications through asset changes without rewriting historical events. Query-time classification now uses `referrer_source_effective`, which resolves the best available key in order: canonical source, referrer URL host, then source name, before falling back to the originally ingested source (most likely 'Other' in this case).

This fixes cases where known referrers were previously shown as `Other`, including Android app referrers and domains missing from Snowplow, while preserving the original referrer URL and raw ingested source.

Closes #882


backend/src/main.rs
```diff
@@ -105,6 +105,21 @@ async fn main() {
...
         .expect("Failed to initialize database");
     db.validate_schema().await.expect("Invalid database schema");
+
+    if let Err(e) = db
+        .sync_referrer_categories(
+            &config.referrer_db_path,
+            &config.ga4_source_categories_path,
+            &config.custom_referrers_path,
+        )
+        .await
```
flopper123: Suggestion: Consider calling `build_referrer_source_categories` out here here, and pass that to `sync_referrer_categories` 😄

backend/src/db/models.rs
```diff
@@ -138,3 +140,10 @@ impl EventRow {
         }
     }
 }
+
+#[derive(clickhouse::Row, Serialize, Debug)]
+pub struct ReferrerSourceCategoryRow {
+    pub generation: u64,
+    pub key: String,
+    pub medium: String,
+}
```
flopper123: Quibble: Move this above the `impl` to struct group 😄

backend/src/referrer/mod.rs
```diff
@@ -145,15 +152,34 @@ pub fn parse_referrer(referrer: Option<&str>, current_url: Option<&str>) -> Refe
...
+        "android-app" => referrer_url
+            .host_str()
+            .and_then(|h| Url::parse(&format!("http://{}/", h)).ok()),
+        _ => {
+            let host = referrer_url.host_str().unwrap_or("");
+            let normalized_host = host.strip_prefix("www.").unwrap_or(host);
+            Url::parse(&format!("http://{}{}", normalized_host, referrer_url.path())).ok()
+        }
+    };
+    let referrer_info = lookup_url.as_ref().and_then(|u| parser.lookup(u));
```
flopper123: We could maybe generalize app scheme. That would automatically catch other app schemes like `ios-app://` or `intent://`, and would simplify branches

```suggestion
    let parser = get_parser();
    let host = referrer_url.host_str().unwrap_or("");
    let normalized_host = host.strip_prefix("www.").unwrap_or(host);
    let path = if matches!(referrer_url.scheme(), "http" | "https") {
        referrer_url.path()
    } else {
        ""
    };
    let lookup_url = Url::parse(&format!("http://{}{}", normalized_host, path)).ok();
    let referrer_info = lookup_url.as_ref().and_then(|u| parser.lookup(u));
```

backend/src/referrer/mod.rs
```diff
@@ -197,9 +223,13 @@ pub fn parse_referrer(referrer: Option<&str>, current_url: Option<&str>) -> Refe
...
-    // Always use the base domain as the source name for consistent grouping
-    let referrer_name = referrer_url.host_str()
-        .and_then(|h| extract_root_domain(h));
+    // For android-app:// referrers the host is a package name (e.g. com.linkedin.android)
+    // which isn't a real domain, so we use it verbatim instead of running it through PSL.
+    let referrer_name = if referrer_url.scheme() == "android-app" {
+        referrer_url.host_str().map(|h| h.to_string())
+    } else {
+        referrer_url.host_str().and_then(|h| extract_root_domain(h))
+    };
```
flopper123: same applies here:

```suggestion
    let referrer_name = if matches!(referrer_url.scheme(), "http" | "https") {
        referrer_url.host_str().and_then(|h| extract_root_domain(h))
    } else {
        referrer_url.host_str().map(|h| h.to_string())
    };
```

maybe even just do the `let is_web_scheme = matches!(referrer_url.scheme(), "http" | "https")` once further up and reuse

dashboard/src/lib/filter-sql.ts
```diff
@@ -1,9 +1,13 @@
 import { z } from 'zod';
 import { FILTER_COLUMNS, type TableFilterColumn } from '@/entities/analytics/filter.entities';
-import { SQL } from './safe-sql';
+import { SQL, safeSql } from './safe-sql';
 
 const TABLE_COLUMN_SCHEMA = z.enum(FILTER_COLUMNS);
 
+const FILTER_COLUMN_SQL_OVERRIDES: Partial<Record<TableFilterColumn, string>> = {
+  referrer_source: 'referrer_source_effective',
+};
```
flopper123: Quibble: Already return a `safeSql` here. That way we remove a branch from the `SQL.Unsafe` below.

```suggestion
const FILTER_COLUMN_SQL_OVERRIDES: Partial<Record<TableFilterColumn, ReturnType<typeof safeSql>>> = {
  referrer_source: safeSql`referrer_source_effective`,
};
```

backend/src/db/models.rs
```diff
@@ -28,6 +28,7 @@ pub struct EventRow {
     pub referrer_source_name: String,
     pub referrer_search_term: String,
     pub referrer_url: String,
+    pub referrer_source_canonical: String,
```
flopper123: Quibble: Move `referrer_source_canonical` up to `referrer_source`
```rust
    pub referrer_source: String,
    pub referrer_source_canonical: String,
    pub referrer_source_name: String,
    pub referrer_search_term: String,
    pub referrer_url: String,
```

migrations/34_referrer_canonical_and_classification.sql
```diff
@@ -0,0 +1,88 @@
...
+    any(os) AS os,
+    any(os_version) AS os_version,
+    any(country_code) AS country_code,
+    any(subdivision_code) AS subdivision_code,
+    any(city) AS city,
+    min(tuple(timestamp, referrer_source)) AS referrer_source,
+    min(tuple(timestamp, referrer_source_name)) AS referrer_source_name,
+    min(tuple(timestamp, referrer_search_term)) AS referrer_search_term,
+    min(tuple(timestamp, referrer_url)) AS referrer_url,
+    min(tuple(timestamp, referrer_source_canonical)) AS referrer_source_canonical,
```
flopper123: Same as other comment 😄

migrations/34_referrer_canonical_and_classification.sql
```diff
@@ -0,0 +1,88 @@
...
+FROM analytics.events
+GROUP BY site_id, session_created_at, session_id;
+
+CREATE TABLE IF NOT EXISTS analytics.referrer_source_categories (
+    generation UInt64,
+    key String,
+    medium LowCardinality(String)
+) ENGINE = ReplacingMergeTree(generation)
+ORDER BY key
+SETTINGS min_rows_for_wide_part = 0, min_bytes_for_wide_part = 0;
```
flopper123: Question: Just out of curiosity, any specific reason for this? 😄

.prettierignore
```diff
@@ -1,4 +1,5 @@
 node_modules
 .next
 docs/src/content/*
-static/replay.js
\ No newline at end of file
+static/replay.js
+/backend/assets/*
```
flopper123: Nitpick: Add a newline to end of file 😅

flopper123 (changes requested):
Code review

migrations/34_referrer_canonical_and_classification.sql
```diff
@@ -0,0 +1,88 @@
...
+GROUP BY key;
+
+CREATE DICTIONARY IF NOT EXISTS analytics.referrer_source_categories_dict (
+    key String,
+    medium String DEFAULT ''
+)
+PRIMARY KEY key
+SOURCE(CLICKHOUSE(NAME referrer_dict_source TABLE 'referrer_source_categories_current' DB 'analytics'))
+LAYOUT(HASHED())
+LIFETIME(MIN 3600 MAX 7200)
```
flopper123: Missing semicolon causing migration to fail:
```suggestion
LIFETIME(MIN 3600 MAX 7200);
```

clickhouse-config/setup/init-users.sh
```diff
@@ -12,8 +12,9 @@ clickhouse-client --query="CREATE ROLE dashboard_role;"
 clickhouse-client --query="CREATE ROLE backend_role;"
 
 # Grant privileges to roles
-clickhouse-client --query="GRANT SELECT ON analytics.*, INSERT ON analytics.*, ALTER ON analytics.*, CREATE VIEW ON analytics.*, SELECT ON system.databases, SELECT ON system.tables TO backend_role;"
-clickhouse-client --query="GRANT SELECT ON analytics.* TO dashboard_role;"
+clickhouse-client --query="GRANT SELECT ON analytics.*, INSERT ON analytics.*, ALTER ON analytics.*, CREATE VIEW ON analytics.*, dictGet ON analytics.*, SELECT ON system.databases, SELECT ON system.tables, SELECT ON system.dictionaries TO backend_role;"
+clickhouse-client --query="GRANT SYSTEM RELOAD DICTIONARY ON *.* TO backend_role;"
+clickhouse-client --query="GRANT SELECT ON analytics.*, dictGet ON analytics.* TO dashboard_role;"
```
flopper123, line 17: Turns out that this code does not rerun if the database already exists, so we'll need to update the migration as well for it to work on dev/prod.

migrations/34_referrer_canonical_and_classification.sql
```diff
@@ -0,0 +1,88 @@
...
+        );
+
+ALTER TABLE analytics.sessions
+    ADD COLUMN IF NOT EXISTS referrer_source_effective String ALIAS
+        coalesce(
+            nullIf(dictGetOrDefault('analytics.referrer_source_categories_dict', 'medium', tupleElement(referrer_source_canonical, 2), ''), ''),
+            nullIf(dictGetOrDefault('analytics.referrer_source_categories_dict', 'medium', domainWithoutWWW(concat('http://', tupleElement(referrer_url, 2))), ''), ''),
+            nullIf(dictGetOrDefault('analytics.referrer_source_categories_dict', 'medium', tupleElement(referrer_source_name, 2), ''), ''),
+            tupleElement(referrer_source, 2)
+        );
```
flopper123, line 87: This should contain:

```sql
GRANT dictGet ON analytics.* TO backend_role;

GRANT SELECT ON system.dictionaries TO backend_role;

GRANT SYSTEM RELOAD DICTIONARY ON *.* TO backend_role;

GRANT dictGet ON analytics.* TO dashboard_role;

```

backend/src/main.rs
Lindharden: I don't think main should be responsible for handling this. With the way it is now, it's just a one-shot startup sync 🤔 

What's your motivation for this? 😄

migrations/34_referrer_canonical_and_classification.sql
Lindharden: Should not be in the migration. It's in the post migration script and init-users.sh 😄

backend/src/main.rs
flopper123: Currently this pushes the responsibility of building the referrer source categories to the `db` module instead though.
By extracting the setup, the `db` only needs to know about the `Vec<ReferrerSourceCategory>`, instead of knowing about the local file-paths / order of the local files.
If we want to later add an additional referrer config, both `main`, `db`, and `referrer` module needs to be updated. In my mind `db` shouldn't need to know about the local file path of the referrers.

Not a blocker though 😄

flopper123 (approved):

PR #902, opened 2026-05-09
Add outbound link filter

### LINKED ISSUE #883: Outbound link as funnel step
Feature request: It would be great if a funnel step could be an Outbound Link.

Woldbye:
Adds `outbound_link_url` as a filter column (entities, MCP registry, session-query main-table list, locale strings, docs) and seeds outbound link events from simulate-events.js for funnel testing.

## Known limitation

On pages with pageview-gated widgets (Pages, Web Vitals, Errors, Overview's pageview chart), filtering by `outbound_link_url` produces empty widgets.  Note that the same issue already exists for `custom_event_name` on main.

I will add a Follow-up PR that hides the column on incompatible pages.

chrome_3OA3JVR70w


Lindharden (approved):
Awesome 👍

PR #903, opened 2026-05-09
Added usage alert emails

Lindharden:
Adds two distinct email cases:

- Usage nearing limits (sent at 80+ %)
- Usage exceeded (sent at 100+ %)

### DIFF
--- a/dashboard/src/app/actions/system/email.action.ts
+++ b/dashboard/src/app/actions/system/email.action.ts
@@ -11,6 +11,8 @@ import { getFirstPaymentWelcomeEmailPreview } from '@/services/email/template/fi
 import { getEmailVerificationPreview } from '@/services/email/template/email-verification-mail';
 import { getInvitationEmailPreview } from '@/services/email/template/invitation-mail';
 import { getReportEmailPreview } from '@/services/email/template/weekly-report-mail';
+import { startOfMonth } from 'date-fns/startOfMonth';
+import { endOfMonth } from 'date-fns/endOfMonth';
 
 async function requireEmailPreviewAccess(): Promise<string> {
   if (!isFeatureEnabled('enableEmailPreview')) {
@@ -74,8 +76,8 @@ export async function sendTestEmail(email: string, template: EmailType) {
           usageLimit: 10000,
           usagePercentage: 95,
           planName: 'Starter',
-          currentPeriodStart: new Date('2026-04-01T00:00:00.000Z'),
-          currentPeriodEnd: new Date('2026-04-30T00:00:00.000Z'),
+          currentPeriodStart: startOfMonth(new Date()),
+          currentPeriodEnd: endOfMonth(new Date()),
           upgradeUrl: 'https://betterlytics.io/billing',
         }, userId));
         break;
--- a/dashboard/src/app/actions/system/email.action.ts
+++ b/dashboard/src/app/actions/system/email.action.ts
@@ -74,6 +74,8 @@ export async function sendTestEmail(email: string, template: EmailType) {
           usageLimit: 10000,
           usagePercentage: 95,
           planName: 'Starter',
+          currentPeriodStart: new Date('2026-04-01T00:00:00.000Z'),
+          currentPeriodEnd: new Date('2026-04-30T00:00:00.000Z'),
           upgradeUrl: 'https://betterlytics.io/billing',
         }, userId));
         break;
--- a/dashboard/src/repositories/clickhouse/usage.repository.ts
+++ b/dashboard/src/repositories/clickhouse/usage.repository.ts
@@ -70,7 +70,7 @@ export async function getDailyEventCountsForSites(
       })
       .toPromise();
 
-    return (result as unknown[]).map((row) => DailySiteUsageSchema.parse(row));
+    return DailySiteUsageSchema.array().parse(result);
   } catch (error) {
     console.error('Failed to get daily event counts for sites:', error);
     throw error;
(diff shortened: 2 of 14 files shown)


dashboard/src/repositories/postgres/subscription.repository.ts
```diff
@@ -2,12 +2,96 @@ import prisma from '@/lib/postgres';
...
+        eventLimit: s.eventLimit,
+        currentPeriodStart: s.currentPeriodStart,
+        currentPeriodEnd: s.currentPeriodEnd,
+        paymentSubscriptionId: s.paymentSubscriptionId,
+        siteIds: s.user.dashboardAccess.map((ud) => ud.dashboard.siteId),
+      }),
+    );
+}
+
+export async function rollForwardStaleFreeTierPeriods(
```
Woldbye: I am worried this will cause latency on unrelated endpoints during scan ticks, due to the unbounded parallelization. `Promise.all` dispatches EVERY callback synchronously => N expired free-tier rows, => N in-flight UPDATE's simultaneously competing with our app for the bounded Prisma pool.

Best practise is usually "slow but never disruptive" for a cron, so how about removing the Promise.all, and just running the roll-forwards serially?

dashboard/src/worker/jobs/billing/usage-threshold-scan.job.ts
```diff
@@ -0,0 +1,110 @@
...
+  if (pct >= 80) return 'approaching';
+  return null;
+}
+
+function formatPlanName(tier: SubscriptionForUsageScan['tier']): string {
+  return tier.charAt(0).toUpperCase() + tier.slice(1);
+}
+
+function indexBySite(rows: DailySiteUsage[]): Map<string, DailySiteUsage[]> {
+  const out = new Map<string, DailySiteUsage[]>();
```
Woldbye: Quibble: consider adding a code comment to replace this entire function with `Map.groupBy(rows, (row) => row.siteId)` once we migrate to Node 21+ :D

dashboard/src/app/actions/system/email.action.ts
```diff
@@ -74,6 +74,8 @@ export async function sendTestEmail(email: string, template: EmailType) {
           usageLimit: 10000,
           usagePercentage: 95,
           planName: 'Starter',
+          currentPeriodStart: new Date('2026-04-01T00:00:00.000Z'),
```
Woldbye: Quibble: For a user-preview test email I would suggest using `date-fns` two utility functions to keep the date time agnostic.

```
import { startOfMonth, endOfMonth } from 'date-fns';
// …
  currentPeriodStart: startOfMonth(new Date()),
  currentPeriodEnd: endOfMonth(new Date()),
``` 

Thereby, we dont have to update the code to prevent the test-email's display of 2026 dates once we reach 2027 🗡️

Woldbye (commented):
Very nice! I had a few small code comment that you might wanna look into, but overall it looks like high code quality and nice changes.

**DISCLAIMER:** I have not tested the visuals of the actual email-templates..

dashboard/src/repositories/clickhouse/usage.repository.ts
```diff
@@ -31,11 +39,44 @@ async function _getSiteEventCountForRange(siteId: string, startDate: DateTimeStr
...
+  `;
+
+  try {
+    const result = await clickhouse
+      .query(query.taggedSql, {
+        params: { ...query.taggedParams, earliest_start: earliestStart },
+      })
+      .toPromise();
+
+    return (result as unknown[]).map((row) => DailySiteUsageSchema.parse(row));
```
flopper123: Quibble: Can be simplified (same for the other repositories). Know we haven't been super consistent here 😄 
```suggestion
    return DailySiteUsageSchema.array().parse(result);
```

dashboard/src/worker/jobs/billing/usage-threshold-scan.job.ts
```diff
@@ -0,0 +1,110 @@
...
+}
+
+function sumUsageForUser(sub: SubscriptionForUsageScan, bySite: Map<string, DailySiteUsage[]>): number {
+  const periodStart = toDateString(sub.currentPeriodStart);
+  let total = 0;
+  for (const siteId of sub.siteIds) {
+    const rows = bySite.get(siteId);
+    if (!rows) continue;
+    for (const row of rows) {
+      if (row.date >= periodStart) total += row.total;
```
flopper123: This does string comparison for comparing dates. They should probably both be converted and matched properly 😄

dashboard/src/worker/jobs/billing/usage-threshold-scan.job.ts
```diff
@@ -0,0 +1,110 @@
...
+  const subs = await rollForwardStaleFreeTierPeriods(rawSubs);
+
+  const allSiteIds = [...new Set(subs.flatMap((s) => s.siteIds))];
+  const earliestStart = subs.reduce(
+    (min, s) => (s.currentPeriodStart < min ? s.currentPeriodStart : min),
+    subs[0].currentPeriodStart,
+  );
+
+  const dailyRows = await getDailyEventCountsForSites(allSiteIds, toDateString(earliestStart));
+  const bySite = indexBySite(dailyRows);
```
flopper123: Question: Why do we do the absolute min period start? This will change from user to user, no? So this should be done per subscription, and not the min for all subscriptions, no? 🤔

flopper123 (changes requested):

dashboard/src/worker/jobs/billing/usage-threshold-scan.job.ts
Lindharden: It bulk fetches from ClickHouse for multiple dashboards, so we find the earliestStart subscription in the batch being processed, fetch all daily events for all sites starting from that date, then index them in-memory to their own respective boundary. This is to prevent querying ClickHouse for each dashboard 😄

Woldbye (approved):

dashboard/src/worker/jobs/billing/usage-threshold-scan.job.ts
Lindharden: I'll skip this to not risk having a floating comment 😄

flopper123 (approved):

PR #905, opened 2026-05-10
Fix nested links in DashboardCard for better accessibility

Fi1osof:
- Replace nested Link components with absolute positioning pattern
- Enable opening dashboard in new tab or middle-click
- Add proper z-index layering for settings link clickability

### DIFF
--- a/dashboard/src/app/(protected)/dashboards/DashboardCard.tsx
+++ b/dashboard/src/app/(protected)/dashboards/DashboardCard.tsx
@@ -23,17 +23,16 @@ export default function DashboardCard({ dashboard }: DashboardCardProps) {
     options: { year: 'numeric', month: 'short', day: 'numeric' },
   });
 
-  const handleCardClick = () => {
-    router.push(`/dashboard/${dashboard.id}`);
-  };
-
   const hasMultipleMembers = dashboard.memberCount > 1;
 
   return (
-    <Card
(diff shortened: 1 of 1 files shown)


Woldbye (approved):
Thanks a lot for the contribution! 👍 Approving and merging!

I think the overlay link pattern is a clean fix. It solves two accesability problems - tab target, and correct link semantics.

While it makes the card tab-targetable sadly the existing hover: / group-hover: rules don't fire on focus, so tabbing doesn't visually lift the card, and the ring-border of the overlay is slightly misaligned with the container.

 I'll look into the right approach as a follow-up.

Fi1osof:
> Thanks a lot for the contribution! 👍 Approving and merging!
> 
> I think the overlay link pattern is a clean fix. It solves two accesability problems - tab target, and correct link semantics.
> 
> While it makes the card tab-targetable sadly the existing hover: / group-hover: rules don't fire on focus, so tabbing doesn't visually lift the card, and the ring-border of the overlay is slightly misaligned with the container.
> 
> I'll look into the right approach as a follow-up.

If you want to support both keyboard tab navigation and proper focus states, then the container probably needs to become a regular div, with the settings link moved outside of the main overlay link so they become sibling links instead. Otherwise it ends up being a link inside another link, which is invalid HTML.

PR #906, opened 2026-05-10
Add localized title, and ring fix on dashboard-card

Woldbye:
Follow-up to #905

  - Localize the dashboard card's overlay `aria-label`/`title` and the  settings link's `title`. Translations added for en/da/it/nb under  `dashboardsPage.goToDashboardTooltip` and `dashboardsPage.settingsTooltip`.
  - Wire keyboard focus to mirror the hover affordances on the card via `focus-within:` / `group-focus-within:`, so tabbing visually lifts the card the way `CreateDashboardCard` already does.
  - Paint the focus ring on the Card itself via `has-[>a:focus-visible]:ring-1 has-[>a:focus-visible]:ring-ring/50` with `outline-none` on the overlay, so the ring traces the Card's rounded border
  - Drop dead leftovers from #905: unused `useBARouter`/`router` and a `stopPropagation` that no longer has anything to stop.


Lindharden (approved):
Nice

PR #907, opened 2026-05-10
Inject providers into login form to avoid layout shift

Woldbye:
Fixes a bug, on initial visit to https://betterlytics.io/signin. Initially the OAuth provider buttons are hidden. The next frame, they populate, causing the layout to shift.

To resolve this, I have injected the providers from the server-side component, so they start populated.


dashboard/src/components/auth/LoginForm.tsx
```diff
@@ -1,7 +1,7 @@
 'use client';
 
 import { useCallback, useEffect, useRef, useState, useTransition } from 'react';
-import { signIn } from 'next-auth/react';
+import { signIn, getProviders } from 'next-auth/react';
```
Lindharden: Import ClientSafeProvider instead similarly to how it's done in auth.ts. This way we avoid pulling a function symbol into the client bundle just for its type 😄 

Note, if the comment further down is done instead, this becomes:
`  providers: ReturnType<typeof getEnabledOAuthProviders>;`

dashboard/src/components/auth/LoginForm.tsx
```diff
@@ -27,9 +26,14 @@ import { Label } from '@/components/ui/label';
 type LoginFormProps = {
   registrationDisabledMessage?: string | null;
   forgotPasswordEnabled?: boolean;
+  providers: Awaited<ReturnType<typeof getProviders>>;
```
Lindharden: Tighten and drop the ?. further down

```suggestion
  providers: Record<string, ClientSafeProvider>;
```

Bonus if you do this in SignupForm.tsx as well 😄

dashboard/src/lib/auth.ts
```diff
@@ -164,6 +165,24 @@ function isUserException(error: unknown): error is UserException {
...
+          id: provider.id,
+          name: provider.name,
+          type: provider.type,
+          signinUrl: `${baseUrl}/api/auth/signin/${provider.id}`,
+          callbackUrl: `${baseUrl}/api/auth/callback/${provider.id}`,
+        },
+      ];
+    }),
+  );
+}
```
Lindharden: This carries a lot of dead weight because the signin/signup functions uses built-in nextauth regardless. We can simplify significantly to:

```suggestion
export function getEnabledOAuthProviders(): { google: boolean; github: boolean } {
  return {
    google: Boolean(env.GOOGLE_CLIENT_ID && env.GOOGLE_CLIENT_SECRET),
    github: Boolean(env.GITHUB_ID && env.GITHUB_SECRET),
  };
}
```

Lindharden (changes requested):
Good catch and nice to have the layout shift fixed 😄

dashboard/src/components/auth/LoginForm.tsx
Lindharden: Fixed in https://github.com/betterlytics/betterlytics/pull/907/commits/55521088417e05ed229544dcba41a565bceffdf2

dashboard/src/components/auth/LoginForm.tsx
Lindharden: Fixed in https://github.com/betterlytics/betterlytics/pull/907/commits/6b06bc7dd7da483bcbb3cdb8bdf239c8c3c50600

dashboard/src/lib/auth.ts
Lindharden: Fixed in https://github.com/betterlytics/betterlytics/pull/907/changes/55521088417e05ed229544dcba41a565bceffdf2

Lindharden (approved):
lgtm

PR #916, opened 2026-05-10
Add accessibility to active query filter close icon

Woldbye:
Replaces the close icon with a proper Button so it's keyboard-focusable and screen-reader labelled, and bumps its target to 24×24 px (WCAG AA min). I have also added a new `removeFilter` translation key across our localized languages.

## Demo
image


Lindharden (approved):
LGTM

Issue #843, opened 2026-04-02
Cancelling UserSettings reverts the theme to default


Woldbye:
### How to reproduce:
1. Login in to the dashboard and set the theme to the opposite of your default settings (I have defualt dark mode, so for me that would be changing it to Light mode), and save-settings.
2. Reopen the `user-settings` modal and CANCEL it => It will revert the settings to default, leaving the applied theme and the UI theme out of sync (the value in the modal is still Light mode, but the actual theme applied to the page is revert to the default Dark mode).

#### Triggering the bug
![Image](https://github.com/user-attachments/assets/998e5ca5-5a40-4de9-9744-a28ee2c914c1)
#### After triggering the bug
<img width="2290" height="1700" alt="Image" src="https://github.com/user-attachments/assets/38c54fa4-5ef2-4867-ad5f-f1b9eeb43f51" />

PR #919, opened 2026-05-12
Chore/nextjs 16 upgrade

flopper123:
Migration to Next 16:
- In AppRouter `generateMetadata` props have changed:
```tsx
// Before
function generateMetadata({ params }: { params: Promise<{ locale: SupportedLanguages }> }) {

// After
function generateMetadata({ params }: PageProps<'/[locale]/terms'>) {
```
- Routes are now generated for type definition
- Sidebar changes due hydration issues
- Most changes are from `pnpm-lock.yaml`
- You may get a `[DEP0108] DeprecationWarning: zlib.bytesRead`, when starting the dev server.
- since [next-themes](https://www.npmjs.com/package/next-themes) is no longer maintained and produces errors with the latest React/Next version, I've changed it to [@wrksz/themes](https://www.npmjs.com/package/@wrksz/themes)
- ThemeProvider now needs to be called from a server component for full SSR support, why is has been moved


dashboard/src/app/(protected)/UserThemeInitializer.tsx
```diff
@@ -13,7 +13,7 @@ export default function UserThemeInitializer({ theme: userTheme }: UserThemeInit
 
   useEffect(() => {
     if (userTheme) {
-      setTheme(userTheme);
+      setTheme(userTheme as 'light' | 'dark' | 'system');
```
Woldbye: Would probably be cleaner to use the ThemeValue type you have declared in this pr

dashboard/src/app/layout.tsx
```diff
@@ -44,7 +44,7 @@ export default async function RootLayout({
   const [locale, seoConfig] = await Promise.all([getLocale(), buildSEOConfig(SEO_CONFIGS.root)]);
 
   return (
-    <html lang={locale} suppressHydrationWarning>
+    <html lang={locale} suppressHydrationWarning data-scroll-behavior='smooth'>
```
Woldbye, line 47: What's the purpose of  hoisting scroll behavior on the html? What consumer reads this ?

dashboard/src/app/layout.tsx
```diff
@@ -63,11 +63,14 @@ export default async function RootLayout({
       </head>
       <body className={`${robotoSans.variable} ${robotoMono.variable} antialiased`}>
         <NextTopLoader color='var(--primary)' height={3} showSpinner={false} shadow={false} />
-        <ThemeColorUpdater />
         <NextIntlClientProvider>
-          <Providers>{children}</Providers>
+          <ThemeProvider attribute='class' defaultTheme='system' enableSystem disableTransitionOnChange>
+            <Providers>
+              {children}
+              <Toaster />
```
Woldbye, line 70: Why did you move the Toaster into the context? The toasts teleport to the toaster root, so it might have some unexpected sideeffects

dashboard/src/components/userSettings/UserThemeSelector.tsx
```diff
@@ -23,7 +23,7 @@ export default function UserThemeSelector({ value, onUpdate }: UserThemeSelector
   const t = useTranslations('components.userSettings.preferences.appearance');
 
   const handleChange = (newTheme: string) => {
-    setTheme(newTheme);
+    setTheme(newTheme as 'light' | 'dark' | 'system');
```
Woldbye: Quibble: use a named type for theme

Woldbye (commented):
Good idea to update!!

PR mostly looks good, but I left a few code comments.

Two additonal question:

1. I don't understand the sidebar changes - what's the hydration issue exactly and what have you done to fix it?

2. We should test if this resolves the known theme bug we have an open issue on:  https://github.com/betterlytics/betterlytics/issues/843

dashboard/src/app/layout.tsx
flopper123: Because `Toaster` uses the `useTheme` hook, which now requires it to be wrapped in the context provider 😄

dashboard/src/app/layout.tsx
flopper123: Next itself uses this now 😄
Next 16 changed default scroll behavior during App Router navigation. Since we've specified `scroll-behavior: smooth;` in out `globals.css` it doesn't auto-detect this anymore, and does a fallback during navigations, which feels janky. So this keeps our previous behavior 🫡

dashboard/src/components/ui/FilterPreservingLink.tsx
```diff
@@ -22,13 +22,19 @@ interface FilterPreservingLinkProps extends Omit<LinkProps, 'href'> {
...
+    let getHrefWithFilters: (h: string) => string;
+    let navigation: ReturnType<typeof useOptionalDashboardNavigation>;
+    let pathname: string;
+    try {
+      getHrefWithFilters = useNavigateWithFilters().getHrefWithFilters;
+      navigation = useOptionalDashboardNavigation();
+      pathname = usePathname();
+    } catch (e) {
+      throw e;
+    }
```
Lindharden, line 34: This seems odd.

Should probably just be:

```suggestion
    const { getHrefWithFilters } = useNavigateWithFilters();
    const navigation = useOptionalDashboardNavigation();
    const pathname = usePathname();
```

dashboard/src/app/layout.tsx
```diff
@@ -3,12 +3,12 @@ import Script from 'next/script';
 import './globals.css';
 import { env } from '@/lib/env';
 import Providers from '@/app/Providers';
+import { ThemeProvider } from '@/app/ThemeProvider';
 import { Toaster } from '@/components/ui/sonner';
 import { StructuredData } from '@/components/StructuredData';
 import NextTopLoader from 'nextjs-toploader';
 import { getLocale } from 'next-intl/server';
 import { NextIntlClientProvider } from 'next-intl';
-import ThemeColorUpdater from '@/app/ThemeColorUpdater';
```
Lindharden, line 11: ThemeColorUpdater is now orphaned and should likely be removed since it's unused.

dashboard/src/app/layout.tsx
```diff
@@ -63,11 +63,14 @@ export default async function RootLayout({
       </head>
       <body className={`${robotoSans.variable} ${robotoMono.variable} antialiased`}>
         <NextTopLoader color='var(--primary)' height={3} showSpinner={false} shadow={false} />
-        <ThemeColorUpdater />
         <NextIntlClientProvider>
-          <Providers>{children}</Providers>
+          <ThemeProvider attribute='class' defaultTheme='system' enableSystem disableTransitionOnChange>
```
Lindharden, line 67: Removing the ThemeColorUpdater breaks the background coloring on mobile view. This should fix it:

```suggestion
          <ThemeProvider
            attribute='class'
            defaultTheme='system'
            enableSystem
            disableTransitionOnChange
            themeColor={{ light: '#f9f9f9', dark: '#101111' }}
          >
```

Lindharden (changes requested):
Perhaps due to the SupportedLanguages prop typeloss, we can compensate with a single Zod (or equivalent) parse in the [locale] layout so descendants still get a narrowed SupportedLanguages?

With that said, I think we should merge this to a separate branch to prevent merging an unfinished upgrade to main. All remaining upgrade PRs can then point to this branch and still be reviewable in chunks. wdyt?

PR #920, opened 2026-05-12
Unify funnel CRUD behavior

Woldbye:
Thread dashboardId through getFunnelById and updateFunnel to match the existing deleteFunnelById signature, so all three CRUD operations scope by dashboard consistently.


flopper123 (approved):
Super nice! 😄

PR #921, opened 2026-05-14
Migrate emails to react-email + add UTM tracking

Lindharden:
Added react-email and deleted the in-app email-preview page and its server action.

Migrated all 6 transactional templates (reset-password, email-verification, dashboard-invitation, usage-alert, first-payment-welcome, weekly-report) from string-builder HTML to react-email + Tailwind. Plaintext versions are now auto-derived; the old generate*EmailText helpers are gone. Shared layout/header/footer/button/info-box/typography live in _components.tsx; weekly-report uses a responsive stacked layout on mobile.

Added UTM tagging to every outbound betterlytics.io link in our emails (utm_source=betterlytics, utm_medium=email, utm_campaign=<email_type>, optional utm_content). Backend already parses these; nothing else needed.

Added pnpm emails (dev preview server with hot reload, replaces the in-app preview).

Removed: email-preview page + components, sendTestEmail/getEmailPreview actions, ENABLE_MAIL_PREVIEW_PAGE feature flag, escape-html dep, dead barrel exports.


dashboard/src/services/email/template/_components.tsx
```diff
@@ -0,0 +1,265 @@
...
+export function H2({ children, className = '' }: { children: ReactNode; className?: string }) {
+  return (
+    <Heading as="h2" className={`mt-8 mb-4 text-[22px] font-semibold text-slate-700 ${className}`.trim()}>
+      {children}
+    </Heading>
+  );
+}
+
+export function P({ children, className = '' }: { children: ReactNode; className?: string }) {
+  return <Text className={`my-4 text-base leading-relaxed text-slate-600 ${className}`.trim()}>{children}</Text>;
```
flopper123: Quibble: Use `cn(...)` for the classnames in this file 😄

dashboard/src/services/email/template/invitation-mail.tsx
```diff
@@ -0,0 +1,76 @@
...
+};
+
+export function DashboardInvitationEmail({
+  inviterName,
+  dashboardName,
+  role,
+  inviteToken,
+  userExists,
+}: DashboardInvitationEmailData) {
+  const acceptUrl = `${sharedEmailEnv.publicBaseUrl}/en/accept-invite/${inviteToken}`;
```
flopper123: Question: Is there a reason to 'force' the English version, instead of just pointing to `/accept-invite/${inviteToken}`?

dashboard/src/services/email/template/invitation-mail.tsx
```diff
@@ -0,0 +1,76 @@
...
+        Betterlytics.
+      </P>
+
+      <P>
+        You've been invited as a <strong>{roleDisplay}</strong>.
+      </P>
+
+      <EmailButton href={withEmailUtm(acceptUrl, CAMPAIGN, 'primary_cta')}>{buttonText}</EmailButton>
+
+      <P className="text-sm text-slate-500">This invitation will expire in 7 days.</P>
```
flopper123: Quibble: It seems as if this file is missing prettier formatting. Some strings are `"` while other are `'`

dashboard/src/services/email/template/reset-password-mail.tsx
```diff
@@ -0,0 +1,84 @@
...
+  expirationTime: string;
+}
+
+export function ResetPasswordEmail({ userName, resetUrl, expirationTime }: ResetPasswordEmailData) {
+  return (
+    <EmailLayout preview="Reset your Betterlytics password" campaign={CAMPAIGN}>
+      <H1>Reset Your Password</H1>
+
+      <P>
+        Hi <strong>{userName}</strong>,
```
flopper123, line 31: Question: I know that `userName` is typed as `string`, however, do we still need to escape text, or are we now safe by default? 😄

dashboard/src/services/email/template/usage-alert-mail.tsx
```diff
@@ -0,0 +1,175 @@
...
+  return 'warning';
+}
+
+function formatBillingDate(value: SerializableDate): string {
+  return format(new Date(value), 'MMM d, yyyy');
+}
+
+function getSubjectLine(percentage: number, planName: string): string {
+  if (percentage >= 100) return `Usage alert: ${planName} plan event limit exceeded`;
+  return `Usage alert: ${percentage}% of your ${planName} plan event limit used`;
```
flopper123: Question: I can see that in `dashboard/src/worker/jobs/billing/usage-threshold-scan.job.ts` we already do format percentage, however, should we maybe also add a format wrapper for percentage here, incase we change the implementation? Just so that we don't end up with emails down the line with random decimal points? 😄

dashboard/src/services/email/template/usage-alert-mail.tsx
```diff
@@ -0,0 +1,175 @@
...
+    ctaLabel: 'Review billing options',
+  };
+}
+
+export function UsageAlertEmail(data: UsageAlertEmailData) {
+  const severity = getSeverity(data.usagePercentage);
+  const statusCopy = getStatusCopy(severity);
+  const periodLabel = `${formatBillingDate(data.currentPeriodStart)} to ${formatBillingDate(
+    data.currentPeriodEnd,
+  )}`;
```
flopper123: Suggestion: Extract the formatting so that the `periodLabel` doesn't get random newlines, since:
```js
const a = `ab
c`;
console.log(a); // Prints 'ab\nc'
```
```suggestion
  const periodStart = formatBillingDate(data.currentPeriodStart);
  const periodEnd = formatBillingDate(data.currentPeriodEnd);
  const periodLabel = `${periodStart} to ${periodEnd}`;
```

dashboard/src/services/email/template/weekly-report-mail.tsx
```diff
@@ -0,0 +1,265 @@
...
+      ...data.reportData,
+      period: { start: new Date(period.start), end: new Date(period.end) },
+      comparisonPeriod: { start: new Date(comparisonPeriod.start), end: new Date(comparisonPeriod.end) },
+    },
+  };
+}
+
+function truncate(s: string, max = 40): string {
+  return s.length > max ? s.slice(0, max - 1) + '…' : s;
+}
```
flopper123: Suggestion: Consider using the existing `formatString` instead for mid-ellipsis since it's formatting paths

dashboard/src/services/email/template/_components.tsx
```diff
@@ -0,0 +1,265 @@
...
+  return { subject, html, text };
+}
+
+/**
+ * Append UTM parameters to an outbound email link so clicks are attributable
+ * to the email in our analytics. mailto: links and unparseable URLs pass through.
+ */
+export function withEmailUtm(url: string, campaign: string, content?: string): string {
+  if (url.startsWith('mailto:')) return url;
+  try {
```
flopper123, line 41: Question: Should we maybe also check if the `url` is for Betterlytics? Because for self-hosters for instance the url could point to their own domain for instance, no? 🤔

flopper123 (changes requested):

dashboard/src/services/email/template/reset-password-mail.tsx
Lindharden: We should now be safe because React JSX auto-escapes text and react-email's render() runs through React's HTML pipeline

dashboard/src/services/email/template/usage-alert-mail.tsx
Lindharden: Is false, the code has newlines inside the function-call, not inside the template literal text - I have made the change for readability regardless 😄

dashboard/src/services/email/template/_components.tsx
Lindharden: No, but we should disable it for self-hosters tho. Selfhosters are very likely not interested in knowing if they themselves click on their own emails 😆

dashboard/src/services/email/template/invitation-mail.tsx
Lindharden: No, the /en/ was already there before this PR; this PR only ported it. I'll remove it tho 👍

flopper123 (dismissed):
Verrry nice 🔥

flopper123 (dismissed):

flopper123 (approved):

PR #923, opened 2026-05-14
Only shows delete button on saved filters when hovering a row

Lindharden:


Woldbye (approved):
Nice idea. It works well!!

flopper123 (approved):

PR #924, opened 2026-05-15
Redesign funnel dialog's to support multi-filter-funnel's per step

### LINKED ISSUE #852: Allow multi filters per funnel step
Currently, funnels only allow one filter per step, instead of allowing for a granular segmentation:

Image

This can be too limiting in cases, such as:
- Wanting to filter on more than one dimension per step.
- Wildcard searching too simple for more complex cases

We could reuse the setup used for regular dashboard filters, where each filters is an `AND` and each value within are `OR`:
Image

Woldbye:
Redesign our funnel dialog to support Multiple filters per funnel step. To support the UX I have used an accordion component for the the funnel steps in a side-by-side view with the preview. As part of the refactor, I have made the view responsive. This means the UI now works properly for mobile (instead of hiding the edit options for mobile like main do). Further, I have enabled full accessability (minus readaloud options for the DND).

**Additional Notes**:
* The changes to the `FunnelBarPlot` are primarily to 
    * ensure the barplot has the same height across loading, empty and filled state to avoid layout shifts
    * make the barplot have a more flexible height, to ensure it fits in the side-by-side view.
* There is a slight alignment issue on right-hand side for mobile (notice how its slightly left-shifted on mobile).
    * The cause of the issue, is that the row needs extra padding on right-side for the close-icon. Without the extra padding, it gets clipped by the container. It's very hard to solve this properly, so I suggest resolving it in a follow-up or just live with it.
 * I have changed the drag-and-drop component from `framer.motion.reorder` to [DND](https://www.npmjs.com/package/@dnd-kit/core), as framer's reorder doesn't work for the accordion items.

### Desktop
image

### Mobile
image


flopper123 (changes requested):
UI/UX feedback. Overall nice, but needs some refinement 😄 

1) When creating a funnel, it now allows empty filters for creation, but not for preview. But the created funnel will never work:

Image

And when the funnel has been created, it then just says something like `URL is`
Image

2) Phone layout has some weird issues when creating/updating funnels

Image

3) Instead of writing `AND` I'd suggest to just write `and`, or add another visual separator:

Image

4) You can now create funnels with no filters

Image

5) If you try to select the funnel step name by dragging the mouse over the text, it tries to re-order the funnel step. The reordering should probably be disabled on clickable/interactable elements

Image

6) Quibble: The blue highlight on the left side doesn't respect the border radius:

Image

7) Consider making the modal wider, keep the width of the filters, but make the preview wider:

Image

8) Dragging a step a little below the final row causes the entire funnel to bug

Image

9) Dragging a filter up (and also down sometimes) while it's open causes a weird visual bug where multiple filters jump around. If you look at the screenshot below, step number 4 is on top of step number 8 for some reason

Image

10) Moving your mouse over different areas of a step will show you many different cursor variation, even though most are for reordering.

--- Additional findings ---

11) On laptop screen sizes, the filter operator looks like it's always `is` , even if `is not` is selected for instance. The values are also overlapping.

image

Woldbye:
Hey :) I have went through all your feedback and resolved most:

> 1. When creating a funnel, it now allows empty filters for creation, but not for preview. But the created funnel will never work:
>
I haven't touched this, its the same behavior on main

> 2. Phone layout has some weird issues when creating/updating funnels
>
I believe I have resolved this on all screen-sizes: I had to create a new screen-size (windowed), to ensure it fits on screens closer to square (nest hub e.g.).

> 3. Instead of writing `AND` I'd suggest to just write `and`, or add another visual separator:
>
Done, changed to 'and'

> 4. You can now create funnels with no filters
>
I haven't touched this, its the same behavior on main.

> 5. If you try to select the funnel step name by dragging the mouse over the text, it tries to re-order the funnel step. The reordering should probably be disabled on clickable/interactable elements
>
On mobile, it has to be reorderable by holding mouse on the interactive input name field. On desktop, I have made lifted all input areas from the reorder 'grab'.

> 6. Quibble: The blue highlight on the left side doesn't respect the border radius:
>
Nice spotted. Resolved

> 7. Consider making the modal wider, keep the width of the filters, but make the preview wider:
>
I made the modal wider, but I also had to bump up the width of the filters slightly to make room

> 8. Dragging a step a little below the final row causes the entire funnel to bug
> 9. Dragging a filter up (and also down sometimes) while it's open causes a weird visual bug where multiple filters jump around. If you look at the screenshot below, step number 4 is on top of step number 8 for some reason
>
Resolved 8/9 by restricting the grabbed funnel step within the scrollable container thorugh a DND plugin.

> 10. Moving your mouse over different areas of a step will show you many different cursor variation, even though most are for reordering.
>
The cursor effects is actually as intended on all elements: It shows pointer click on the header, to visually illustrate that it expands/collapses on click. The grab handle only shows cursor-grab, since it's excluded from the expand/collapse.

> 11. On laptop screen sizes, the filter operator looks like it's always `is` , even if `is not` is selected for instance. The values are also overlapping.
>
It's resolved on most screen sizes, there's still a few edge screen sizes, where the top of the 'is not' gets cut off to 'is no'. Let me know if this is not good enough for you, and we will have to cook space further.

dashboard/prisma/schema.prisma
```diff
@@ -252,16 +252,28 @@ model FunnelStep {
...
+
+  column   String
+  operator String
+  values   String[]
+
+  funnelStep   FunnelStep @relation(fields: [funnelStepId], references: [id], onDelete: Cascade)
+  funnelStepId String
+
+  @@index([funnelStepId])
+}
```
Lindharden, line 275: Don't we need to have some sort of position to indicate in which order the filters are? Otherwise, we can't guarantee the order of steps? 🤔

dashboard/src/repositories/clickhouse/funnels.repository.ts
```diff
@@ -4,6 +4,34 @@ import { SQL, safeSql } from '@/lib/safe-sql';
 import { BAQuery } from '@/lib/ba-query';
 import { DateTimeString } from '@/types/dates';
 import { FunnelStep } from '@/entities/analytics/funnels.entities';
+import { QueryFilter } from '@/entities/analytics/filter.entities';
+
+const isUsableFilter = (f: QueryFilter) =>
+  Boolean(f.column) && Boolean(f.operator) && f.values.every(Boolean);
```
Lindharden: This is a duplicate of the isUsableFilter function defined in ba-query.ts

dashboard/messages/en.json
```diff
@@ -2391,6 +2391,7 @@
         "title": "Filters",
         "emptyNoneSelected": "No filters selected - apply to save",
         "addFilter": "Add filter",
+        "addFilterPlaceholder": "Add one or more filters",
```
Lindharden: Unused language key across all translations

dashboard/src/components/funnels/FunnelBarplot.tsx
```diff
@@ -9,71 +9,143 @@ import React, { useEffect, useRef, useState } from 'react';
...
+                  </p>
                 </div>
-              )}
+                <h4 className='text-foreground/70 truncate text-sm font-semibold'>{step.name}</h4>
+              </div>
+              <div
+                className={cn(
+                  'hidden w-full pt-2 sm:flex',
+                  fill ? 'sm:flex-1' : 'h-40',
+                  i === funnel.steps.length - 1 &&
```
Lindharden: This is inside the empty-steps loop, but the condition checks against the number of filled steps instead of the number of empty steps. The dashed divider ends up on the wrong placeholder, it drifts toward the middle (or disappears) depending on the filled/empty ratio rather than staying at the last column. Should perhaps reference the empty-steps length, or be dropped if the boundary line above already covers it?

dashboard/src/components/funnels/FunnelBarplot.tsx
```diff
@@ -9,71 +9,143 @@ import React, { useEffect, useRef, useState } from 'react';
...
+                    'dark:border-muted/40 border-muted-foreground/10 border-r-2 border-dashed',
+                )}
+              >
+                <HorizontalProgress key={i} percentage={1} isFirst={false} isLast={i === emptySteps.length - 1} />
+                {i < emptySteps.length - 1 && <HorizontalConnector previousPercentage={1} currentPercentage={1} />}
+              </div>
+              <div className='flex h-40 w-40 flex-col sm:hidden'>
+                <VerticalProgress key={i} percentage={1} />
+                {i < funnel.steps.length - 1 && <VerticalConnector previousPercentage={1} currentPercentage={1} />}
```
Lindharden: Same mix-up as above. Not visible in practice (mobile-only block), but worth aligning for consistency.

dashboard/src/components/funnels/FunnelBarplot.tsx
```diff
@@ -170,10 +255,14 @@ function FunnelStep({
...
       </TooltipTrigger>
       <TooltipContent sideOffset={8} className='bg-popover border shadow-xl'>
-        <FilterDescription
-          filter={step.step}
-          className='text-popover-foreground rounded-md p-2 font-medium'
-        />
+        <div className='text-popover-foreground flex flex-wrap items-center gap-1 rounded-md p-2 font-medium'>
+          {step.step.filters.map((filter, i) => (
+            <React.Fragment key={filter.id}>
+              {i > 0 && <span className='text-muted-foreground/80'>and</span>}
```
Lindharden: Seems like a missing translation

dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/funnels/FunnelDialogContent.tsx
```diff
@@ -1,186 +1,138 @@
...
+          steps={funnelSteps}
+          initialOpenId={initialOpenId}
+          onReorder={setFunnelSteps}
+          onUpdateStep={updateFunnelStep}
+          onRemoveStep={removeFunnelStep}
+          globalPropertyKeys={globalPropertyKeys}
+          hasAttemptedSubmit={hasAttemptedSubmit}
+        />
+        <FunnelBarplot
+          className='hidden max-h-[min(420px,var(--funnel-row-h))] pt-3 lg:flex lg:col-span-11 [&_[data-slot=funnel-barplot-card]]:bg-secondary/50 min-h-[min(24rem,var(--funnel-row-h))]'
```
Comment (Lindharden) [inline dashboard/src/app/(protected)/dashboard/[dashboardId]/(dashboard)/funnels/FunnelDialogContent.tsx]: Seems like no element in FunnelBarplot has that data-slot 🤔

dashboard/src/components/funnels/FunnelBarplot.tsx
```diff
@@ -9,71 +9,143 @@ import React, { useEffect, useRef, useState } from 'react';
...
+        className={cn(
+          'bg-card w-full overflow-x-auto rounded-xl border shadow-sm transition-opacity',
+          fill && 'flex flex-1 flex-col',
+          refetching && 'pointer-events-none opacity-60',
+        )}
+      >
+        <div className={cn('group/steps relative flex w-fit flex-col sm:flex-row', fill && 'sm:flex-1')}>
+          {funnel.steps.map((step, i) => (
+            <FunnelStep
+              key={i}
```
Lindharden: Each step holds ResizeObserver connector state, so reorder/add/remove in the live preview can mis-reconcile. I believe a better approach is to use step.step.id / step.id.

dashboard/src/components/funnels/FunnelBarplot.tsx
```diff
@@ -9,71 +9,143 @@ import React, { useEffect, useRef, useState } from 'react';
...
+              step={step}
+              index={i}
+              funnel={funnel}
+              hasEmptySteps={hasEmptySteps}
+              fill={fill}
+            />
+          ))}
+          {emptySteps?.map((step, i) => (
             <div
+              key={i}
```
Lindharden: Same as above 😄

Lindharden (changes requested):
Awesome addition. Excited to see it in production 😄 

1) The "no filters" turn red to indicate invalid state / missing filters, but it is possible to save filters without any filter values. Perhaps wwe should disallow this:

Image

Although keep in mind that the DB entity representation should allow this, so that existing funnels with empty values does not cause issues.

2) There seems to be some disalignments in the dialog, see for instance these icons where all disaligns vertically:

Image

3) The "Add step" button seems mispositioned. It currently resides in the same section as the preview, but not where the actual steps are located. The button should be colocated because they are strongly connected.

Image

4) These should not be our primary blue, but muted/outline styled:

Image

dashboard/prisma/schema.prisma
Lindharden: #963

flopper123 (dismissed):

Lindharden (approved):

flopper123 (approved):

PR #926, opened 2026-05-15
Add a hard cap on the maximum allowed filter rows

Woldbye:
Limit the max number of filters to 10
- If user adds more than 10 filters via URL parameters, it will be trimmed down to the first 10 filters.
- If user attempts to add more than 10 filters in the UI, it will be blocked (see Demo screenshot below)
- If user has 10 filters already, and he clicks a multi table to trigger an 11th filter:
   - If: one of the existing 10 filters are the same column, it replaces (same existing behavior)
   - else: prevent adding the filter, and show the warning toast (see demo below)

**Important**: IF a user has more than 10 filter rows saved, the wont be able to view any of their saved filters in the UI once this gets merged. Hence, this should not be merged before we have confirmed that no users has saved a filter with more than 10 filter rows saved. 

## Demo
image

Toast when adding more than 10 filters and clicking a multi table
image


Lindharden (approved):
Awesome 👍

dashboard/src/entities/analytics/savedFilters.entities.ts
```diff
@@ -11,7 +11,7 @@ export const SavedFilterEntrySchema = z.object({
 export const SavedFilterSchema = z.object({
   id: z.string().cuid(),
   name: z.string().min(1).max(64),
-  entries: z.array(SavedFilterEntrySchema),
+  entries: z.array(SavedFilterEntrySchema).max(MAX_FILTER_ROWS),
```
Lindharden: This will cause all existing filters to disappear for all users who already has saved filters with more than 10 filters, regardless of whether these filters are applied or not.

Lindharden (changes requested):
I found a backwards compatiblity bug that needs to be fixed 😄

Woldbye:
> I found a backwards compatiblity bug that needs to be fixed 😄 

Actually not a bug. See @important section in pr description.

I want to double check Max saved rows on production: this might be an imaginary problem (i double anyone has over 10 filter rows on a single filter saved).

Lindharden:
> > I found a backwards compatiblity bug that needs to be fixed 😄
> 
> Actually not a bug. See @Important section in pr description.
> 
> I want to double check Max saved rows on production: this might be an imaginary problem (i double anyone has over 10 filter rows on a single filter saved).

What I suggest we can do, is simply remove the zod .max validation from the `SavedFilterSchema` and keep the `.max` on `CreateSavedFilterSchema`. That way **if** any saved filters already exists with more than 10 entries, it can still load them, but creating will be blocked. What do you think of this? 😄 

This will also further prevent issues with self-hosters when they upgrade, in case any self-hosters have saved filters with more than 10 filters.

flopper123 (commented):
General question: Should this cap also be present for Funnels steps? 😄

Lindharden (approved):

PR #938, opened 2026-05-18
Site setup mails

Lindharden:


dashboard/src/repositories/postgres/sentEmail.repository.ts
```diff
@@ -18,3 +18,16 @@ export async function recordSent(
...
+ * Returns the set of recipientKeys that already have a sent email for the given campaign.
+ * Use to bulk pre-dedup candidate sets before enqueueing emails.
+ */
+export async function findSentRecipientKeys(recipientKeys: string[], campaignKey: string): Promise<Set<string>> {
+  if (recipientKeys.length === 0) return new Set();
+  const rows = await prisma.sentEmail.findMany({
+    where: { campaignKey, recipientKey: { in: recipientKeys } },
+    select: { recipientKey: true },
+  });
+  return new Set(rows.map((r) => r.recipientKey));
```
flopper123, line 32: Since this model has `@@unique([recipientKey, campaignKey])` each returned `recipientKey` will already be unique 😄

dashboard/src/repositories/postgres/user.repository.ts
```diff
@@ -205,6 +207,32 @@ export async function acceptTermsForUser(userId: string, version: number): Promi
...
+  window: { signedUpAfter: Date; signedUpBefore: Date },
+  limit: number,
+): Promise<UserWithoutDashboardCandidate[]> {
+  try {
+    const users = await prisma.user.findMany({
+      where: {
+        deletedAt: null,
+        email: { not: null },
+        createdAt: { gt: window.signedUpAfter, lt: window.signedUpBefore },
+        dashboardAccess: { none: { dashboard: { deletedAt: null } } },
```
flopper123: This could maybe feel a little spammy in cases like:
A user is invited to another user's dashboard, they use it for a day or two, they decide it's not for them, they delete the dashboard, but the user doesn't delete their user. Then they'd receive the nudge email.

It's an edge-case of course, but it may make sense to make the filter handle some cases like that. Thoughts? 🤔

dashboard/src/repositories/postgres/user.repository.ts
```diff
@@ -205,6 +207,32 @@ export async function acceptTermsForUser(userId: string, version: number): Promi
...
+      },
+      select: { id: true, email: true, name: true },
+      orderBy: { createdAt: 'asc' },
+      take: limit,
+    });
+
+    return users.flatMap((u) =>
+      u.email ? [UserWithoutDashboardCandidateSchema.parse({ userId: u.id, email: u.email, name: u.name })] : [],
+    );
```
flopper123: Side-note: We should probably make `email` required.

Quibble:
```suggestion
    return users
      .filter((u) => u.email)
      .map((u) => UserWithoutDashboardCandidateSchema.parse({ userId: u.id, email: u.email, name: u.name }));
```

flopper123 (changes requested):
Super nice. Only a few comments 😄

dashboard/src/repositories/postgres/sentEmail.repository.ts
Lindharden: The Set isn't there for deduplication but for O(1) lookup inside the candidate loop 😄

dashboard/src/repositories/postgres/user.repository.ts
Lindharden: I have removed this. But we cannot prevent users who were invited to a dashboard and had their access revoked. They'll get an email. We can fix this once access entries are soft-deleted.

flopper123 (approved):
Very nice 😄

PR #939, opened 2026-05-18
Added security emails suite

Lindharden:
Adds 2FA disabled/enabled email + password changed email.

### DIFF
--- a/dashboard/src/services/email/template/two-factor-disabled-mail.tsx
+++ b/dashboard/src/services/email/template/two-factor-disabled-mail.tsx
@@ -1,0 +1,41 @@
+import type { EmailData } from '@/services/email/types';
+import { EmailLayout, Greeting, H1, P, PrimaryLink, renderEmailTemplate } from './_components';
+
+const CAMPAIGN = 'two_factor_disabled';
+
+export interface TwoFactorDisabledEmailData extends EmailData {
+  userName: string | null;
+}
+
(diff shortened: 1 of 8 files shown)


dashboard/src/services/email/template/two-factor-disabled-mail.tsx
```diff
@@ -0,0 +1,41 @@
...
+      preview='Two-factor authentication has been disabled on your Betterlytics account'
+      campaign={CAMPAIGN}
+    >
+      <H1>Two-factor authentication disabled</H1>
+
+      <Greeting userName={userName} />
+
+      <P>
+        Two-factor authentication (2FA) was just disabled on your Betterlytics account. Your account is now
+        protected by password only.
```
flopper123, line 22: OAuth users can also enable 2FA, so it probably makes sense to rewrite the `"Your account is now protected by password only."`. However, for OAuth users it doesn't seem like we actually check for 2FA, which we probably should do 😄

flopper123 (changes requested):

dashboard/src/services/email/template/two-factor-disabled-mail.tsx
Lindharden: I have opened a [PR ](https://github.com/betterlytics/betterlytics/pull/938) for properly disabling 2FA for oauth providers

flopper123 (dismissed):

flopper123 (approved):

PR #945, opened 2026-05-21
Fix form-dialog's close flickers

Woldbye:
Many of our dialogues reset the form before the dialogue has actually finished animation. This causes a layout shift / flicker, as the content clears before the overlay closes.

The pattern I have used to resolve this depends on the type of dialogue:
- If the dialogue should always be empty (default value) when it opens, the simplest fix is to clear it when it opens. 
   * Applied to `SaveQueryFilterDialog`
   * Applied to `CreateDashboardDialog` 

- If the dialogue should only clear on specific actions, the best pattern is to use a ref to track when to clear, and bind it to `onAnimationEnd` - I have created a reusable hook to simplify this (`useOverlayReset`) as part of this PR.
   * Applied to `CreateMonitorDialog`
   * Applied to `CreateFunnelDialog`
   * Applied to `EditMonitorSheet`
   * Applied to `BugReportDialog`
   * Applied to `AnnotationDialogs`

**Important**: `CreateMonitor` has a small behavior change beyond the flicker fix. I noticed that the Cancel button used to silently not clear at all.  I have adjusted it to actually clear, since there's no reason to have a cancel and close button, if they do the same


flopper123 (approved):
Awesome! 😄

PR #947, opened 2026-05-22
Removed empty name block on save button

Lindharden:


Woldbye:
Would be nice with some context - is this related to the bug reported by @giorgiodg ?

flopper123 (approved):

Lindharden:
> Would be nice with some context - is this related to the bug reported by @giorgiodg ?

Yes 😄

PR #948, opened 2026-05-22
Recapture originalTheme on dialog open

### LINKED ISSUE #843: Cancelling UserSettings reverts the theme to default

Lindharden:
Closes #843

### DIFF
--- a/dashboard/src/components/userSettings/UserSettingsDialog.tsx
+++ b/dashboard/src/components/userSettings/UserSettingsDialog.tsx
@@ -119,7 +119,7 @@ function UserSettingsDialogContent({
 }: UserSettingsDialogContentProps) {
   const { setTheme } = useTheme();
   const { data: session } = useSession();
-  const [originalTheme, setOriginalTheme] = useState<Theme | undefined>(settings.theme);
+  const [themeOnOpen, setThemeOnOpen] = useState<Theme | undefined>(settings.theme);
   const { isFeatureFlagEnabled } = useClientFeatureFlags();
   const router = useRouter();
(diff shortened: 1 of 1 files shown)


flopper123 (dismissed):

dashboard/src/components/userSettings/UserSettingsDialog.tsx
```diff
@@ -119,7 +119,7 @@ function UserSettingsDialogContent({
 }: UserSettingsDialogContentProps) {
   const { setTheme } = useTheme();
   const { data: session } = useSession();
-  const [originalTheme] = useState<Theme | undefined>(settings.theme);
+  const [originalTheme, setOriginalTheme] = useState<Theme | undefined>(settings.theme);
```
Woldbye: Name is misleading. Its technically not the originaltheme anymore 

Could you rename this to something more reflective ? (prevTheme e.g.)

Lindharden: Renamed to literal "setThemeOnOpen", this should spell out exactly when it was captured 😄

Lindharden:
Closing, fixed by #952

PR #949, opened 2026-05-22
Updated favicon fetch to use DDG

### LINKED ISSUE #944: Bug: favicon location is not parsed from HTML body
Currently Betterlytics assumes that the favicon is at the root of the domain. For many domains, static content, like favicon.ico/svg/png is hosted elsewhere. 

Therefore, Betterlytics should attempt to parse `<link rel="icon">` / `<link rel="apple-touch-icon">` so that the Betterlytics dashboard shows the right favicons of the websites in the dashboard.

Lindharden:
Closes #944.

Previously the favicon proxy only probed a fixed list of well-known root paths (`/favicon.ico`, `/favicon.png`, `/favicon.jpg`, `/apple-touch-icon.png`) on the user's domain. Some sites declare their favicon via `<link rel="icon">` pointing at a CDN or hashed asset path, so those sites showed no favicon in the dashboard.

This PR replaces the well-known-path probing with a single request to DuckDuckGo's favicon service (`https://icons.duckduckgo.com/ip3/<domain>.ico`), which parses the site's HTML internally and handles redirects, CDN-hosted assets, and other edge cases for us.

For a privacy-focused product, DDG is the standard choice: it solves the parsing problem without us shipping an HTML parser, and unlike a roll-your-own approach it handles Cloudflare-fronted sites, redirects, and manifest icons.

- We **proxy server-side** rather than letting browsers hit DDG directly. End-user IPs are not exposed to DDG.
- DDG sees a list of domains Betterlytics has looked up favicons for. This is metadata about our customer base, not about end-users (visitors). The visitor-privacy story is unaffected.
- Existing SSRF guard, byte cap, redirect limit, and timeout are preserved.

- `dashboard/src/app/api/favicons/route.ts`: replaced the protocols × paths loop with a single DDG fetch. Removed now-unused `FAVICON_PATHS` and `SUPPORTED_PROTOCOLS` constants. All existing safety kept.

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -74,7 +74,6 @@
     "dotenv": "^16.5.0",
     "geojson": "^0.5.0",
     "i18n-iso-countries": "^7.14.0",
-    "ipaddr.js": "^2.2.0",
     "is-cidr": "^6.0.1",
     "is-ip": "^5.0.1",
     "leaflet": "^1.9.4",
--- a/dashboard/src/app/api/favicons/route.ts
+++ b/dashboard/src/app/api/favicons/route.ts
@@ -1,18 +1,14 @@
 import { NextRequest, NextResponse } from 'next/server';
-import { lookup } from 'dns/promises';
-import ipaddr from 'ipaddr.js';
 import { domainValidation } from '@/entities/dashboard/dashboard.entities';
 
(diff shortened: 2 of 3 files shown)


flopper123 (approved):

bjorndbuilder:
Thank you for fixing this!

PR #952, opened 2026-05-23
User settings dialog update

Lindharden:


dashboard/src/app/(protected)/layout.tsx
```diff
@@ -17,13 +19,19 @@ export default async function ProtectedLayout({ children }: { children: React.Re
     }
   }
 
+  const userSettingsResult = await getUserSettingsAction();
+  if (!userSettingsResult.success) {
+    throw new Error('Failed to load user settings');
+  }
+  const initialSettings = userSettingsResult.data;
+
```
flopper123, line 25: Question: I think `session.user.settings` contains the same data as `getUserSettingsAction`, as they both call `getUserSettings` under the hood.
Is there any reason as to why not just use the `session.user.settings`? 😄

dashboard/src/components/userSettings/account/UserDangerZoneSettings.tsx
```diff
@@ -0,0 +1,79 @@
...
+    if (!session?.user?.id) {
+      toast.error(t('toast.unable'));
+      return;
+    }
+
+    startTransition(async () => {
+      const result = await deleteUserAccountAction();
+      if (result.success) {
+        toast.success(t('toast.success'));
+        window.location.href = '/';
```
flopper123: Question: Should this do this instead to also handle cookie deletion?
```ts
import { signOut } from 'next-auth/react';
...
  if (result.success) {
    toast.success(t('toast.success'));
    await signOut({ callbackUrl: '/' });
  }
...
```

dashboard/src/app/(protected)/layout.tsx
Lindharden: Yes, in general it's bad practice to rely on settings storage within the session. The session object should primarily be used for authorization, not source of truth of settings. But I've reverted it to use the session for now, until we fix the general issue with sessions 👍

flopper123 (dismissed):

flopper123 (approved):

PR #958, opened 2026-05-25
Task/blog eu ga alternatives

flopper123:


docs/src/app/blog/components/BlogCta.tsx
```diff
@@ -0,0 +1,38 @@
...
+};
+
+export function BlogCta({ href, children, variant = "primary" }: Props) {
+  const isExternal = /^https?:\/\//.test(href);
+
+  const className = cn(
+    "not-prose inline-flex items-center rounded-md px-5 py-2.5 text-sm font-semibold transition-opacity",
+    variant === "primary"
+      ? "bg-primary text-primary-foreground hover:opacity-90"
+      : "border-border text-foreground hover:bg-muted border"
```
Woldbye: There is no secondary in the variant for secondary. I suggest renaming it to default if you dont plan on using secondary

docs/src/app/api/og/blog/route.tsx
```diff
@@ -0,0 +1,217 @@
+import { ImageResponse } from "next/og";
+import { NextRequest } from "next/server";
+import { readFile } from "node:fs/promises";
+import { join } from "node:path";
+import { getBlogPostBySlug } from "@/app/blog/lib/registry";
+
+// Variant E.3.3 of the locked design: dark canvas, two hairline dividers,
```
Woldbye: Trim AI design comment

Woldbye (commented):
Extremely good idea with a blog feature!!  Some general feedback from my first round of review (note i haven't been through all the files yet)

**Major:**
* M1: The layout, deployment and runtime-resources are deeply interconnected and shared between our /docs and /blog since you hoisted /blog into the /docs repo. I think the cleanest would be to seperate it. If you don't want to do that, some ideas to remedy the issue: 
  * Add seperate error boundaries
  * Decouple sitemap/generateStaticParams, so blog doesn't break docs build
  * Move cover generation til build time, so it doesn't block runtime-resources  (probably a good idea either way)
* M2: A lot of raw icon components hoisted and added to the project here. The BlogPostHeader and BlogPostFooter has the exact same icon declared inline (ArrowLeftIcon) and the footer has the identical `ArrowRightIcon` and `NextArrowIcon` declared in the same file. How about pulling in Lucide icons before it grows out of control? 

**Quibble:** 
* Q1: The new folder convention introduced by @Lindharden seems to be that /app components live in `_components` not `components`. I suggest following this pattern, since you are making a new folder anyway.

* Q2: The content of the blog post is not properly centered on mobile: There's more margin on left-hand-side than right-hand-side.
Image

Woldbye:
M1: Moving the covers to build time is exactly what I was after! 
Though I don't see an error/not-found boundary for /blog. I think a blog-side throw lands on the shared error page. How about adding a blog-level error.tsx + not-found.tsx and have the registry skip-n-warn on a bad post instead of throwing? That gets most of the isolation without doing the actual monorepo split.

M2: I still see a lot of svg duplicaiton. ArrowRightIcon and NextArrowIcon in BlogPostFooter are the same svg in different size. I think it would clean the code a lot to pull in lucide-react as we use in the dashboard. WDYT?

Q1: Resolved
Q2: Resolved

docs/src/app/blog/page.tsx
```diff
@@ -0,0 +1,156 @@
...
+                    <span className="reading">
+                      {featured.readingTimeMinutes} min read
+                    </span>
+                  </div>
+                </div>
+              </Link>
+            )}
+
+            <div className="bi-grid">
+              {posts.map((p) => {
```
Woldbye, line 125: the featured post also appears in the grid below.  I checked some other blogs and it seems to be a standard pattern to show it in both. Just checking to confirm its intentional

docs/src/app/blog/[slug]/opengraph-image.tsx
```diff
@@ -0,0 +1,204 @@
...
+const LOGO_PATH = join(
+  process.cwd(),
+  "public",
+  "betterlytics-logo-light-simple.svg"
+);
+
+let fontCache: { w500: Buffer; w700: Buffer } | null = null;
+let logoDataUrl: string | null = null;
+
+async function loadFonts() {
```
Comment (Woldbye) [inline docs/src/app/blog/[slug]/opengraph-image.tsx:51]: Have you hit /api/og/blog on an actual built/deployed image? the docs Dockerfile only copies public/, .next/standalone and .next/static into the runner. Concerned it will 500s with ENOENT on the font read

docs/next.config.mjs
```diff
@@ -14,5 +14,4 @@ if (result.error) {
 
 export default withNextra({
   output: "standalone",
-  basePath: "/docs",
```
Woldbye, line 17: Remember to update the reverse proxy to match before we go into prod, otherwise the docs/blog load from / and collide with the dashboard app.

docs/src/app/blog/lib/registry.ts
```diff
@@ -0,0 +1,101 @@
...
+  const slugs = listBlogMdxFiles();
+  const posts: BlogPost[] = [];
+
+  for (const fileSlug of slugs) {
+    const page = await importPage(["blog", fileSlug]);
+    const fm = page.metadata as Record<string, unknown> | undefined;
+    if (!fm) continue;
+
+    const parsed = blogFrontmatterSchema.safeParse(fm);
+    if (!parsed.success) {
```
Woldbye: Related to my M1: this throw runs inside `generateStaticParams` so a bad frontmatter field fails the entire docs build. How about we console.warn + skip the post instead?

docs/src/app/blog/page.tsx
flopper123: Yes, it's intentional 😎

docs/next.config.mjs
flopper123: Thanks, that should be resolved now! 😄

docs/src/app/blog/[slug]/opengraph-image.tsx
Comment (flopper123) [reply docs/src/app/blog/[slug]/opengraph-image.tsx:51]: No, I had not. But it's resolved now 🫡

Woldbye (dismissed):

Woldbye (approved):
Still lgtm 😇

Issue #626, opened 2025-12-14
Allow users to filter individual steps in User Journey flows


Lindharden:
Currently, the User Journey visualization displays all steps in a visitor’s path by default. Users have requested the ability to filter or highlight specific steps within a journey flow to better analyze how visitors navigate key pages or events.

This feature should allow users to apply URL-based filters to individual steps in the journey. For example:
- Filter so that Step 2 only includes journeys that pass through a specific URL.
- Filter so that journeys start at a specific URL.
- Filter so that journeys end at a specific URL.

These filters should enable users to focus on specific paths and better understand how visitors move through the site.

PR #960, opened 2026-05-31
Per-page filter-column exclusion

Woldbye:
Hides filter columns that can't produce data on a given page, shows any existing filter on such a column as a disabled-but-deletable row, and excludes it from the query while leaving it in the URL.

## Description of why
- Several dashboard pages are locked to a single `event_type` (errors → `client_error`, outbound-links → `outbound_link`, etc.), so columns like `custom_event_name` and `outbound_link_url` can never match data there.
- Previously these columns were still offered in the filter UI and sent to the query, producing confusing always-empty results.
- This makes the available filter columns reflect what each page can actually query, without permanently discarding a user's filters when they navigate between pages.

## Key decisions of PR
- A filter on a hidden column is shown disabled-but-deletable (with a "not available on this page" tooltip) and excluded from the query, but kept in the URL so it re-activates automatically on pages where the column is valid - rather than being silently stripped.
- Per-page hiding and demo-mode restrictions are unified into a single `useFilterColumnStatus` hook that returns a `{ disabled, reason }` status, with per-page hiding taking precedence over the demo reason.
- Visibility is provided via context (`QueryFilterColumnsVisibilityProvider`); the default value is an empty set, so pages without the provider keep full filter visibility (no no-op wrappers needed).
- Contextually-disabled filter rows are kept fully inert/non-tabbable (cursor-not-allowed, disabled controls) rather than focusable. The broader focusable `aria-disabled` + tooltip accessibility model is intentionally left to a separate app-wide a11y PR.
- Funnels, pages, overview, devices, geography, referrers and campaign are intentionally not gated - all of their filter columns are feasible (the events table is fully denormalized).

## Changes
- Added `QueryFilterColumnsVisibilityProvider` context exposing a hidden `Set<TableFilterColumn>` and `useHiddenQueryFilterColumns()`.
- Added `use-is-filter-column-allowed.ts`: `useFilterColumnStatus`, `useFilterColumnDisabledMessage`, `useIsFilterColumnAllowed`, and `useAllowedQueryFilters`.
- Updated the filter UI (FilterColumnDropdown, QueryFilterInputRow, ActiveQueryFilters, QueryFiltersSelectorContent, MultiSelect) to omit hidden columns from the dropdown and render existing filters on them as disabled, tooltip-annotated, deletable rows.
- Gated query execution: `use-analytics-query.ts` runs filters through `useAllowedQueryFilters`, and `use-filter-click.ts` blocks chart-click filters on disallowed columns.
- Wrapped errors, events, outbound-links, user-journey and web-vitals pages with the provider, each declaring the columns that don't apply there.
- Added the `components.filters.notAvailableOnPage` string to the da/en/it/nb message files.


flopper123 (commented):
Very nice! 😄
Has some awesome additions, it's nice that it simply ignores the bad filter now!

However, I've played around with it for a bit and think it's worth having a discussion about whether removing the filter from the dropdown list fully is the correct play, instead of just disabling it:

1. The user doesn't know that the filter is missing: To give you an example, I could imagine that someone goes to our User Journey page to see all the journeys where a user has a specific custom event emitted. They'll have no way of knowing that the filter is not there since we've hidden it because we haven't modeled the User Journey data that way, which is frustrating since they'll probably look at the filters list for a good while before giving up or realizing it's not there. So the lack of feedback is what I'm concerned about mainly.

2. It creates inconsistencies: We usually don't just hide things, but instead disabling them with feedback. So this feels like it goes a bit against that. I understand your point that it's "bloat", however, from what I can see it's max 1-2 filters on some pages that'll need to be disabled, which we can give users feedback on.

Let me know what you think! 🫡

dashboard/src/components/filters/FilterColumnDropdown.tsx
```diff
@@ -51,13 +61,15 @@ export function FilterColumnDropdown<TEntity>({
...
             className={cn(
-              'flex h-9 w-full items-center justify-between gap-2 rounded-md border border-input bg-transparent px-3 py-2 shadow-xs',
+              'border-input flex h-9 w-full items-center justify-between gap-2 rounded-md border bg-transparent px-3 py-2 shadow-xs',
               'cursor-pointer text-sm whitespace-nowrap outline-none',
               'dark:bg-input/30 dark:hover:bg-input/50',
               'data-[placeholder]:text-muted-foreground',
               '[&_svg]:text-muted-foreground [&_svg:not([class*="size-"])]:size-4',
-              'focus-visible:border-ring focus-visible:ring focus-visible:ring-ring/50',
+              'focus-visible:border-ring focus-visible:ring-ring/50 focus-visible:ring',
+              'disabled:cursor-not-allowed disabled:opacity-50',
```
flopper123: Quibble: `cursor-not-allowed` feels a little aggressive for disabled filters 😄 Consider just using regular cursors

dashboard/src/components/filters/QueryFilterInputRow.tsx
```diff
@@ -41,27 +46,42 @@ export function QueryFilterInputRow<TEntity>({
...
-      />
+      <DisabledTooltip disabled={disabled} message={disabledMessage} wrapperClassName='[grid-area:col]'>
+        {(isDisabled) => (
+          <FilterColumnDropdown
+            filter={filter}
+            onFilterUpdate={onFilterUpdate}
+            globalPropertyKeys={globalPropertyKeys}
+            disabled={isDisabled}
+            className='[grid-area:col]'
+          />
```
flopper123, line 57: Filter column dropdown still works on a hidden filter:

Image

dashboard/src/components/filters/ActiveQueryFilters.tsx
```diff
@@ -1,43 +1,65 @@
...
+                  className='text-muted-foreground/80 size-6 cursor-pointer hover:text-foreground focus-visible:text-foreground -mx-0.5'
+                  onClick={() => removeQueryFilter(filter.id)}
+                >
+                  <XIcon className='size-3.5' />
+                </Button>
+              </TooltipTrigger>
+              <TooltipContent>{t('removeFilter')}</TooltipContent>
+            </Tooltip>
+          </Badge>
+        );
```
flopper123, line 53: I think it may make sense to wrap it all in a single `DisabledTooltip` like elsewhere. Currently there seems to be a different line height between the two:

This is the two active filters next to each other. The text is on the same line height:
Image

Here the event is hidden, but the text is slightly higher than on the other screen 😄 
Image

One you see it, it can't be unseen 🤣

dashboard/src/hooks/use-filter-click.ts
```diff
@@ -30,8 +32,8 @@ export function useFilterClick(defaults?: Options) {
 
   const applyFilter = useCallback(
     (column: FilterColumn, value: string, opts?: Options) => {
-      if (isDemo && column !== 'url' && column !== 'device_type') {
-        toast.info(t('interactionDisabled'));
+      if (!isFilterColumnAllowed(column)) {
+        if (isDemo) toast.info(t('interactionDisabled'));
         return;
```
flopper123, line 37: Quibble: Consider also adding feedback for not-allowed once

flopper123 (changes requested):
Code/etc

dashboard/src/components/filters/FilterColumnDropdown.tsx
Woldbye: Agree, I did it to match the default from the shadcn component we use. I have instead overwritten default, so all disabled on the filters are regular cursors.

dashboard/src/hooks/use-filter-click.ts
Woldbye: Done in 4bea546. It only toasted in demo mode before, so a page-hidden click returned silently. It now branches on the disabled reason and reuses the same message the row tooltip shows: interactionDisabled for demo, notAvailableOnPage for page.

dashboard/src/components/filters/QueryFilterInputRow.tsx
Woldbye: Fixed in 84ed7d7 by passing disabled to BADropdownMenuTrigger. The `<button>` had disabled, but Radix gates the open logic on the Trigger's own disabled prop rather than the child's, so pointerdown still reached `onOpenToggle`.

Woldbye:
@flopper123 Looked through your review comments and resolved the code ones :)

**On hidden vs disabled:**
I'd argue hiding is what keeps this consistent. We already use disabled to mean "you could unlock this": `PermissionGate` (get access) and `CapabilityGate` (upgrade). Those are states the user can resolve. Page-incompatibility isn't - nothing a user does on the errors page makes `custom_event_name` meaningful there. Reusing that affordance for "never applicable" overloads a signal that currently means "unlockable" and leaves a control nothing can enable.

 The PR already applies both treatments in the same dropdown:
  - Demo-blocked columns: shown, greyed, "Not available". Because demo is unlockable - the user can resolve it by leaving demo.
  - Page-incompatible columns: hidden. Because nothing the user does on the errors page makes `custom_event_name` mean anything.

disabled means you can unlock it, hidden means it can never apply here.

Though worth mentioning that the count does bite for #626. I have it implemented on a local branch: each journey step gets a header with its own subset of the dashboard filters. This is exactly why visibility sits behind a provider here - each step can scope differently from the page. Disabling there would leave a wall of permanently dead controls in every step.

dashboard/src/components/filters/FilterValueSearch.tsx
```diff
@@ -56,6 +58,7 @@ export function FilterValueSearch<TEntity>({
   return (
     <MultiSelect
       options={multiSelectOptions}
+      disabled={disabled}
```
Lindharden, line 62: Disabled rows still fetch value suggestions because FilterValueSearch always mounts useQueryFilterSearch, which fires filters.getFilterOptions for the hidden column as soon as the popover opens. Consider passing disabled into the hook and gating the query with enabled: shouldSearchServer && !disabled.

dashboard/src/components/filters/ActiveQueryFilters.tsx
```diff
@@ -1,43 +1,63 @@
...
+            key={filter.id}
+            variant='outline'
+            className={cn(
+              'border-input bg-muted/50 hover:bg-muted/70 dark:bg-secondary dark:hover:bg-secondary/90 gap-1.5 p-1 px-1.5',
+              status.disabled && 'opacity-50',
+            )}
+          >
+            {disabledMessage ? (
+              <Tooltip>
+                <TooltipTrigger asChild>
```
Lindharden: This tooltip does not work for some reason, I think it might be more appropriate to use our existing DisabledTooltip wrapper 😄

Lindharden (changes requested):
Very nice @Woldbye. I left just a few comments. And I also noticed that the Pages page needs gating too, a custom_event_name or outbound_link_url filter always produces zero rows there. Suggest wrapping it with hide={['outbound_link_url', 'custom_event_name']} like user-journey.

Lindharden:
@Woldbye I'm leaning towards providing a tooltip on the disabled event types with feedback such as "_Not available on this page_", mostly for the following reasons:
- They're also disabled in the demo page with a similar tooltip feedback
- It makes it explicitly clear to the user that it's not a bug, but rather that the event itself is not supported on this page
- Prevents users from wasting time trying to find filters without knowing they're disabled
- Likely prevents users from filing bug-report for missing filters.
- Minorly it also makes filters discoverable

If this ever grows to a point, where it's 5-6 filters per page, then I'd argue hiding them entirely might be a better play. But I'm not sure.

Woldbye:
@Lindharden @flopper123 Please review when you have the time. 

I have resolved the code comments in the latest commits

> I'm leaning towards providing a tooltip on the disabled event types with feedback such as "Not available on this page",

I did what you and @flopper123 want: They are now disabled with an informative message. Note that I have deliberately left the "Hide" path in the code, even though I know this breaks YAGNI principle. I absolutely need this for the upcoming user-journey filters, as there will only be 2-5 filters per user-journey step. I could technically leave it out, and include in the user journey PR, but it feels wrong to not include it here, when it has been thoroughly tested and verified already here.

> ...the Pages page needs gating too, a custom_event_name or outbound_link_url filter always produces zero rows there. Suggest wrapping it with hide={['outbound_link_url', 'custom_event_name']} like user-journey.

I have deliberately kept the /pages page to all filters. The `Exit` and `Entry` tabs makes perfect sense with both filters.
Its a design decision of the feature, to ONLY hide filter X if ALL consumers on the page cant use a filter X.

image

image

Lindharden:
> I have deliberately kept the /pages page to all filters. The `Exit` and `Entry` tabs makes perfect sense with both filters. Its a design decision of the feature, to ONLY hide filter X if ALL consumers on the page cant use a filter X.

Fair, I forgot about Entry/Exit. Agreed on keeping Pages ungated then 😄

Lindharden (approved):

flopper123 (approved):
Awesome 👏👏👏

PR #965, opened 2026-06-08
Add secret rotating salt for visitor fingerprint anonymization

Lindharden:
- Secret 16-byte random salt, rotated daily, persisted in Postgres (`AnalyticsSalt`) so that it survives restarts.
- Postgres, not ClickHouse, for two reasons: (1) keeps the secret separate from the analytics/fingerprint data, so a ClickHouse breach can't unmask visitors;  (2) Postgres is transactional (atomic `ON CONFLICT` upsert), whereas ClickHouse treats
  per-day deletes as heavyweight async mutations.
- Rotation is lazy + fire-and-forget: the first event past UTC midnight kicks off a background rotation; events never block on it (in-flight guard + retry cooldown).
- Keep today's + yesterday's salt so sessions survive the midnight rotation; old salts are pruned, so once a day's salt is gone its fingerprints can no longer be reversed.


backend/src/salt/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+    info!("Salt service initialised for UTC date {}", today);
+    Ok(())
+}
+
+/// Return today's `current` salt and the `previous` salt.
+///
+/// Lock-free read on the hot path. When the UTC day has rolled over, the cached salts (still
+/// a valid secret, just yesterday's) are returned immediately and the rotation is kicked off
+/// in the background, so an event never blocks on Postgres.
+pub async fn current_and_previous() -> (Salt, Option<Salt>) {
```
flopper123: This does no async work, so `async` can be removed. This will also make the `visitor::identify(...)` sync, which I think is slightly cleaner 😄

flopper123 (changes requested):

flopper123 (approved):

PR #967, opened 2026-06-11
Added session cache warming on bootup

Lindharden:
Warms the session cache on startup. This is to prevent in-flight sessions from being assigned a new session when the backend is redeployed.

### DIFF
--- a/backend/src/db/mod.rs
+++ b/backend/src/db/mod.rs
@@ -11,7 +11,7 @@ use crate::config::Config;
 use crate::processing::ProcessedEvent;
 
 mod models;
-pub use models::{EventRow, ReferrerSourceCategoryRow, SessionReplayRow};
+pub use models::{ActiveSessionRow, EventRow, ReferrerSourceCategoryRow, SessionReplayRow};
 
 const NUM_INSERT_WORKERS: usize = 1;
 const EVENT_CHANNEL_CAPACITY: usize = 100_000;
@@ -36,6 +36,17 @@ impl Database {
         Self::spawn_dispatcher(event_rx, worker_senders);
 
         Ok(Self { clickhouse, event_tx, config })
+    }
+
+    /// Fetch the current session of every visitor active within `window`, from `analytics.sessions`
+    pub async fn fetch_active_sessions(&self, window: Duration) -> Result<Vec<ActiveSessionRow>> {
+        let rows = self
+            .clickhouse
+            .inner()
+            .query(&active_sessions_query(window.as_secs()))
+            .fetch_all::<ActiveSessionRow>()
+            .await?;
+        Ok(rows)
     }
 
     fn create_channels() -> (mpsc::Sender<ProcessedEvent>, mpsc::Receiver<ProcessedEvent>) {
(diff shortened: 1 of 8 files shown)


backend/src/db/mod.rs
```diff
@@ -307,3 +318,17 @@ async fn run_inserter_worker(
     );
     Ok(())
 }
+
+/// SQL to load each active visitor's most recent session. Filtering on `session_end` uses the
+/// `idx_session_end` minmax skip index, so the cost is bounded by the number of *active*
+/// sessions, not total history; `argMax(.., session_end)` picks each visitor's current session.
+fn active_sessions_query(window_secs: u64) -> String {
```
flopper123: Quibble: Pass the Duration to this function and call `.as_secs()` in here, that way you can't accidentally send ms or something like that

backend/src/session/mod.rs
```diff
@@ -1,58 +1,152 @@
...
+    /// Return the visitor's existing session, or assign a new one. On a miss, the lazily-computed
+    /// `previous_fingerprint` is tried so a visitor keeps their session across a midnight salt rotation.
+    fn get_or_create(
+        &self,
+        site_id: &str,
+        visitor_fingerprint: u64,
+        previous_fingerprint: impl FnOnce() -> Option<u64>,
+        event_timestamp: DateTime<Utc>,
+    ) -> (u64, DateTime<Utc>) {
+        let cache_key = format!("{}-{}", site_id, visitor_fingerprint);
```
flopper123, line 45: This feels a little off that we need to prefix with `site_id` to ensure tenant isolation. We should probably consider adding `site_id` to the fingerprint generation instead, no? That would prevent the issue entirely 🤔

backend/src/session/mod.rs
```diff
@@ -1,58 +1,152 @@
...
 
-    if let Some(entry) = SESSION_CACHE.get(&cache_key) {
-        return Ok((entry.session_id, entry.created_at));
-    }
+/// Pre-populate the cache with recovered sessions so they survive a restart; returns the number
+/// inserted. (moka resets the idle timer on insert, so a warmed entry can outlive its real last
+/// activity by up to `SESSION_EXPIRY`)
+pub fn warm(sessions: impl IntoIterator<Item = WarmSession>) -> usize {
+    SESSION_CACHE.warm(sessions)
+}
```
flopper123, line 108: Question: So this could in theory cause a session to be reactivated after 59 minutes. If the server is restarted like 1 minute before a session expires, then on cache warm, the session expiry will be refreshed 🤔 I don't think it's an actual issue though, but could it cause any data issues or something like that? 🤔

flopper123 (commented):
Minor things

backend/src/session/mod.rs
Lindharden: Changing the hash inputs would change every visitor_id value going forward. Out of scope

backend/src/session/mod.rs
Lindharden: I think you missed the comment. It's documented there:

> (moka resets the idle timer on insert, so a warmed entry can outlive its real last
/// activity by up to `SESSION_EXPIRY`)

flopper123 (approved):

PR #972, opened 2026-06-13
Highlight empty filter values in funnel steps and lock the contract

### LINKED ISSUE #946: Disallow empty filter values in funnels
Since adding OR filters to Funnels, you can now create funnels with empty values, like:

Image

The direct fix is to simply add `.min(1)` to:
```ts
export const QueryFilterSchema = z.object({
  id: z.string(),
  column: FilterColumnSchema,
  operator: z.enum(FILTER_OPERATORS),
  values: z.array(z.string()).min(1), // <---
});
```

However, a users may've created a funnel with empty values, so we'll probably need to do a data migration to ensure their dashboard doesn't fail on load.
So we'll either need to create a data migration to ensure that's not the case, or query against the DB to ensure no such funnel exists, or just add client side validation for this.

Woldbye:
Closes #946

Empty filter values were already blocked on save by the transform in #924 (the funnel dialog redesign): `''`/`[]` can't be persisted through create/update. This PR just makes that visible in the UI and pins it down with a small test (see screenshot):

  - Red border on empty value inputs, scoped to a step that's actually blocking the save, so it's clear *why* submit is disabled.
  - `isNonEmptyValue` one shared definition of "empty" across the funnel write-path checks.
  - `funnels.entities.test.ts` a regression test guarding the #924 behaviour so it can't silently regress: `''`/`[]` get stripped, while a whitespace-only value is deliberately preserved (ingestion stores whitespace global-property values verbatim, so a funnel may legitimately filter on them).

No schema change, no migration => the DB stays permissive for existing funnels.

image

### DIFF
--- a/dashboard/src/app/(app)/(protected)/dashboard/[dashboardId]/(dashboard)/funnels/FunnelStepAccordionItem.tsx
+++ b/dashboard/src/app/(app)/(protected)/dashboard/[dashboardId]/(dashboard)/funnels/FunnelStepAccordionItem.tsx
@@ -307,6 +307,7 @@ function FunnelStepAccordionItemComponent({
             filters={step.filters}
             onChange={handleFiltersChange}
             globalPropertyKeys={globalPropertyKeys}
+            showEmptyValueErrors={showFilterEmptyError}
           />
         </AccordionContent>
       </AccordionItem>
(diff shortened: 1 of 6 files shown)


flopper123 (approved):
Awesome! 😄

Issue #977, opened 2026-06-14
Click a value in the Event Details table to filter by it


Woldbye:
Depends on Custom Event Properties being filterable by value (#961)

When you expand a custom event in the Event Details table on /events, you get the top values for each property - but they're just for show right now. You should be able to click a value to filter the dashboard by it, same as clicking a value in the overview breakdown tables.

Keep the property row's expand/collapse as-is; just make the values themselves clickable, with the usual "Filter by …" hover.

Notes now that the row-click conventions (#1018/#1022) have landed:

- A value click should apply `custom_event_name = <event>` and `cep.<key> = <value>` atomically through `useFilterClick.applyFilters`, like the compound browser + version rows in the Devices breakdown. Demo mode and per-page gating then come for free.
- Property rows should carry the custom event property source icon (`BracesIcon`, sized `h-4 w-4`), matching the filter dropdown and the tags icon on the overview Properties tab.

Issue #973, opened 2026-06-13
Not-equals (!=) filters on event-scoped columns have no effect on session pages


Woldbye:
### Problem

On session-grain pages (e.g. Geography, Devices, Pages), a not-equals (`!=`) filter on an event-scoped column - `custom_event_name`, `url`, or `outbound_link_url` - has effectively no effect. The result still includes virtually all sessions instead of excluding the ones matching the filtered value.

The equals (`=`) form of the same filter works correctly on these pages; only the negation is affected. The same `!=` filter also behaves correctly on event-grain pages - the discrepancy is specific to session-grain pages.

### Steps to reproduce

1. Open a session-grain page (e.g. Geography).
2. Add a filter: `custom_event_name != <an event name that exists in your data>`.
3. The result is essentially unchanged - sessions that contain that event are not excluded.

### Expected

`column != value` on a session-grain page should exclude sessions matching the value, consistent with the `=` behaviour and with how the same filter behaves on event-grain pages.

### Scope

- Affects the event-scoped columns `custom_event_name`, `url`, and `outbound_link_url` on session-grain pages.
- Global properties and session-native columns (device, country, browser, etc.) are not affected.

PR #980, opened 2026-06-20
Add custom event property filters

### LINKED ISSUE #961: Add support for filtrering by custom event property values
It should be possible for users to filter their custom event properties by specific values through our dashboard filters

This likely requires a db migration to the same representation as the global properties

### LINKED ISSUE #1016: Drop sampling from custom event property discovery
From the #980 review: property key/value discovery applies the high-traffic SAMPLE gate, which is driven by total event volume. Custom events are typically a small fraction of a high-traffic site's events, so rare cep keys and values can vanish from the filter dropdowns on exactly the sites with the most data.

The #961 bench notes recommend unsampled cep discovery - the custom-guarded scan is already pruned by the (site_id, event_type, date) primary key, so its cost is bounded by custom-event volume, not total volume.

Consider dropping the SAMPLE from cep discovery (`getPropertyKeys` / `getPropertyValues` in `filters.repository.ts`).

Woldbye:
Make custom event property values filterable across the dashboard, the same way global properties already are (closes #961).

## Key decisions

  - Modeled cep exactly like global properties. The parsed filter column collapses into a single { kind: 'property', source } branch, so every consumer (SQL builders, discovery, UI) has one property code path instead of a gp-specific one.
  - One server-only per-source SQL descriptor. gp reads the global_properties_* arrays; cep reads the custom_event_json via JSONExtractString / JSONHas. No blob migration and no tracking-script, backend, or schema change required.
  - cep is filterable everywhere, not just on the events page. Event-grain queries filter the blob directly; session-grain queries ride the shared events bridge from #985/#987 (one IN per event type, one shared NOT IN), so combined custom_event_name + cep filters must match the same event and != genuinely excludes sessions.

## Changes

  - Add a property-source registry and a shared per-source SQL descriptor; collapse ParsedFilterColumn to a single property kind.
  - Filter event-grain and session-grain queries by custom event properties via the shared events bridge.
  - Generalize property key/value discovery (repository, service, tRPC route) over the source; discovery hides keys that cannot be used as filter columns.
  - Drop invalid query filters individually on URL decode instead of resetting the whole filter state (rest of the malformed-key hardening: #1011).
  - Exclude cep on pages without custom events (web-vitals, errors, user-journey, outbound-links), building on #960.
  - UI: a reusable per-source submenu in the filter column dropdown, source-aware icon/label, and a usePropertyKeys hook; threaded through the filter selector and funnel step editors.
  - With #1007 merged, the MCP query and funnel-preview filter schemas accept cep columns too, described alongside gp; a cep key discovery tool (mirroring list_global_properties) is a follow-up.

## Additional notes

I have deliberately left out #973 and #977


flopper123 (changes requested):
Code looks fine! 😄 

I've found some bugs / feedback:
1 - (Bug) Adding `\t` to the property key erases all your filters:
Try to send the following event `betterlytics.event('yessir', { 'ran\tdom': 'b' })`.
Then you'll see the CEP as `ran dom`. Clicking on it, you'll see two issues:
- No values can be found (though you'd expect to see `b`)
- Insert a random value and apply the filters: Now all the filters are gone, and the page refreshes
We may need to add some more input validation on the filters, or so 😎 

2 -  (Feedback) It's a bit hard to know exactly what event the property is coming from. Would it make sense to prefix the keys with the event name? So would it make sense to do something like `search - page`, `signup - code`, etc.? 😄

flopper123:
On a second note. I think the negation for this is outdated.
I think merging main into this branch and doing it like on main would both give a performance boost and fix negation! 😄
So I'd change the strategy as well like on main.

Lindharden (changes requested):
Amazing.

I ran some benchmarks for this:

## Benchmarks

**perf-large =** 50M events / 10M sessions / 90 days, 9.6% custom events, realistic per-event JSON (high-cardinality order IDs & search queries, low-cardinality plans), gp on 70% of sessions

**perf-heavy =** 20M events, 40% custom with fat 6–10-key blobs (product-analytics style) 

| Query (devices page, session-grain) | perf-large | perf-heavy |
|-------------------------------------|-----------:|-----------:|
| No filter (baseline) | 174ms | 102ms |
| `gp.plan = pro` (existing feature) | 496ms | - |
| `cep.plan = pro` | 268ms | 680ms |
| `cep.plan != pro` | 316ms | 701ms |
| `cep` high-cardinality / wildcard | 232ms | 289ms |
| 3 negated `cep` filters (worst case) | 887ms | 1.7–2.6s |
| 3 negated (old per-filter strategy) | 1032ms, 2.2× the I/O | - |
| `cep =` over full 90 days | 694ms | - |

### **Summary:**
Events page with CEP filters: 67–116ms (large), ~500ms (heavy). 
Key discovery: 69ms cep vs 241ms gp. 
Value discovery: 28–82ms cep vs 1.4s for gp

Apparantly, the existing gp value discovery is the slowest query in the entire suite; CEP is 20x cheaper because `event_type='custom'` prunes on the primary key.

I also directly tested the redesign that #961 mentions (similar approach to Global Properties with parallel `cep_keys`/`cep_values `arrays + bloom index): arrays are 1.6–4x faster on probes and ~20x faster for rare/absent keys (bloom skipping), at ~10 bytes/custom-event storage. 

**This is definitely something we should implement as a follow-up optimization issue, but I don't think it has to be a blocker, nor have to be part of this PR.** I'll create a follow-up issue for this.

## Minor feedback

- Sampling actively hurts discovery UX here. For example, a `SAMPLE 0.25` over `visitor_id` can make a rare key or value vanish from the dropdown entirely. I think we should remove sampling here entirely. In addition, the current sampling system is simply incomplete, so we're trying to migrate away from sampling entirely, until we can introduce a robust and more correct system.

Woldbye:
@Lindharden @flopper123 Ready for your rereview. I'm sorry I had to make quite big changes...
 
 Since last review:
 - Merged main twice and reworked the CEP session filters to ride the events bridge from #985/#987: = joins the per-event-type IN bridge, != collapses into the shared NOT IN. Negation is correct now, and combined `custom_event_name` + `cep.*` filters must match the same event.
- Invalid filters in the URL are dropped individually on decode instead of resetting the whole filter state, so the \t bug can no longer wipe your filters. Remaining hardening (ingest sanitization + discovery filtering) is tracked in #1011.
- CEP is excluded per-page alongside custom_event_name (web-vitals, errors, user-journey, outbound-links), building on #960.
- Reverted my MCP schema widening so #1007 lands untouched; CEP via MCP follows for free after that merge.

**The \t bug:**
This was three stacked issues: 
1. CEP keys aren't sanitized at ingest (gp keys are), 
2. Discovery offers keys the filter schema rejects
3. One invalid filter in the URL reset the entire filter state. 

This PR fixes the last one: invalid filters are now dropped individually, so applying that filter can't erase anything anymore. Ingest sanitization and filtering discovery output are tracked in #1011 since they touch the backend and existing data. `simulate-events.js` now seeds a yessir event with a ran\tdom key if you want to repro it yourself 😄

**Event-name prefix:**
I agree it's hard to see where a key comes from. I'd rather not encode the event name into the key - a key used by N events becomes N dropdown entries. So the filter column would need a new grammar. After the bridge rework, custom_event_name = signup + cep.plan = pro already means "signup events where plan=pro", so the pairing exists by combining filters. My suggestion for a follow-up: annotate/group keys in the dropdown by the event names they occur on, and possibly scope key discovery when a custom_event_name filter is active. Please LMK if you agree, then I will open the issue 😄

**Negation issue:**
Fixed - merged main and folded CEP into the shared bridge from #985/#987: one IN per event type plus one shared NOT IN, so it picked up the perf boost as well.

dashboard/src/repositories/clickhouse/filters.repository.ts
```diff
@@ -74,25 +79,35 @@ export async function getGlobalPropertyKeys(
...
   const searchClause =
     search && search.trim()
       ? safeSql`AND value ILIKE ${SQL.String({ search: `%${search.trim()}%` })}`
       : safeSql``;
 
   const query = safeSql`
-    SELECT DISTINCT global_properties_values[indexOf(global_properties_keys, ${SQL.String({ prop_key: propertyKey })})] AS value
-    FROM analytics.events
+    SELECT DISTINCT ${valueExpr} AS value
+    FROM analytics.events ${sample}
```
Lindharden: We should avoid sampling 😄

Lindharden (changes requested):
Don't know if this was missed, but restating my previous comment! 😄 

> Sampling actively hurts discovery UX here. For example, a SAMPLE 0.25 over visitor_id can make a rare key or value vanish from the dropdown entirely. I think we should remove sampling here entirely. In addition, the current sampling system is simply incomplete, so we're trying to migrate away from sampling entirely, until we can introduce a robust and more correct system.

Lindharden (changes requested):
One small follow-up that I think might be a good idea to include in this PR:

> The MCP describe tool now actively misleads client. Describe tells MCP clients: "This dashboard may record custom event properties you can filter by using the gp.<key> form". That wording predates this PR, but now that cep.<key> exists and is accepted by McpFiltersSchema, it steers clients to the wrong column form for custom event properties, and nothing in the describe output mentions cep.<key> at all. 

The PR body says cep is "described alongside gp" in the MCP schemas, but I couldn't find that anywhere in the code. The follow-up cep discovery tool is fine to defer; the misleading sentence is probably good to fix now 😄

Woldbye:
@Lindharden I removed sampling and adjusted the MCP documentation in latest commits. Please have a look 👍

dashboard/src/mcp/tools/describe.ts
```diff
@@ -37,7 +38,11 @@ export function getSchemaDescription(): SchemaDescription {
...
+        'Global properties are key-value metadata attached to every analytics event sent from a page. Filter them with the gp.<key> form, e.g. { column: "gp.plan", operator: "=", values: ["pro"] }. These are filter-only (not metrics or dimensions). Call the list_global_properties tool to list the available keys, then call it again with a "key" argument to see example values for a specific property.',
+    },
+    customEventProperties: {
+      description:
+        'Custom event properties are the per-event payload of a specific tracked event. Filter them with the cep.<key> form, e.g. { column: "cep.plan", operator: "=", values: ["pro"] }. These are filter-only. Combine with a custom_event_name filter to target one event type - both must match the same event. There is no discovery tool for these keys yet, so use keys known from the site\'s tracking setup.',
```
Lindharden: Small suggestion: the cep example (cep.plan = pro) is identical to the gp example, which blurs the distinction the descriptions are trying to draw, a plan tier is exactly the kind of metadata you'd expect as a global property. Maybe use a genuinely per-event key and show the custom_event_name pairing the text mentions, e.g. custom_event_name = "purchase" + { column: "cep.currency", operator: "=", values: ["USD"] }. Since there's no cep discovery tool yet, the example is the only concrete guidance a client gets 😄

Lindharden (changes requested):
One last thing from me. 😄

Woldbye:
@Lindharden 
Good feedback. You are absolutely right, important to differentiate them

Lindharden (dismissed):

flopper123 (dismissed):

Lindharden (approved):

flopper123 (approved):

PR #981, opened 2026-06-23
Initial public status page backend

flopper123:


Dockerfile.selfhost
```diff
@@ -33,7 +33,8 @@ RUN echo "Building dashboard..."
 ENV NEXT_TELEMETRY_DISABLED=1
 ENV PNPM_HOME="/pnpm"
 ENV PATH="$PNPM_HOME:$PATH"
-RUN corepack enable
+ENV NODE_OPTIONS="--max-old-space-size=4096"
```
Woldbye: What process needs the extra memory? Should probably document with a comment, or it's not clear when it's safe to remove again (if ever)

Dockerfile.selfhost
```diff
@@ -33,7 +33,8 @@ RUN echo "Building dashboard..."
...
+RUN corepack enable && corepack prepare pnpm@10.9.0 --activate
```
Woldbye: I have an open pr for this exact change also.. it's absolutely necessary or the build breaks on a cold cache (https://github.com/betterlytics/betterlytics/pull/990).

Feels a bit weird to sneak it into this pr, as it's a general problem completely unrelated to public status pages, hence we will get a false audit trail, which will make it harder to understand when it's safe to remove

Dockerfile.selfhost
flopper123: It's just for the build itself. Last time I ran it, the source map generation took up more memory, causing an OOM.
But I'll remove it from here as it's outside the scope

Dockerfile.selfhost
flopper123: Yes, I'll remove it from here

Dockerfile.selfhost
Woldbye: Interesting.. my build seems to succeed locally without it last I tried. Though I am not denying that this is an issue 

You don't necessarily have to lift it out, but a small comment would be nice. My main concern, is that extra memory might not be available on all cis

flopper123: Yeah, I think the code from this + the frontend work caused the memory usage to tip over.

And this memory is not used on others systems. This is only for when we build the image on our Github workflow 😄

Lindharden (approved):

PR #982, opened 2026-06-24
feat(i18n): add Japanese (ja) language support

### LINKED ISSUE #842: Add Japanese as a supported language

mahirhir:
## Problem

Closes #842. Japanese was not available as a UI language.

## Change

Registers `ja` as a supported language:
- `dashboard/src/constants/i18n.ts` — adds `ja` to `SUPPORTED_LANGUAGES`, its metadata (name 日本語, code `JP`, ogLocale `ja_JP`), and the `date-fns/locale` `ja` import.
- `dashboard/messages/ja.json` — new locale file.

This is a **first pass** covering the dashboard application UI: navigation, metrics, filters, time ranges, the pages / referrers / devices / geography / events views, onboarding, banners, validation messages and toasts. Marketing/legal pages (`public.*`), pricing, monitoring, integrations and member/invitation flows are intentionally left in English for now and will follow up in subsequent PRs to keep this review manageable.

Translations follow standard Japanese analytics conventions (直帰率 = bounce rate, 参照元/流入元 = referrers, セッション, ユニーク訪問者, …). Product and technical names are kept in English (Next.js, React, npm, UTM, Cookie, GDPR, SSL, Web Vitals). All ICU placeholders (`{count, plural, ...}`, `{domain}`) and inline tags are preserved verbatim.

## Testing

- `ja.json` parses as valid JSON; key structure is identical to `en.json` for the translated sections (0 missing / extra keys).
- No new dependencies — the `date-fns` `ja` locale is already bundled.
- ICU argument names, plural categories and tags verified against `en.json`.

### DIFF
--- a/dashboard/messages/ja.json
+++ b/dashboard/messages/ja.json
@@ -1,0 +1,3913 @@
+{
+  "public": {
+    "root": {
+      "seo": {
+        "title": "Betterlytics | Privacy-First Web Analytics",
+        "description": "Betterlytics is a privacy-first, open-source alternative to Google Analytics. Get GDPR-compliant, cookieless insights without invasive tracking, while respecting user privacy.",
+        "keywords": [
+          "Google Analytics Alternative",
+          "Privacy-First Analytics",
+          "Cookieless Web Analytics",
+          "GDPR Compliant Analytics",
+          "Open Source Analytics Platform",
+          "Privacy-Friendly Analytics",
+          "Website Traffic Analysis",
+          "Visitor Analytics",
+          "Web Analytics Tool",
+          "Betterlytics"
+        ]
+      }
+    },
+    "notFound": {
+      "title": "404 - Page Not Found",
+      "description": "This page seems to have wandered off into the analytics dimension. Don't worry though, we've got you covered!",
+      "goHome": "Back to home"
+    },
+    "about": {
+      "seo": {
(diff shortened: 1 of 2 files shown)


Woldbye:
Thank you for your contribution! :D
 
The translations I tested seemed high quality, and I think it's fair to postpone parts of the translation to a follow-up.
 
That said, I noticed a problem with the Japanese glyphs that we'll need to take care of before we can merge. It's some existing tech debt in our formatters rather than anything in your PR:
 
**The problem:** 
a lot of our display strings (filter chips, table cells, page paths, chart labels) are trimmed to fit by counting characters, which quietly assumes every character is Latin-width. Japanese glyphs are roughly twice as wide, so instead of getting trimmed they spill out of their containers and break the layout.

To be clear, none of this is on you - it's existing tech debt in our formatters that your PR is simply the first to surface, since Japanese is our first full-width language. We'll handle the formatter fix in a separate PR and land the two together, so there's nothing you need to change here in the meantime. 

Thanks again for kicking this off! 🙏

---

@flopper123 @Lindharden Would love to get your input on how to resolve the formatter issue, I tried a few things last night, but couldnt find a proper solution (see #983)

Issue #984, opened 2026-07-05
Bare wildcard (*) filter values should mean value present / absent


Woldbye:
### Background

A bare `*` filter value (transformed to `%`) currently degenerates on columns that can be empty: `custom_event_name = *` and `utm_source = *` are silent no-ops because `'' ILIKE '%'` matches.

### Proposal

Define bare `*` as "value present" (`=`) / "value absent" (`!=`) consistently:

- All columns that can be empty (`custom_event_name`, `outbound_link_url`, `utm_*`, `referrer_*`, `city`, ...), mirroring how global-property filters already treat `gp.key = *` as "key exists".
- Both grains: `ba-query.ts` (events, 14 repositories + MCP query builder) and `ba-session-query.ts` (sessions).

### Performance note

For the session bridge, gating `custom_event_name` / `outbound_link_url` on their `event_type` should be applied to the `=` bridge at the same time - it is blocked today only because it changes `= *` behaviour, which this issue defines. The guard lets the bridge subquery prune on the `(site_id, event_type, toDate(timestamp))` primary key instead of scanning all event types in range.

### Scope warning

This is a deliberate behaviour change for saved filters/funnels/reports containing `*` values and probably warrants a changelog entry.

PR #985, opened 2026-07-05
Not equals filters on event-scoped columns now exclude sessions on session pages

### LINKED ISSUE #973: Not-equals (!=) filters on event-scoped columns have no effect on session pages

Woldbye:
Closing #973

Fixes != filters on custom_event_name, url, and outbound_link_url being silent no-ops on session-grain pages.

The correct session-level semantics for != is "the session contains no matching event", consistent with = behaviour, with event-grain pages, and with how global-property filters already negate (NOT arrayExists over all_props).

## Key decisions

  - != filters get their own session_id NOT IN (...) bridge over the positive match; = filters keep the existing IN bridge unchanged. Multiple != filters are OR'd inside a single subquery (one events scan, equivalent to ANDed exclusions).
  - The negation predicates for custom_event_name / outbound_link_url are guarded by their event_type ('custom' / 'outbound_link'). This is semantically free (only those event types carry the column), lets the subquery prune on the (site_id, event_type, toDate(timestamp)) primary key (EXPLAIN indexes=1 confirms the key condition includes event_type), and makes all-wildcard patterns read sensibly: custom_event_name != * = "sessions with no custom events".
  - Broader bare-* semantics (= * as "value present", all columns, both grains) deliberately deferred to #984.

## Additional Notes

  - Performance on large sites improves: the broken != materialized an IN-set of nearly all session_ids in range on every query; the fixed NOT IN set contains only matching sessions, and the event_type guard prunes the subquery scan to the relevant event-type.


Lindharden (approved):
Awesome, well done 👏 

Measured on the dev site: the guarded subquery reads 136/774 granules instead of 774/774, and the materialized session-ID set drops from 3.55M to 508k even for the site's most common event. So it's faster and lighter while also being correct 😎

PR #987, opened 2026-07-12
Bare wildcard (*) filter values now mean value present or absent

### LINKED ISSUE #984: Bare wildcard (*) filter values should mean value present / absent

Woldbye:
## Bare wildcard (*) filter values now mean value present or absent

Filtering with a bare `*` now does something useful: `column = *` keeps only rows where the column has a value, and `column != *` keeps only rows where it does not.

## Description of why

Closes #984. Until now a bare `*` was a silent no-op on columns that can be empty, because an empty string happily matches `ILIKE '%'`. Filters like `custom_event_name = *` or `utm_source != *` looked like they worked but changed nothing. Global-property filters already treat `gp.key = *` as "key exists", so this brings regular columns in line with them.

The change lands in all three query builders - event grain, session grain, and the hourly materialized-view fast paths - so a page gives the same answer no matter which path serves it.

## Changes

- A filter whose only value is `*` (or a literal `%`, which is indistinguishable after the wildcard transform) now compiles to `col != ''` or `col = ''` instead of an ILIKE pattern. `event_type` is exempt because it is an Enum8, and comparing an Enum8 to an empty string throws.
- On session pages, equals filters on event columns now get the same event-type guards the not-equals side got in #985, letting ClickHouse prune on the primary key. The bridge also builds one subquery per event type, so combining `Event = *` with `Outbound link = *` means "the session did both" rather than matching nothing.
- Added 24 unit tests across the three builders, including a fast/slow-path parity check.

## Additional Notes

- Stacked on #985 - the base branch is the 973 one; retarget to main once that merges.
- This deliberately changes what saved filters containing `*` return, so it deserves a line in the next release changelog.
- `referrer_source = *` and `!= *` still do nothing, since that column is never empty (it defaults to `direct`). Whether `!= *` should mean "direct traffic" is an open product question, left out of scope here.

### DIFF
--- a/dashboard/src/lib/ba-session-query.ts
+++ b/dashboard/src/lib/ba-session-query.ts
@@ -10,7 +10,12 @@ import { GranularityRangeValues } from '@/utils/granularityRanges';
 import { z } from 'zod';
 import { safeSql, SQL, type SQLTaggedExpression } from './safe-sql';
 import { DateTimeString } from '@/types/dates';
-import { filterColumnSql, filterColumnSqlForSession } from './filter-sql';
+import {
+  filterColumnSql,
+  filterColumnSqlForSession,
+  isMatchAnyValueFilter,
+  matchAnyValueFilterSql,
+} from './filter-sql';
 
 // Filters
 const MAIN_TABLE_FILTERS: TableFilterColumn[] = [
@@ -287,8 +292,7 @@ function buildGlobalPropertyFilterQuery(filter: GpFilter) {
   const key = SQL.String({ [`gp_key_${filterHash}`]: filter.gpKey });
   const isEquals = filter.rawOperator === '=';
 
-  const isMatchAnyValue = filter.values.length === 1 && filter.values[0] === '%';
-  if (isMatchAnyValue) {
+  if (isMatchAnyValueFilter(filter.values)) {
     return isEquals
       ? safeSql`arrayExists(t -> t.1 = ${key}, all_props)`
       : safeSql`NOT arrayExists(t -> t.1 = ${key}, all_props)`;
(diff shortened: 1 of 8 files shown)


dashboard/src/lib/ba-session-query.ts
```diff
@@ -269,6 +301,11 @@ function buildGlobalPropertyFilterQuery(filter: GpFilter) {
 
 function buildSessionFilterQuery(filter: StandardFilter) {
   const column = filterColumnSqlForSession(filter.col, SESSION_TUPLE_COLUMNS);
+  const isMatchAnyValue = filter.values.length === 1 && filter.values[0] === '%';
+  if (isMatchAnyValue) {
+    return filter.rawOperator === '=' ? safeSql`${column} != ''` : safeSql`${column} = ''`;
+  }
```
flopper123: Suggestion: This seems to be done quite a few places, and sometimes it seems to guard for `col !== 'event_type'` and other times there's also an `filter.values[0] === '*'`. Should there maybe be an sql utils for this, to ensure consistency? 😄

Lindharden (changes requested):
Well done, super nice.

Could you add a single clause somewhere in the docs, otherwise this won't ever be discovered by users 😄

Woldbye:
@Lindharden @flopper123 Ready for your rereview whenever you have the time.

I did as you suggested: 
1. Extracted the SQL logic into `filter-sql.ts` to reusable utilities
2. Added a line to `filtering.mdx` about the match pattern

Lindharden (approved):
Awesome 👏

PR #988, opened 2026-07-12
Added referrer domain click filtering to Traffic Sources

Woldbye:
makes Traffic Sources referrer parent rows clickable to filter, matching Devices and Global Properties, via a shared row-annotation pattern across all overview sections.

## Description of why
- Parent referrer rows showed a pointer cursor and a 'Filter by ...' tooltip but silently did nothing; child URLs and every other overview section already apply filters on click
- No schema change needed: ingest already stores `referrer_source_name` as the referrer's root domain - exactly the value the rollup displays as the group label

## Key decisions
- All five overview sections converge on one pattern: rows declare `filterColumn`/`filterValue` and the new `useProgressTableFilterClick` hook derives click handling + interactivity from the annotation, so a row can no longer look clickable while doing nothing
- Rollup SQL deliberately untouched: on date ranges reaching before March 2026 (PR #813 changed what `referrer_source_name` stores) the filter undercounts older events - the same accepted limitation the Referrers page already has; a backfill was considered, but I rejected as too risky for little benefit
- Known limitation: on ranges spanning the #813 deploy, parent rows mix legacy parser names and root domains in `referrer_source_name`; the click filters on the topK-majority value, so it can partially match the row's own traffic - superseded by #1005 (dedicated `referrer_domain` column)

## Changes
- `ProgressBarData` gains optional `filterColumn`/`filterValue`; new shared `useProgressTableFilterClick` hook
- Traffic Sources: referrer parents filter `referrer_source_name`, children keep `referrer_url`, channels keep `referrer_source`
- Devices, Events/Global Properties, Geography and Pages migrated to row annotations; five bespoke `onItemClick` switch handlers deleted
- Update our simulate-events script, so it actually add traffic sources

## Demo
msedge_yUjH5SMh7Y

### DIFF
--- a/dashboard/src/repositories/clickhouse/referrers.repository.ts
+++ b/dashboard/src/repositories/clickhouse/referrers.repository.ts
@@ -176,6 +176,7 @@ export async function getReferrerUrlRollup(
       SELECT
         cutToFirstSignificantSubdomain(concat('http://', referrer_url)) as source_name,
         referrer_url,
+        referrer_source_name,
         session_id,
         _sample_factor
       FROM analytics.events ${sample}
@@ -198,6 +199,7 @@ export async function getReferrerUrlRollup(
       source_name,
       referrer_url,
       uniq(session_id) * any(_sample_factor) as visitors,
+      any(nullIf(referrer_source_name, '')) as stored_source_name,
       grouping(referrer_url) as is_rollup
     FROM enriched
     WHERE source_name IN (SELECT source_name FROM top_parents)
--- a/dashboard/src/repositories/clickhouse/referrers.repository.ts
+++ b/dashboard/src/repositories/clickhouse/referrers.repository.ts
@@ -199,7 +199,8 @@ export async function getReferrerUrlRollup(
       source_name,
       referrer_url,
       uniq(session_id) * any(_sample_factor) as visitors,
-      any(nullIf(referrer_source_name, '')) as stored_source_name,
+      /* pre-#813 events store legacy parser names here, not root domains; topK picks the majority */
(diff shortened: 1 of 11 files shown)


dashboard/src/repositories/clickhouse/referrers.repository.ts
```diff
@@ -198,6 +199,7 @@ export async function getReferrerUrlRollup(
       source_name,
       referrer_url,
       uniq(session_id) * any(_sample_factor) as visitors,
+      any(nullIf(referrer_source_name, '')) as stored_source_name,
```
Lindharden: I think we can make this deterministic as follows:
`arrayElement(topK(1)(nullIf(referrer_source_name, '')), 1)`

This should make it stable across queries.

Lindharden (changes requested):
Very nice, well done 👏 

Before merge, could you add a short comment at the any(nullIf(referrer_source_name, '')) line in getReferrerUrlRollup noting: for events ingested before #813, referrer_source_name holds legacy parser names rather than root domains, so on date ranges spanning that deploy this picks one of the mixed values nondeterministically and the parent-row filter can partially match its own row. Known limitation, superseded by #1005  (dedicated referrer_domain column). A line in the PR description would help too

dashboard/src/repositories/clickhouse/referrers.repository.ts
Woldbye: Agree nice suggestion :)

Woldbye:
@Lindharden I have done as you suggested, please have a look again when you can find the time 👍

Lindharden (approved):
Nice 👏

PR #993, opened 2026-07-15
Fail fast when an outdated selfhost checkout drives a newer image

Woldbye:
Makes the selfhost upgrade path fail safely and stops `:latest` from delivering images that outdated checkouts cannot boot.

## Description of why
- Since v1.3.5, three contract changes (deleted post-migrate scripts, required `INTEGRATION_ENCRYPTION_KEY`, required `SALTS_DATABASE_URL`) mean an old `betterlytics-selfhost` checkout driving a newer image crash-loops — and the old entrypoint runs **all database migrations first**, so the user discovers the problem only after their ClickHouse events table has been rewritten, with no rollback to the old version and a cryptic module-not-found error.
- The selfhost compose file pins a version tag going forward, so `:latest` no longer has a job — freezing it means legacy `docker compose pull` becomes a harmless no-op instead of a breaking upgrade.

## Changes
- `publish.yml`: stop applying the `latest` tag to selfhost images (`flavor: latest=false`).
- `scripts/run-migration.js`: preflight — `SECRET_BASE` set (selfhost-only signal) with `SALTS_DATABASE_URL` missing (only derived by an updated selfhost entrypoint) → print a clear "run git pull" message and exit **before any migration runs**.
- Re-add the two deleted post-migrate script names as error stubs with the same message.

## Additional Notes
- Verified end-to-end: built this image and ran it against a genuine v1.3.5 selfhost checkout — clean actionable error, zero migrations executed, databases untouched.
- Cloud and dev are unaffected: neither sets `SECRET_BASE`, so the preflight never fires there. The stubs are dead code except for legacy selfhost entrypoints.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -27,9 +27,7 @@ jobs:
         uses: docker/metadata-action@v5
         with:
           images: ghcr.io/${{ github.repository_owner }}/betterlytics-selfhost
-          # Do not move :latest — the selfhost compose file pins a version tag,
-          # and legacy checkouts pulling :latest must not receive images their
-          # mounted config cannot boot.
+          # Never move :latest - legacy checkouts pull it but can't boot newer images
           flavor: |
             latest=false
 
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -27,6 +27,11 @@ jobs:
         uses: docker/metadata-action@v5
         with:
           images: ghcr.io/${{ github.repository_owner }}/betterlytics-selfhost
+          # Do not move :latest — the selfhost compose file pins a version tag,
+          # and legacy checkouts pulling :latest must not receive images their
+          # mounted config cannot boot.
+          flavor: |
+            latest=false
 
       - name: Building selfhost image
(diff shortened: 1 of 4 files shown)


.github/workflows/publish.yml
```diff
@@ -27,6 +27,9 @@ jobs:
         uses: docker/metadata-action@v5
         with:
           images: ghcr.io/${{ github.repository_owner }}/betterlytics-selfhost
+          # Never move :latest - legacy checkouts pull it but can't boot newer images
+          flavor: |
+            latest=false
```
flopper123, line 32: Question: Is this not enough for fixing the issue? 😄

Woldbye: This does fix the default path (`docker compose pull` on an old checkout becomes a no-op), but not the manual one. If someone bumps the pinned tag without `git pull`, he will get the new image with old config. Without the migration changes, it would just die midway with cryptic errors against their DB. The changes turn that into a clear "run git pull, no migrations have been run" instead.

flopper123 (approved):
Nice one! 👏

Lindharden (approved):

PR #997, opened 2026-07-15
Added mentions of public status page to public pages

Lindharden:
Adds public status page mentions across public pages: landing, features, pricing comparison, and vs pages, plus SEO/README. Docs and changelog follow in separate PRs (#998).


Woldbye (commented):
Overall looks good, but found a few things worth a second look.

1. You have stated the "own domain" claim unconditionally in a few places, but the statusPageCustomDomain is Professional+ in your own comparison table (from this PR). The landing card correctly hedge it in English ("public status pages you can host on your own domain"), but the features.items.statusPages just asserts it ("hosted on your own domain"). And then da/it/nb drop the hedge on the landing card as well. The result is a Growth visitor on the Danish landing page gets told they have a custom domain. I suggest we pick one phrasing and use it in all four locales? "you can host on your own domain" works for me. Just to clarify: Only the domain part is a problem, "branded" is fine to keep. Its only the Betterlytics badge and the domain that are gated.

2. I think status pages might have been forgotten on the Pricing Cards. This is a very nice feature that we should illustrate on the cards! It has a clear 1/3/unlimited ladder plus the Pro-only domain and while-label unlocks. Seems like the kind of thing we should be selling! :D Wdyt?

3. There is no PR description and no linked issue, so I have no idea what your intended scope is with this PR. I suggest adding a description before its merged for our own audit :)

4. I assume updating /docs is outside scope here, but yea it definintly needs an update before we release the feature.

5. Last one: I'd suggest merging the feature before the public pages. If this lands first, main is claiming a feature that isn't there, so any deploy in the meantime ships invalid descriptions. You never know what bugs the feature runs into :)

flopper123 (dismissed):

flopper123 (approved):

PR #998, opened 2026-07-16
Public status page docs

Lindharden:


Woldbye (dismissed):
LGTM

The only thing I would consider is to add an explicit message in the docs denoting that its not supported for self-hosters.

flopper123 (approved):

PR #1001, opened 2026-07-19
Move docs zod override into the Dockerfile

Woldbye:
#994 placed `pnpm.overrides` in `docs/package.json`, but pnpm ignores member-level overrides in a workspace and warns about it on every install. The override is only needed in the standalone docs image build (no lockfile in that context, so deps resolve fresh) — workspace installs pin zod 4.3.2 via the root lockfile.

## Changes
- Inject scoped `nextra>zod` / `nextra-theme-docs>zod` overrides in `docs/Dockerfile` (`npm pkg set` before `pnpm install`) and remove the override from `docs/package.json` — no more workspace warning, and docs' own direct zod stays on v3 in the image, matching dev.
- Add the same scoped overrides in `pnpm-workspace.yaml` so a future lockfile regeneration can't re-resolve nextra to a broken zod 4.4.x. Scoped on purpose: a plain `zod` override would force the dashboard from zod 3 to 4.

## Verification
- `pnpm install` in the workspace: warning gone; lockfile records the scoped overrides with no resolution changes (+2 lines).
- Cold `docker build` of `docs/`: passes, all pages prerender; inside the image docs' own zod resolves 3.25.x while nextra's is pinned 4.3.2.

### DIFF
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -13,6 +13,8 @@ FROM base as builder
 WORKDIR /app
 
 COPY package*.json .
+# zod >= 4.4 breaks nextra's MDX prerender.
+RUN npm pkg set pnpm.overrides.zod=4.3.2
 RUN pnpm install
 
 COPY . .
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -14,7 +14,7 @@ WORKDIR /app
 
 COPY package*.json .
 # zod >= 4.4 breaks nextra's MDX prerender.
-RUN npm pkg set pnpm.overrides.zod=4.3.2
+RUN npm pkg set "pnpm.overrides[nextra>zod]=4.3.2" "pnpm.overrides[nextra-theme-docs>zod]=4.3.2"
 RUN pnpm install
 
 COPY . .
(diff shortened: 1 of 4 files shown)


Lindharden (dismissed):
LGTM, nice cleanup.

Verified on my end: re-ran `pnpm install --lockfile-only` on the branch. 0 lockfile diff and the member-override warning is gone. Also confirmed both nextra and nextra-theme-docs depend on zod directly in the lockfile, which matters since the parent>child selector only matches direct deps, so the scoped overrides will actually bite if a lockfile regen ever tries to pull zod 4.4.x. Scoping instead of a root-level zod override is the right call given the dashboard is still on zod 3.

One non-blocking thought: the image and workspace now resolve docs' own zod differently: docs depends on zod ^3.25.63 directly, and the global override in the Dockerfile bumps that to 4.3.2 in the image while dev keeps zod 3. It's pretty much harmless for us today, since we only realy use a few zod functions, which is identical across the two major zod versions, but thereoretically it could pass local builds while fail in the image build. If you want exact parity you could scope the Dockerfile override the same way:

RUN npm pkg set "pnpm.overrides[nextra>zod]=4.3.2" "pnpm.overrides[nextra-theme-docs>zod]=4.3.2"

Approved, regardless of what you decide 😄

Woldbye:
@Lindharden I did as you suggested but your review got stale in the process :(

Lindharden (approved):
Awesome 👏

flopper123 (approved):

PR #1007, opened 2026-07-24
Added global properties to MCP schema definition

### LINKED ISSUE #1006: Expose global properties as a filter type in the MCP server
## Background

Betterlytics supports **global properties** (custom event properties) as a filter type: the filter schema, entity layer, and SQL builder all understand `gp.`-prefixed columns. `FilterColumnSchema` accepts `gp.<key>` (`dashboard/src/entities/analytics/filter.entities.ts`), `parseFilterColumn` narrows `gp.` columns into `{ kind: 'gp', key }` for SQL, and the MCP query builder delegates filtering to the shared `BAQuery.getFilterQuery` (`dashboard/src/mcp/query-builder/builder.ts:50`), which already handles the `gp` kind. `getFilterColumnDescription` even has a branch for `gp.` columns (`dashboard/src/mcp/registry/filterColumns.ts:27`). So today an LLM *can* technically pass a `{ column: "gp.plan", operator: "=", values: ["pro"] }` filter over the MCP and it works.

The gap is **discovery**. The `describe` tool only enumerates the static `FILTER_COLUMNS` array (`dashboard/src/mcp/tools/describe.ts:29`) and never advertises which global-property keys exist for the current dashboard. An LLM has no way to *discover* that e.g. `gp.plan` or `gp.user_type` are filterable — it only sees the 20 standard columns, so global-property filtering is effectively invisible over the MCP.

The dashboard filter UI already solved discovery: it lists a dashboard's available global-property keys via `getTopGlobalPropertyKeys` in `dashboard/src/repositories/clickhouse/globalProperties.repository.ts`. The MCP should surface the same thing.

Note: the query / journey / funnel tools already accept and execute `gp.` filters — this is purely about **discoverability** in `describe`, so no changes are needed there.

## Proposal

1. **Surface available global properties in the `describe` output**, so an LLM can discover them instead of having to already know a key exists.
   - List the current dashboard's top global-property keys, each labeled so the model knows to use the `gp.<key>` form.
   - *Decision:* nest them under the existing `filterColumns` list, or expose a dedicated `globalProperties` section in the schema output.
2. **Reuse the existing discovery repository** rather than writing a new query.
   - `getTopGlobalPropertyKeys` (`dashboard/src/repositories/clickhouse/globalProperties.repository.ts`) is the same source the dashboard filter UI uses.
   - *Decision:* pick a sensible key limit + time window for discovery, mirroring whatever the dashboard filter UI already uses.
3. **Give `getSchemaDescription()` site context and make it async.** Global-property keys are per-dashboard and time-scoped, so discovery needs a site id + a time window; today `getSchemaDescription()` is a pure static function (`dashboard/src/mcp/tools/describe.ts`).
4. *(Optional)* **Surface example values per key** via `getTopGlobalPropertyValuesForKeys`, at the cost of a second ClickHouse query per `describe` call.

## Acceptance criteria

- [ ] The `describe` output lists the current dashboard's available global-property keys, each shown in the `gp.<key>` form an LLM can filter by
- [ ] An MCP client can take a global-property key from `describe` and successfully filter a `query` by it end-to-end, with no prior knowledge of the key
- [ ] Standard filter columns and the query / journey / funnel tools are unchanged in behavior

flopper123:
Closes #1006

### DIFF
--- a/dashboard/src/mcp/server.ts
+++ b/dashboard/src/mcp/server.ts
@@ -91,7 +91,7 @@ export function createMcpServer(context: McpContext): McpServer {
     'list_global_properties',
     {
       description:
-        'List the custom event (global) properties recorded for this dashboard. Without arguments it returns the property keys, each as a column in the gp.<key> form you can use directly in a filter, e.g. { column: "gp.plan", operator: "=", values: ["pro"] }. Pass a "key" argument to see example values for that specific property. Use this to discover which custom properties are filterable.',
+        'List the custom event (global) properties recorded for this dashboard. Without a "key" argument it returns the property keys, each as a column in the gp.<key> form you can use directly in a filter, e.g. { column: "gp.plan", operator: "=", values: ["pro"] }. Pass a "key" argument to see example values for that specific property. Use this to discover which custom properties are filterable.',
       inputSchema: McpListGlobalPropertiesInputBaseSchema.shape,
     },
     (params) => runTool('list_global_properties', () => executeListGlobalProperties(params, context.siteId)),
--- a/dashboard/src/mcp/server.ts
+++ b/dashboard/src/mcp/server.ts
@@ -6,6 +6,10 @@ import { executeUserJourneys, McpUserJourneysInputBaseSchema } from '@/mcp/tools
 import { executeFunnelPreview, McpFunnelPreviewInputBaseSchema } from '@/mcp/tools/funnelPreview';
 import { executeListFunnels, McpListFunnelsInputBaseSchema } from '@/mcp/tools/listFunnels';
 import {
+  executeListGlobalProperties,
+  McpListGlobalPropertiesInputBaseSchema,
+} from '@/mcp/tools/globalProperties';
+import {
   executeListErrors,
   McpListErrorsInputBaseSchema,
   executeGetError,
@@ -81,6 +85,16 @@ export function createMcpServer(context: McpContext): McpServer {
       inputSchema: McpFunnelPreviewInputBaseSchema.shape,
     },
     (params) => runTool('funnel_preview', () => executeFunnelPreview(params, context.siteId)),
+  );
+
+  server.registerTool(
+    'list_global_properties',
+    {
+      description:
+        'List the custom event (global) properties recorded for this dashboard. Without arguments it returns the property keys, each as a column in the gp.<key> form you can use directly in a filter, e.g. { column: "gp.plan", operator: "=", values: ["pro"] }. Pass a "key" argument to see example values for that specific property. Use this to discover which custom properties are filterable.',
+      inputSchema: McpListGlobalPropertiesInputBaseSchema.shape,
+    },
+    (params) => runTool('list_global_properties', () => executeListGlobalProperties(params, context.siteId)),
   );
 
   server.registerTool(
--- a/dashboard/src/mcp/tools/describe.ts
+++ b/dashboard/src/mcp/tools/describe.ts
@@ -219,7 +219,7 @@ export function getSchemaDescription(): SchemaDescription {
             type: 'string',
             required: false,
             description:
-              'A global-property key (with or without the "gp." prefix). If given, returns example values for that key instead of the key list.',
+              'A property key as returned by this tool, e.g. "gp.plan". If given, returns example values for that key instead of the key list.',
           },
         ],
       },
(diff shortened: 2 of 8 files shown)


Woldbye (dismissed):
Nice work! Seems to me that it closes #1006, so might wanna link the issue 👍

dashboard/src/mcp/tools/globalProperties.ts
```diff
@@ -0,0 +1,64 @@
...
+    userJourney: { numberOfSteps: 1, numberOfJourneys: 1 },
+  };
+
+  if (input.key) {
+    const bareKey = input.key.startsWith(GP_PREFIX) ? input.key.slice(GP_PREFIX.length) : input.key;
+    const values = await getTopGlobalPropertyValuesForKeys(siteQuery, [bareKey], GP_VALUE_LIMIT);
+    return {
+      column: `${GP_PREFIX}${bareKey}`,
+      values: values.map((v) => v.value),
+    };
```
Lindharden, line 67: The keys response below reports truncated, but this values response doesn't, even though it's capped at GP_VALUE_LIMIT = 20. An agent that gets exactly 20 values back can't tell "these are all the values" from "these are the top 20 of thousands". Suggest adding truncated: values.length === GP_VALUE_LIMIT here too.

dashboard/src/mcp/tools/globalProperties.ts
```diff
@@ -0,0 +1,64 @@
...
+// stay light on context. Values are the heavy part, so they're fetched per key on demand.
+const GP_KEY_LIMIT = 100;
+const GP_VALUE_LIMIT = 20;
+
+export const McpListGlobalPropertiesInputBaseSchema = McpDateRangeSchema.extend({
+  key: z
+    .string()
+    .optional()
+    .describe(
+      'A global-property key (with or without the "gp." prefix). If given, returns example values for that key instead of the key list.',
```
Lindharden: `key: "gp."` passes this schema and returns` { column: "gp.", values: [] }`, a column that `FilterColumnSchema `will then reject if the agent uses it in a filter. Suggest validating the bare key (after prefix-stripping) against the same `GP_KEY_PATTERN `from `filter.entities.t`s, or at minimum` .min(1)`, so the tool errors clearly instead of handing back an unusable column.

Lindharden: Since the tool returns keys already in gp. form, the agent never needs to know both forms are accepted. Suggest: 'A property key as returned by this tool, e.g. "gp.plan". If given, returns example values for that key instead of the key list.' 

and keep the prefix-stripping in the executor as silent tolerance.

dashboard/src/mcp/tools/describe.ts
```diff
@@ -197,6 +202,27 @@ export function getSchemaDescription(): SchemaDescription {
...
+            required: true,
+            description: 'Time range preset. Valid values are listed in the timeRanges field of this schema.',
+          },
+          { name: 'timezone', type: 'string', required: false, description: 'IANA time zone. Defaults to UTC.' },
+          {
+            name: 'key',
+            type: 'string',
+            required: false,
+            description:
+              'A global-property key (with or without the "gp." prefix). If given, returns example values for that key instead of the key list.',
```
Lindharden: Since the tool returns keys already in gp. form, the agent never needs to know both forms are accepted. Suggest: 'A property key as returned by this tool, e.g. "gp.plan". If given, returns example values for that key instead of the key list.' 

and keep the prefix-stripping in the executor as silent tolerance.

dashboard/src/mcp/server.ts
```diff
@@ -83,6 +87,16 @@ export function createMcpServer(context: McpContext): McpServer {
     (params) => runTool('funnel_preview', () => executeFunnelPreview(params, context.siteId)),
   );
 
+  server.registerTool(
+    'list_global_properties',
+    {
+      description:
+        'List the custom event (global) properties recorded for this dashboard. Without arguments it returns the property keys, each as a column in the gp.<key> form you can use directly in a filter, e.g. { column: "gp.plan", operator: "=", values: ["pro"] }. Pass a "key" argument to see example values for that specific property. Use this to discover which custom properties are filterable.',
```
Lindharden: timeRange is required, so "Without arguments" will mislead an agent into calling this with no args and hitting a validation error. Suggest "Without a "key" argument it returns the property keys…" to match the `describe.ts` wording.

Lindharden (changes requested):
Put a few comments. Remember to update tests accordingly as you address my comments 😄

Lindharden (approved):

PR #1017, opened 2026-07-26
Made the event log live: stable cursor pagination plus polled new events with a badge

### LINKED ISSUE #773: Real-time events are based on selected range
Here I have filter for Feb 11., which was 7 days ago:

Image

 This should only show real-time events, instead of being affected by the selected time range

flopper123:
The event log now paginates with a timestamp cursor instead of the dashboard time range and offset, so loaded pages never shift as new events arrive and the query key stops churning every minute. A 30s poll fetches everything newer than the top row and prepends it live with a brief highlight and a clickable "N new events" badge; if the gap overflows the poll limit (500), a fresh first page is swapped in without a loading spinner. The footer count is now filter-aware but time-range independent, fetched once and bumped locally by the poll. The card visuals were also tidied (shared ScrollArea, edge shadows while scrolled, header alignment).

Closes betterlytics/betterlytics#773

Reviewer notes:
- Event rows have no unique id and timestamps are second-precision, so the cursor is (timestamp, skip) with a deterministic ORDER BY tiebreak. `computeNextEventLogCursor` carries the skip across pages that end on the same second and has unit tests.
- `recentEvents` and `totalEventCount` moved from `analyticsProcedure` to `dashboardProcedure` with explicit query filters, since they no longer depend on the time range.
- Prepends compensate scroll position manually (native overflow anchoring is disabled), the poll guards against overlapping ticks, failures, and stale responses, and rows' relative timestamps refresh whenever the list changes.
- Translation changes are just two new keys (`newEvents`, `endOfLog`) replacing `noEvents` in all four locales.


dashboard/src/app/(app)/(protected)/dashboard/[dashboardId]/(dashboard)/events/EventLog.tsx
```diff
@@ -122,59 +226,135 @@ export function EventLog({ pageSize = DEFAULT_PAGE_SIZE }: EventLogProps) {
...
+                {!hasNextPage && allEvents.length > 0 && (
+                  <div className='text-muted-foreground border-border/60 border-t py-6 text-center text-xs'>
+                    {t('endOfLog')}
+                  </div>
+                )}
+              </>
+            )}
+          </ScrollArea>
+          {/* Smaller counterpart of the top scrim, marking the footer edge. */}
+          {allEvents.length > 0 && (
```
Comment (Woldbye) [inline dashboard/src/app/(app)/(protected)/dashboard/[dashboardId]/(dashboard)/events/EventLog.tsx]: The scrim shows whenever the list is non-empty, instead of when the list overflows. The top-scrim is scroll-state driven, this one should probably be too.

`computeScrollState` in ba-scroll-container.tsx already gives you `canScrollDown`. I'd suggest exporting it, and reusing it here, to keep code consistent. It also saves you a lot of trouble by solving some weird edge cases such as scrollTop under browser zoom.

dashboard/src/app/(app)/(protected)/dashboard/[dashboardId]/(dashboard)/events/EventLog.tsx
```diff
@@ -122,59 +226,135 @@ export function EventLog({ pageSize = DEFAULT_PAGE_SIZE }: EventLogProps) {
...
+                type='button'
+                tabIndex={newEventsBadge.visible ? undefined : -1}
+                onClick={() => {
+                  const el = scrollRef.current;
+                  if (el) {
+                    // Cap the smooth-scroll distance so the glide stays short and consistent
+                    // from any depth; beyond that, jump instantly first.
+                    const glideDistance = el.clientHeight * 8;
+                    if (el.scrollTop > glideDistance) el.scrollTop = glideDistance;
+                    el.scrollTo({ top: 0, behavior: 'smooth' });
```
Comment (Woldbye) [inline dashboard/src/app/(app)/(protected)/dashboard/[dashboardId]/(dashboard)/events/EventLog.tsx:298]: Quibble: I tested this locally with ~300 rows loaded. The glide felt very sluggish - the 8x viewport cap is still a lot of smooth scrolling over a big DOM. I'd make it instant (scrollTo({ top: 0 })) or cap at ~2 viewports.

dashboard/src/app/(app)/(protected)/dashboard/[dashboardId]/(dashboard)/events/EventLog.tsx
```diff
@@ -47,58 +55,154 @@ const LoadingMoreIndicator = ({ t }: { t: EventLogTranslation }) => (
...
 
-  const allEvents: EventLogEntry[] = useMemo(
-    () => data?.pages.flatMap((page: EventLogEntry[]) => page) ?? [],
-    [data],
+  const allEvents: EventLogEntry[] = useMemo(() => data?.pages.flatMap((page) => page.events) ?? [], [data]);
+
+  // Rows never re-render on their own (stable identity + memo), so their relative
+  // timestamps freeze. Refresh them whenever the list changes — prepends, appends,
+  // resets — which is exactly when mixed stale/fresh labels would show.
+  const now = useMemo(() => Date.now(), [data]);
```
Comment (Woldbye) [inline dashboard/src/app/(app)/(protected)/dashboard/[dashboardId]/(dashboard)/events/EventLog.tsx:89]:
```suggestion
  // eslint-disable-next-line react-hooks/exhaustive-deps
  const now = useMemo(() => Date.now(), [data]);
```

Quibble: Like you do in EventLogItem

dashboard/src/repositories/clickhouse/events.repository.ts
```diff
@@ -122,42 +128,60 @@ export async function getRecentEvents(
...
+      device_type,
+      browser
     FROM analytics.events
-    WHERE site_id IN ({site_ids:Array(String)})
-      AND timestamp > now() - INTERVAL {within_days:UInt32} DAY
-    LIMIT 1
+    WHERE
+          site_id = {site_id:String}
+      AND event_type = 'custom'
+      AND timestamp > {since:DateTime}
```
Woldbye: DateTime uses second precision. So timestamp > since permanently drops events stamped in the same second.  The cursor solved this by (timestamp, skip). Seems like we should mirror it here?

dashboard/src/entities/analytics/events.entities.ts
```diff
@@ -45,9 +45,39 @@ export const EventLogEntrySchema = z.object({
   browser: z.string(),
 });
 
+export const EventLogCursorSchema = z.object({
+  timestamp: z.date(),
+  skip: z.number().int().min(0),
```
Woldbye: Skip is unbounded. I fear this leaves us vulnerable to malicious users DDOSing. Something like max(10_000) should solve it?

dashboard/src/app/(app)/(protected)/dashboard/[dashboardId]/(dashboard)/events/EventLog.tsx
```diff
@@ -47,58 +55,154 @@ const LoadingMoreIndicator = ({ t }: { t: EventLogTranslation }) => (
...
+        setNewUids(new Set(fresh.map(getUid)));
+        utils.events.recentEvents.setInfiniteData({ ...input, limit: pageSize }, (old) => {
+          if (!old) return old;
+          const [first, ...rest] = old.pages;
+          return {
+            ...old,
+            pages: [{ ...first, events: [...fresh, ...first.events] }, ...rest],
+          };
+        });
+        setNewEventsBadge({ count: fresh.length, visible: true });
```
Comment (Woldbye) [inline dashboard/src/app/(app)/(protected)/dashboard/[dashboardId]/(dashboard)/events/EventLog.tsx]: If I'm scrolled down, and dont click the pop up, then the next poll overwrites the count with just its own batch (3 new events, then 2 new events, instead of 5). How about accumulating while the prv badge hasnt been "noticed"? Something like  `setNewEventsBadge((b) => ({ count: fresh.length + (b?.visible ? b.count : 0), visible: true }))`

Woldbye (commented):
Nice feature ❤️  I really like the UX of the "infinite-scroll" with hints when new events arive and the styling of the radix scrollbar. Would it be possible to write it into a reusable wrapper we can use for other infinite scroll pages (LiveFeedScrollArea feels like a good name) - its definitely something we want on other pages.

Additional comments:
- There are some very long narrative code comments, consider if you can write them a bit more concise. 

- scroll-area is autogenerated component. I'd suggest writing a wrapper that uses the primitives instead of editing the /ui file

dashboard/src/app/(app)/(protected)/dashboard/[dashboardId]/(dashboard)/events/EventLog.tsx
```diff
@@ -47,58 +55,154 @@ const LoadingMoreIndicator = ({ t }: { t: EventLogTranslation }) => (
...
   });
 
-  const allEvents: EventLogEntry[] = useMemo(
-    () => data?.pages.flatMap((page: EventLogEntry[]) => page) ?? [],
-    [data],
+  const allEvents: EventLogEntry[] = useMemo(() => data?.pages.flatMap((page) => page.events) ?? [], [data]);
+
+  // Rows never re-render on their own (stable identity + memo), so their relative
+  // timestamps freeze. Refresh them whenever the list changes — prepends, appends,
+  // resets — which is exactly when mixed stale/fresh labels would show.
```
Comment (Woldbye) [inline dashboard/src/app/(app)/(protected)/dashboard/[dashboardId]/(dashboard)/events/EventLog.tsx]:
```suggestion
  // Memoized rows freeze their relative timestamps; recompute whenever the list changes.
```
Quibble

dashboard/src/app/(app)/(protected)/dashboard/[dashboardId]/(dashboard)/events/EventLog.tsx
Comment (flopper123) [reply dashboard/src/app/(app)/(protected)/dashboard/[dashboardId]/(dashboard)/events/EventLog.tsx:298]: Weird, I tested it with +2k rows loaded 🤔 Did notice that if I had Dev Console open that it would hurt performance greatly, like with the number roll thingy. Could it maybe be caused by that?
Else I'll change it of course 😄

dashboard/src/repositories/clickhouse/events.repository.ts
flopper123: Nice catch! 😄 Reusing the cursor/skip will not necessarily work as expected here, since new rows interleave into the sort order. But to fix the bug, I did add `timestamp >= {since:DateTime}` and deduplication by content count 🫡

dashboard/src/entities/analytics/events.entities.ts
flopper123: Insane DDOS angle 😂 But I've added the max 😄

dashboard/src/app/(app)/(protected)/dashboard/[dashboardId]/(dashboard)/events/EventLog.tsx
Comment (flopper123) [reply dashboard/src/app/(app)/(protected)/dashboard/[dashboardId]/(dashboard)/events/EventLog.tsx]: I'm very open to doing this, I have a question though about if the the badge stay open until you click/scroll up if we do this? 🤔
Because it may be confusing that it pops up every 30s, but continuesly grows. Let me know! 😄

dashboard/src/repositories/clickhouse/events.repository.ts
Woldbye: I agree with your take. Nice pushback, and confirmed this works as intended.

dashboard/src/app/(app)/(protected)/dashboard/[dashboardId]/(dashboard)/events/EventLog.tsx
Comment (Woldbye) [reply dashboard/src/app/(app)/(protected)/dashboard/[dashboardId]/(dashboard)/events/EventLog.tsx]: Feel free to do as we discussed - either:
A. Leave exactly as is
B. Change it to show accumulation, but also change it so its not on a fixed timer, but shows until acknowledged. 

I personally prefer UX of B, as it better encapsulates the use-case where users leave the tab open in the browser, but A is definintly also acceptable :)

You can resolve this either way, whichever direction you decide.

dashboard/src/app/(app)/(protected)/dashboard/[dashboardId]/(dashboard)/events/EventLog.tsx
Comment (Woldbye) [reply dashboard/src/app/(app)/(protected)/dashboard/[dashboardId]/(dashboard)/events/EventLog.tsx:298]: It was indeed caused by dev-tools, though this means its a likely edge case on old hardware not as performance resillient.

I'd personally recommend not smooth scrolling either way:  a polling prepend landing mid-scroll-glide assigns `scrollTop` in the effect, so I think there's a rare edge case which cancels an in-progress smooth scroll and strands you partway up. I couldnt reproduce this myself, but from the code it should be possible

Either way :) Existing is fine for me if you prefer the smooth scrolling

Woldbye (dismissed):

Lindharden (approved):
LGTM 👍

Issue #1003, opened 2026-07-22
City filter click merges same-named cities across regions


Woldbye:
Clicking a city row in the Geography section (and applying a city filter generally) filters on `city` alone, but city rows are grouped by `(city, subdivision_code, country_code)`. Two distinct rows can share a name - e.g. Springfield (US has dozens) - and clicking either one filters the dashboard to the union of every same-named city worldwide, which does not match the clicked row.

Side effect: same-named city rows also collide on their React list key in MultiProgressTable (key is the label), sharing expand/hover identity.

Pre-existing behavior, surfaced during review of #988 (the row-annotation migration preserved it unchanged).

Possible direction: rows already carry `subdivision_code`/`country_code`, so a city click could apply a compound filter (city + country_code, or city + subdivision_code). Requires the click path to support applying more than one filter at once.

PR #1018, opened 2026-07-26
Add browser and OS version filter columns

### LINKED ISSUE #1004: Browser/OS version rows say "Filter by Chrome 120" but filter by browser only
In the Devices section, expanding a browser/OS shows version child rows (e.g. "Chrome 120"). Their tooltip reads "Filter by Chrome 120", but clicking applies `browser = Chrome` - the version is dropped, because no `browser_version`/`os_version` filter column exists. Same for OS versions. Confirmed pre-existing on main

Possible fixes:
- Add `browser_version`/`os_version` to the filter schema (the ClickHouse columns already exist via the device-versions migration) and annotate version rows with a real version filter.
- Or make the row tooltip reflect the applied filter value instead of the label. Note a naive swap breaks Global Properties parent rows, whose filter value is the `*` wildcard while the label is the property key - the tooltip would become "Filter by *".

Woldbye:
Closes #1004.

Clicking a version row in the Devices section now applies browser + version as two filters atomically, backed by new `browser_version` / `os_version` filter columns exposed everywhere filters are (dropdown with autocomplete, funnels, MCP, all four locales). The ClickHouse columns already existed; the backend stores major versions only.

To support this, `MultiProgressTable` rows now declare their filters as a data-driven `filters` list (any number of column/value pairs per row), which also makes #1003's merged-city rows a clean follow-up.

Also extends `simulate-events.js` with weighted browser/OS/device profiles so seeded data contains version variety.


Lindharden (changes requested):
I have one minor request. But other than that, it looks super good!

When you have applied a browser + version filter, then clicking the parent rollup row "Chrome" reads as "filter to Chrome, any version." But in reality, this keeps the browser_version as is, which contradicts what the row you clicked really represents.

Woldbye:
> When you have applied a browser + version filter, then clicking the parent rollup row "Chrome" reads as "filter to Chrome, any version." But in reality, this keeps the browser_version as is, which contradicts what the row you clicked really represents.

@Lindharden agreed - a parent row click now applies the browser and resets any version filter, so its truthful. I also think this  is better UX

Lindharden (approved):
LGTM! 👍

flopper123 (approved):
Awesome! 😄

PR #1019, opened 2026-07-27
Fixed monitor check intervals above 1h failing to save

Lindharden:
The monitor form offered check intervals up to 24h while the schema capped them at 1h, so any multi-hour selection failed with a generic error. The interval range (60s to 24h) now lives in `MONITOR_LIMITS`, and both the Zod schema and the slider marks derive from it so they cannot drift apart again. Backend backoff no longer snaps base intervals to a hardcoded ladder that mirrored the dashboard's values: the effective interval is now base x 2^level capped at 24h, meaning a 44m monitor actually probes every 44m (the ladder silently rounded it to 1h), and future slider changes need no backend edit.

Closes betterlytics/betterlytics-internal#4

**Reviewer notes**

- Migration 36 widens `monitor_results.effective_interval_seconds` to UInt32 (24h overflows UInt16). The usual initializer-before-backend deploy order covers it, but an old backend writing into the widened column errors, so deploy both together.
- Also bumps the enterprise plan minimum interval from 30s to 60s: the schema floor was always 60, so 30s was dead config that could never be selected or saved.
- First unit tests in `backend/src/monitor` cover the new backoff policy.

**Verification**

On a live worktree: monitors at 60s/3m/5m/30m/24h all saved, first probes recorded the exact configured interval, and observed cadence matched within the 10% scheduling jitter (60s monitor: 58s/58s/64s gaps; 3m monitor: 195s). Hourly uptime charts show "no data" for hours a multi-hour monitor does not check in; that is inherent to infrequent checking, not a regression.

### DIFF
--- a/backend/src/monitor/backoff.rs
+++ b/backend/src/monitor/backoff.rs
@@ -9,8 +9,6 @@ const DEFAULT_MAX_EFFECTIVE_SECS: u64 = 86_400;
 const DEFAULT_FAILURE_THRESHOLD: u16 = 25;
 const DEFAULT_SUCCESS_THRESHOLD: u16 = 2;
 
-/// Escalation is defined relative to the monitor's own base interval (base × multiplier^level,
-/// capped), so the policy is independent of whichever interval values the dashboard offers.
 #[derive(Clone, Copy, Debug)]
 pub struct BackoffPolicy {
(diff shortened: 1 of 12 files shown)


flopper123 (approved):
Very nice 😄

PR #1020, opened 2026-07-28
Resolved all pnpm and cargo audit vulnerabilities

Lindharden:
Bumped every dependency flagged by pnpm audit (dashboard and docs) and cargo audit (backend) that could move without breaking changes, one dependency per commit. Advisories that cannot be fixed in place are now explicitly ignored with rationale comments (auditConfig in pnpm-workspace.yaml, backend/.cargo/audit.toml), so both audits exit 0 and can later serve as CI gates. Also bumped pnpm 10.9.0 to 10.34.5, since 10.9 silently ignores auditConfig in pnpm-workspace.yaml.

Reviewer notes:
- Nearly the entire diff is the two lockfiles. The real review surface is the package.json files, pnpm-workspace.yaml (new overrides and audit ignores), backend/.cargo/audit.toml, and the pnpm version in both Dockerfiles.
- Three one-line, type-level code changes were forced by the bumps: Next 15.5 route-param validation (share layout locale param) and newer Buffer typings against Prisma Bytes (status page images, now Uint8Array).
- Accepted advisories, each documented at the ignore site: nodemailer raw option (we never pass raw), sharp (pinned by next's own ^0.34 range), brace-expansion DoS (dev-only eslint tooling), maxminddb open_mmap (API unused), rustls-webpki 0.101 (unused legacy rustls 0.21 connector under aws-smithy).
- Dependabot alerts are computed against main, so merging auto-closes all but the accepted ones above; those need manual dismissal in the UI.

Verified with dashboard and docs production builds, the 239 dashboard unit tests, a backend build with its 51 tests, and clean runs of both audits after the final commit.


dashboard/src/app/(app)/[locale]/share/[dashboardId]/layout.tsx
```diff
@@ -42,7 +42,7 @@ export async function generateMetadata({
 }
 
 type PublicLayoutProps = {
-  params: Promise<{ locale: SupportedLanguages; dashboardId: string }>;
+  params: Promise<{ locale: string; dashboardId: string }>;
```
Comment (Woldbye) [inline dashboard/src/app/(app)/[locale]/share/[dashboardId]/layout.tsx:45]: What vulnerability is this change for specifially?

Comment (Lindharden) [reply dashboard/src/app/(app)/[locale]/share/[dashboardId]/layout.tsx:45]: Fix for the next 15.4 -> 15.5 bump. Next 15.5 started validating route export types at build time, and a layout's params must now be assignable from what Next actually passes, where locale is a plain string. Narrowing it to SupportedLanguages in the props type fails the build on 15.5

flopper123 (approved):

Woldbye (approved):

PR #1021, opened 2026-07-29
Fixed active filter pills overflowing the viewport on long filter values

### LINKED ISSUE #885: Active query filter pills overflow on mobile
Very long active query filters overflow on mobile 

image

Lindharden:
Long filter values made active filter pills wider than the screen, on mobile via the badge's unbounded `w-fit whitespace-nowrap` sizing and on desktop via the `lg:` header flex row stretching to the pill's unwrapped max-content width. Pills are now capped at their container width and the values text truncates with an ellipsis, with `min-w-0` at each nested flex level so the truncation actually engages.

Closes betterlytics/betterlytics#885
Closes betterlytics/betterlytics-internal#13

**Reviewer notes:** `FilterDescription` is shared with the funnel bar tooltip, where very long values now stay on one line instead of wrapping (column labels already behaved this way there).

**Verification:** reproduced the issue's filter (long URL value lists) against a live worktree at 390px and 1440px: no horizontal overflow at either width, and the column label, ellipsized values, and remove button all stay visible.

### DIFF
--- a/dashboard/src/components/dashboard/DashboardFilters.tsx
+++ b/dashboard/src/components/dashboard/DashboardFilters.tsx
@@ -20,7 +20,7 @@ export default function DashboardFilters({
   const isMobile = useIsMobile();
 
   return (
-    <div className='space-y-2'>
+    <div className='min-w-0 space-y-2'>
       <div className='flex flex-col justify-end gap-x-8 gap-y-2 sm:flex-row'>
         {!isMobile && <div className='flex gap-4'>{children}</div>}
         {showQueryFilters && <QueryFiltersSelector />}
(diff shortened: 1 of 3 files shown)


Woldbye (approved):
Approved but not sure you should close the issues

I'm fine with merging this, as it's an improvement from current state. But i dont Think it properly closes either issue, it just adds a bandaid. 

The mid elipsis string format is still cooked, and is the real issue. 

I don't see how this solves the row overflowing either: if you have 10 filters, it will still look like shit on mobile.

Also worth noting that the width tested on according to pr description, is not the smallest realistic mobile device

Lindharden:
> Approved but not sure you should close the issues
> 
> I'm fine with merging this, as it's an improvement from current state. But i dont Think it properly closes either issue, it just adds a bandaid.
> 
> The mid elipsis string format is still cooked, and is the real issue.
> 
> I don't see how this solves the row overflowing either: if you have 10 filters, it will still look like shit on mobile.
> 
> Also worth noting that the width tested on according to pr description, is not the smallest realistic mobile device

Discussed in person. Merging

PR #1022, opened 2026-07-29
Apply the full geography hierarchy on row filter clicks

### LINKED ISSUE #1003: City filter click merges same-named cities across regions

Woldbye:
Closes #1003

City rows in the Geography section group by (city, region, country), but clicking one applied only `city = <name>` - merging same-named cities across regions into one filter broader than the row clicked. Same convention gap #1018 fixed for browser/OS versions.

## Changes
- Geography rows now declare their full hierarchy: a city click applies city + region + country, a region click applies region + country.
- `country_code` and `subdivision_code` added to `DEPENDENT_FILTER_COLUMNS`, so clicking a broader row clears the narrower filters (a country click resets region and city pills).
- Cities without a GeoIP subdivision skip the region filter; a stale region pill is still cleared via the country's dependent expansion.
- Row tooltips and keys use the full hierarchy too: two Springfields now read "Filter by Springfield, Illinois, United States" vs "..., Massachusetts, ..." instead of identical tooltips, and no longer collide as React keys.

## Additional Notes
- The world map only pins tooltips and applies no filters, so it needs no changes; the overview table is the only geo filter-click surface.


Lindharden (approved):
LGTM 👍

PR #1023, opened 2026-07-30
Scope filter value suggestions by dependency-related filters

### LINKED ISSUE #851: Dashboard Filter value options don't respect the previous active filters
When multiple filters are active, the value dropdown for each filter shows all possible values for that column - ignoring the other active filters.

For example, if the first filter is country_code = US (narrowing to 5% of traffic), adding a second filter for browser still shows all browsers across all countries, not just browsers used by US visitors.

This means users see options that would return zero results when combined with their existing filters. It also makes it harder to explore data progressively, as the filters feel disconnected

---

## Example

Image
- The options of the region filter does not consider the previous URL filter.

## Suggested change
the value options for filter N should be scoped to events that already match filters 1 through N-1.

Woldbye:
Value suggestions in the filter builder ignored sibling rows, so dependent columns offered contradictions: Safari versions next to a `browser = Chrome` row, Houston under `country = Denmark`, and inversely a country dropdown offering Germany while `city = Aarhus` guarantees zero results. Suggestions for a column are now scoped by usable sibling draft filters on its dependency-related columns, in both directions.

## Changes
- `dependencyScopeFilters` in the filter entities reads `DEPENDENT_FILTER_COLUMNS` both ways (parents and dependents); applied client-side for tight cache keys and re-derived server-side so the rule is enforced regardless of client input.
- `filters.getFilterOptions` accepts optional `scopeFilters`, applied to the DISTINCT query through the existing filter SQL builder.
- Draft rows thread down as `siblingFilters` from the filter selector and the funnel step editor; unrelated columns and `gp.*` never cross-scope, and typed values remain unrestricted (`creatable`).
- The search hook resets its client-vs-server search decision when the scope changes structurally, so a parent value change refetches even after the dropdown switched to client-side filtering.

## Additional Notes
- Deliberately narrower than the closed #851 suggestion (scope by all prior filters): only dependency-related columns scope each other, so orthogonal lists stay complete and edits cannot dead-end.


Lindharden (changes requested):
Mechanics are clean, but I'd drop the upward direction. The issue is:

Rows `country_code = DK` + `city = Aarhus` -> open the country dropdown to switch to
Germany and it offers only Denmark. `updateQueryFilter` replaces one row by id and doesn't clear dependents, so you have to delete the city row first. Same for `browser` under `browser_version`.

Downward scoping carries the real value anyway as a bare `browser_version` list is
useless without the browser. I suggest we scope by dependents only (drop the parents lookup + the `scopes upward` test), and clear dependent rows when a parent's value changes so the builder matches the row-click semantics.

Two small ones:
- `id` is in the tRPC input but unused server-side, and temp ids churn which splits the
  query cache. Send `{column, operator, values}`.
- I believe "can never empty the suggestion list" isn't true because `browser != *` compiles to
  `browser = ''`, and two rows on the same column AND to nothing

Woldbye:
Parking this in draft for now, let's discuss this when there's more time to focus on the behavior

Woldbye:
@Lindharden   Adjusted after your round - the upward direction is gone: `dependencyScopeFilters` now looks up only the parent columns of the column being suggested, and the old scopes upward test asserts the opposite. That alone fixes your trap example: the country dropdown is never scoped anymore, so switching to Germany doesn't require deleting the city row first - the stale city = Aarhus row then surfaces through the city list itself (now scoped to German cities) instead of blocking the parent.

On clearing dependent rows when a parent's value changes: I'd rather not add another silent auto-clear to the builder. 
 
 The two small ones:

  - Adopted: ids are off the wire - `ScopeFilterSchema` omits id, the client sends `{column, operator, values}`, so temp-id churn no longer splits the query cache.
  - Conceded: you're right that the list can empty - browser != * compiles to browser = '', and same-column rows AND together. Empty lists are possible but truthful (browserless sessions genuinely have no browser_version), and they guide rather than gate since typed values stay unrestricted.

Lindharden (approved):
Looks good to me now 👏

flopper123 (approved):

flopper123:
I think it makes sense to align on this.

The one-directional scoping feels a little off to me 🤔 With a filter like `country = DK` and `city = Aarhus` hiding `Houston` removes a filter that's still valid. The filter `city IN (Aarhus, Houston)` and `country = DK` doesn't change the actual result. So it's guarding a no-op instead of an invalid filter.

I checked what the other sites do, 3 out of 4 don't scope values by filters at all, and only one scope all filters.
So having the one-directional for some filters seem like a novel approach.

So I think I'm leaning either towards scoping everything or nothing. Thoughts? 😄

PR #1024, opened 2026-07-30
Fixed filter value dropdown flashing its post-pick state during the close animation

Woldbye:
Picking a value in the filter value multi-select briefly flashed a different list state (highlight jump, scroll snap, reordering) while Radix played the popover's exit animation. Closing without picking was always smooth - the flash only appeared on close paths that also mutate state.

Resolves betterlytics/betterlytics-internal#84

## Key decisions
- Neither #945 pattern fits. Reset-on-open leaves stale search text in the always-visible anchor input, and deferring resets to animation-end (`useOverlayReset`) can't stop the selection-driven list changes, which must apply at click time.
- Freezing the rendered list during the exit was tried and reverted: cmdk holds its own external store, and the cleared input triggers select-first + `scrollIntoView` inside the still-mounted list - per-item store subscriptions bypass any React-level snapshot.
- Final fix: drop the closed-state animation classes on this popover so Radix unmounts it synchronously. Picking closes instantly, leaving no window to paint into; the open animation is kept.

## Changes
- `MultiSelect.tsx`: removed `data-[state=closed]` animation classes from the popover content (one line vs main).

### Before
msedge_3uWzoOB6KG

### After
msedge_OBkQlsNBZB

### DIFF
--- a/dashboard/src/components/MultiSelect.tsx
+++ b/dashboard/src/components/MultiSelect.tsx
@@ -457,63 +457,6 @@ export const MultiSelect = ({
     return undefined;
   }, [creatable, commandProps?.filter]);
 
-  const listContent = isLoading ? (
-    <>{loadingIndicator}</>
-  ) : (
-    <>
-      {EmptyItem()}
-      {!selectFirstItem && <CommandItem value='-' className='hidden' />}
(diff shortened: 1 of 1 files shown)


Lindharden (dismissed):
lgtm

Lindharden (approved):
lgtm

PR #1025, opened 2026-07-30
Added country flags to the country filter dropdown and applied filter chips

flopper123:
Country values across the filter UI now render their flag: the value dropdown options, the selected badges inside the input, and the applied filter chips (funnel filter labels too, via the shared `FilterDescription`). Values stay as ISO codes, so search behavior is untouched. Unknown codes (e.g. "Localhost") fall back to FlagIcon's existing help icon, and hovering a flag shows the localized country name.

Closes betterlytics/betterlytics-internal#85

**Beyond the ticket:** flags in the applied filter chips were not in the issue (it only listed the dropdown, plus badges as optional), added for consistency.

**Reviewer notes:** `MultiSelect` is extended with an optional `renderOption` prop rather than rich option labels, since it round-trips options through `JSON.stringify` (React nodes would not survive). Also, `FlagIcon`'s light-mode shadow was softened (`shadow-foreground/80` to `/50`), which affects every flag surface app-wide, not just the filters.

Light mode:
image

Dark mode:
image


Lindharden (changes requested):
- Bare wildcard * values now render a confusing "?" icon.
- There's some almost-duplicate code between FilterDescription.tsx and FilterValueSearch.tsx, consider extracting to a resuable helper.

dashboard/src/components/filters/FilterDescription.tsx
```diff
@@ -25,7 +27,22 @@ export function FilterDescription({ filter, className }: FilterDescriptionProps)
     >
       <FilterColumnLabel column={filter.column} />
       <span className='text-muted-foreground/80'>{operator}</span>
-      <span>{values}</span>
+      {filter.column === 'country_code' ? (
+        filter.values.map((value, index) => (
+          <span key={value} className='inline-flex items-center gap-1'>
```
Woldbye: This is duplicated between `FilterValueSearch` and `FilterDescription` - the only real differences are the root className and the label. Could we extract a small shared component (e.g. `CountryFlagLabel({ countryCode, className, children })`) that also applies `toUpperCase()` + `getCountryName`? 

Maybe we could make `CountryDisplay` in components/language, (which does almost exactly this except for span instead of div) general enough to cover both

dashboard/src/components/filters/FilterDescription.tsx
```diff
@@ -25,7 +27,22 @@ export function FilterDescription({ filter, className }: FilterDescriptionProps)
...
+            <FlagIcon
+              countryCode={value.toUpperCase() as FlagIconProps['countryCode']}
+              countryName={getCountryName(value, locale)}
+            />
+            <span>
+              {strategy.formatValue(value, locale)}
```
Woldbye: Quibble: Both files now branch on `filter.column === 'country_code'`. I'll need custom value rendering for other columns in an upcoming task. WDYT of moving this into the filter column strategy instead, e.g. an optional `renderValue(value, locale)` next to `formatValue`, so they can both just call `strategy.renderValue?.(...) ?? formatValue(...)`

WDYT of this

Woldbye (dismissed):
Nice idea, looks crisp. 

I had two architectural suggestions but no blockers!!

dashboard/src/components/filters/FilterDescription.tsx
flopper123: Yes sir! 🫡

Woldbye (approved):

Lindharden (approved):

PR #1026, opened 2026-07-30
Added expiration to MCP tokens with a lifetime picker at creation

flopper123:
Token creation now takes a lifetime (30, 90, or 180 days, default 90, or no expiry). The client sends a lifetime enum and the server computes `expiresAt`, so no client-supplied dates. The token list shows when each token expires (or that it never does) and renders already-expired tokens in red. Pre-existing tokens keep `null` expiry and continue working; the auth path already rejected expired tokens, so nothing changes there.

Closes betterlytics/betterlytics-internal#57

**Reviewer notes**
- No migration: the `expiresAt` column already existed, only creation never wrote it.
- About a third of the diff is the new strings across the four locale files, plus a short note on the MCP docs page.


Lindharden (changes requested):
1. Warn before expiry. When a token is within ~7 days of expiring, append a notice to successful tool results, e.g. "expires in 3 days, create a replacement under Settings > MCP".

Once it expires we reject at the HTTP layer, before an MCP session exists. There's no tool result and no turn for a message to travel in, so the agent can't tell the user anything. From their side the integration just stops working and looks like our server broke. While the token still works is the only point where we can get a message through.

validateToken already loads expiresAt, so it can go through McpContext, and runTool is a single choke point for all eight tools.

2. The 401 doesn't say why it failed. A client can't tell "rotate your token" from "server is down", and -32001 is reused for missing header, invalid token, and expired token alike. Not sure what the right shape is. These might tell us what to do:

https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization
https://datatracker.ietf.org/doc/html/rfc6750#section-3
https://datatracker.ietf.org/doc/html/rfc9728#name-www-authenticate-response

My agent get the following response when using an expired token:

```
betterlytics:
  Scope: Local config (private to you in this project)
  Status: ✘ Failed to connect
  Type: http
  URL: http://mcp-improvements.localhost:3050/api/mcp
  Headers:
    Authorization: Bearer btl_ff91f5e04e8fa841f8a5dc57c5b52239

To remove this server, run: claude mcp remove "betterlytics" -s local
```

Lindharden (approved):

PR #1027, opened 2026-07-31
Added global properties to betterlytics.io's own tracking (surface, logged_in, locale)

flopper123:
The betterlytics.io analytics script tags (app and docs site, one shared site id) now send demo-safe global properties via `data-global-properties`: `surface` (`app` / `docs`), `logged_in` (session cookie presence, no auth lookup) and `locale`. They ride the script tag attribute, so the initial pageview already carries them. Also bumps `@betterlytics/tracker` to 0.2.2.

Closes betterlytics/betterlytics-internal#77

The ticket also specified a client-side `theme` property. It was implemented but backed out for now: `setGlobalProperties` calls made before analytics.js finishes loading are silently dropped by the pre-load stub, so the value never reached most page loads. Queuing those calls is tracked in betterlytics/betterlytics-internal#102, and `theme` returns with it.

Reviewer notes: most of the diff is the pnpm lockfile from the tracker bump. The root layout was already per-request dynamic (`getLocale()` reads cookies), so the added session cookie read changes nothing about rendering.


dashboard/src/app/layout.tsx
```diff
@@ -25,7 +26,11 @@ export default async function RootLayout({
 }: Readonly<{
   children: React.ReactNode;
 }>) {
-  const [locale, seoConfig] = await Promise.all([getLocale(), buildSEOConfig(SEO_CONFIGS.root)]);
+  const [locale, seoConfig, sessionToken] = await Promise.all([
+    getLocale(),
+    buildSEOConfig(SEO_CONFIGS.root),
+    getCurrentSessionTokenFromCookies(),
```
Lindharden: Gate the cookie read on the tracking flag so self-host deploys don't have to needlessly pay for this.

```suggestion
    env.ENABLE_APP_TRACKING ? getCurrentSessionTokenFromCookies() : undefined,
```

Lindharden (changes requested):

Lindharden (dismissed):

Lindharden (approved):

PR #1028, opened 2026-07-31
Added full event list, search, and property value drilldown to the events page

Lindharden:
The events table now fetches the whole event set with its own limit (1000, overview card stays at top 10) and handles search, sorting, and pagination client-side so ordering is correct across the full set. Expanded rows get their property summaries from a single aggregated ClickHouse query instead of parsing up to 10k raw JSON rows in Node, making counts exact at any volume, and each property can now drill down to its full value list ("Show all N values", capped at 1000) on demand.

Closes betterlytics/betterlytics-internal#14

Spec 5 (optional cap indicator) is implemented for property value lists ("Showing top 1,000 of N values") but not for the event list itself, which caps silently at 1000 event names.

**Reviewer notes:** risk is concentrated in `events.repository.ts` (two new aggregate queries using window functions and LIMIT BY); the large `EventsTable.tsx` diff is mostly mechanical pagination/filter wiring, and the four locale files change identically. To see it: expand any event on /events, then "Show all N values" on a property.

Verified against a seeded worktree (130 event names, one event with 60 properties); counts cross-checked against ClickHouse directly.


flopper123 (changes requested):
Very nice! 😄 

Some UI/UX feedback:
1) You can only change page size if the initial page is above 10 unique events:

Here there are more than 10:
Image

Here there are less than 10:
Image

2) If you set the page to be above the max number of unique events, you can't set the page size down again:

Image

3) We show the number of unique events twice. Should one of them maybe be removed?

Image

4) `EventTableSkeletonRow` complains about duplicate keys:

Image

5) The button says "Show all x values", but when clicked it only shows max 1000 events.

Image

Image

6) (Nitpick) For phone layout, should this maybe be placed down here, or something like that:

Image

Lindharden:
> Very nice! 😄
> 
> Some UI/UX feedback:
> 
> 1. You can only change page size if the initial page is above 10 unique events:
> 2. If you set the page to be above the max number of unique events, you can't set the page size down again:

Good catch, fixed 👍

> 3. We show the number of unique events twice. Should one of them maybe be removed?

No. One shows the filtered count (based on your search input) and relates to pagination. The header count is the primary indicator what the card covers

> 4. `EventTableSkeletonRow` complains about duplicate keys:

Fixed 👍 

> 5. The button says "Show all x values", but when clicked it only shows max 1000 events.

This was a deliberate decision. When scrolling in a list where only 10 rows fit at once, scrolling past 1000 becomes insanely tedious and an uninteded user flow, thus this is a really rare case and it was made explicit exactly by the "showing x of z values". I've adjusted the text to be explicit about this, although I don't think it's neccessary.

On the other side, we don't want to ship uncapped amount of rows, as the payload will just increase linearly in size.

> 6. (Nitpick) For phone layout, should this maybe be placed down here, or something like that:

Unrelated to this PR and is the existing common paginationControls component that needs this change. I'll file an issue on this instead 😄 

https://github.com/betterlytics/betterlytics-internal/issues/121

flopper123 (dismissed):

flopper123 (approved):

PR #1034, opened 2026-08-03
Added the design system foundation: token layer and Badge primitive in components/v2

flopper123:
First vertical slice of the design system: a two-tier oklch token layer (raw primitives, semantic roles, and a Tailwind v4 bridge) plus a hand-written Badge primitive (intent x appearance, plus pill) living in `dashboard/src/components/v2/`. Components are flat files imported per file, shadcn-style (`@/components/v2/badge`), no barrels. Three exemplar areas migrated off the legacy shadcn badge: billing plan status, errors status badges, and the events unique-count chip.

Reviewer notes:
- The decisions worth scrutiny are in `tokens/`: primitives are theme-invariant oklch ramps, semantics map them to roles, and `.dark` redefines the complete semantic key set (even values that don't change) so editing a light token can never silently restyle dark mode.
- Migrated badges colors:
image

- Incidents and every other badge call site are untouched and stay on the legacy badge; they migrate in follow-ups.


dashboard/src/components/v2/badge.tsx
```diff
@@ -0,0 +1,114 @@
...
+      },
+      {
+        intent: 'warning',
+        appearance: 'soft',
+        className: 'border-intent-warning-border bg-intent-warning-surface text-intent-warning-content',
+      },
+      {
+        intent: 'warning',
+        appearance: 'solid',
+        className: 'border-transparent bg-intent-warning text-content-on-accent',
```
Woldbye: White text on amber-600 is a little bit hard to read on 12px. Might just be my screen or eyes, but seems like it may need a slightly sharper contrast. Maybe even black text.

dashboard/src/components/v2/tokens/semantic.css
```diff
@@ -0,0 +1,67 @@
...
+  --intent-success-surface: oklch(from var(--bl-green-500) l c h / 0.1);
+  --intent-success-border: oklch(from var(--bl-green-500) l c h / 0.3);
+  --intent-success-content: var(--bl-green-400);
+
+  --intent-warning: var(--bl-amber-600);
+  --intent-warning-surface: oklch(from var(--bl-amber-500) l c h / 0.1);
+  --intent-warning-border: oklch(from var(--bl-amber-500) l c h / 0.3);
+  --intent-warning-content: var(--bl-amber-400);
+
+  --intent-danger: oklch(from var(--bl-red-450) l c h / 0.6);
```
Woldbye: This is the only base solid token implemented as a translucent veil - every other base is opaque. If 60% alpha the backdrop contributes 40% of the rendered color, so "solid" danger will differ substantially between surfaces and the label's contrast shifts with it. It might be intended but it would make it easier to reason about to just use an opaque value here.
  
My thoughts in general on the -surface/-border veils, is that translucency is fine at small scale, but it doesnt scale to large areas (cards, sticky headers) over moving content. a translucent card can't be marked opaque by the compositor, so it pays blending on every frame content move. This exact pattern is something I have personally struggled with in the past on the FunnelDialogStep cards on drag, and on the CWV Gauges. Its actually documented in the existing code if you look at FunnelStepAccordionItem.tsx:285-286:
```
  'bg-card dark:bg-secondary/50',
  isDragging && 'dark:bg-secondary',
```

I think its a fine limitation, but its worth documenting, so we ensure large areas with moving content that needs bg opacity are instead flattened against the actual bg: `color-mix(in oklch, <primitive> 10%, var(--surface-base))` (Note that there currently is no surface-base to mix against, just an illustration of the idea).

dashboard/src/components/v2/tokens/theme.css
Woldbye, line 1: This file is basically a mirror of every semantic token. Note that I know that we use the same pattern today in our css-theme setup.

Though, now we have a chance to do it right: How about adding a small test to assert they exist so they dont drift silently?

Woldbye (dismissed):
Extremely lovely read. Im genuinely excited for this rework 🗡️ 
As is, its a real improvement. But since you asked for harsh feedback, I did find a few code comments, worth clearing up, but overall it seems good to me 👍

dashboard/src/components/v2/badge.tsx
flopper123: I've updated them 😄

dashboard/src/components/v2/tokens/semantic.css
flopper123: Noice, I've updated the design document and such 😄

dashboard/src/components/v2/tokens/theme.css
flopper123: I've opened an issue for this https://github.com/betterlytics/betterlytics-internal/issues/130 😄

Woldbye (approved):

PR #1036, opened 2026-08-04
Fixed unknown event names permanently killing event ingestion

Lindharden:
Any client could send a non-custom event with an unknown event name: row conversion panicked on it, killing the insert task, after which every event was silently dropped (still 200-acked, /health still green) until a container restart. Validation now rejects unknown non-custom event names with a 400, and row conversion drops the row with a warning instead of panicking as a second layer of defense.

Reviewer notes: reproduced live before the fix (a single curl stopped all ingestion, zero rows stored afterward). The whitelist is exactly the set of names that map to the ClickHouse event_type enum; custom events keep free-form names. Unit tests cover both layers.

### DIFF
--- a/backend/src/validation/mod.rs
+++ b/backend/src/validation/mod.rs
@@ -103,7 +103,15 @@ impl EventValidator {
         self.validate_required_fields(raw_event)?;
         self.validate_payload_sizes(raw_event)?;
         self.validate_formats(raw_event, ip_address)?;
-        
+
+        // Non-custom events must use a known event name; anything else has no
+        // ClickHouse event_type enum value and can never be stored.
+        if !raw_event.is_custom_event && !is_known_event_name(&raw_event.event_name) {
+            return Err(ValidationError::InvalidEventName(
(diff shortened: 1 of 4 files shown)


backend/src/validation/mod.rs
```diff
@@ -406,6 +414,15 @@ impl EventValidator {
     }
 }
 
+/// Event names for non-custom events that map to a ClickHouse event_type enum
+/// value ("cwv" is matched case-insensitively downstream).
+fn is_known_event_name(name: &str) -> bool {
+    matches!(
+        name,
+        "pageview" | "custom" | "outbound_link" | "scroll_depth" | "engagement" | "client_error"
+    ) || name.eq_ignore_ascii_case("cwv")
```
flopper123: What's the reason why we ignore the case only for `cwv`? 😄

Lindharden: Pre-existing convention. I have removed it from both the pre-existing function and here, so it matches the other events :)

flopper123 (approved):

PR #1040, opened 2026-08-04
Added a usage breakdown to billing settings

Lindharden:
Adds a usage breakdown to the Usage section in billing settings, showing what a user's billable events are made of by event type and by site. The headline usage bar is now segmented — the filled portion is divided by what consumed it, so there is still exactly one capacity on screen rather than a bar per category. Percentages are share of the plan limit rather than share of usage, which is what makes the segments decompose the headline figure exactly.

Closes betterlytics/betterlytics-internal#75

image

### Deviation from the ticket

The ticket specified adding a count column per billable event type to `usage_by_site_daily`. Instead, `analytics.usage_daily` carries `event_type` as a row dimension, with an explicit target table plus a `TO` view. Adding a billable type is then a view recreate plus a targeted backfill rather than a full rebuild, and both breakdown dimensions come from one grouped query. The four existing billing queries were repointed at the new table.

The daily accumulation trend from spec 1 is not included. The type and site lists answer "what is using my quota"; a trend answers a different question, and the projected total already covers "am I on track". Easy to add later if anyone asks for it.

### Reviewer notes

- Migration 37 drops `usage_by_site_daily`, creates `usage_daily` + `usage_daily_mv`, and backfills, so it wants the usual services-down window. Worth confirming nothing outside this repo (Grafana panels, ad hoc ops queries) still selects from the old view, since it is dropped with no compatibility alias.
- The breakdown loads with the page. An earlier revision hid it behind a disclosure and fetched lazily; now that the section is always visible, `useUsageBreakdown` lost its `enabled` argument and the query fires on mount.
- Sites beyond the top 5 collapse into one "Other" segment so the meter stays readable. The show-more toggle expands the full list in the legend only — the bar stays at six segments however long the tail.
- Segment colours are CSS custom properties (`--usage-segment-*`), a single-hue ramp stepped by size in both themes, so the bar reads as one meter rather than five categories. They are assigned by rank, not by category, so an event type changes colour if its ranking changes between periods.
- Segment tooltips only repeat what the legend already shows, so nothing is hover-only.
- The by-type / by-site tabs only appear for accounts with more than one site; single-site accounts get the type legend directly.
- Locale files are 84 of ~674 changed lines.
- Self-host is unaffected: the action returns an empty breakdown behind `enableBilling`, and the billing tab is already hidden there.
- The billable whitelist lives in the migration and is mirrored by `BILLABLE_EVENT_TYPES` for display labels. Comments on both sides name each other, but nothing enforces it.

Verified against a seeded account across 43 sites, both at 80% of plan and pushed over the limit at 106%: the breakdown total, and each dimension's percentages, sum exactly to the headline usage figure.


dashboard/src/app/globals.css
```diff
@@ -114,6 +114,13 @@
   --chart-comparison: oklch(0.7 0.02 265); /* Neutral gray */
   --chart-error: rgb(248, 113, 113); /* red-400 */
 
+  --usage-segment-1: oklch(58% 0.18 268);
+  --usage-segment-2: oklch(67% 0.15 264);
+  --usage-segment-3: oklch(76% 0.12 260);
+  --usage-segment-4: oklch(84% 0.09 256);
+  --usage-segment-5: oklch(90% 0.06 252);
+  --usage-segment-other: oklch(72% 0.02 265);
+
```
Woldbye, line 123: We already have this exact ramp for cwv as `--cwv-p50..p99`, and `--usage-segment-other` is `--chart-comparison` with 2% more light. How about we promote one shared sequential ramp (e.g. --seq-blue-1..5) that both CWV and usage consume, and reuse `--chart-comparison` for "other"?

If you dont wanna do it now, we should probably add an issue to backlog about streamlining our different color ramps throughout the app.

dashboard/src/components/billing/UsageRow.tsx
```diff
@@ -0,0 +1,66 @@
...
+    <div className='grid grid-cols-[minmax(0,1fr)_auto] items-center gap-x-4 gap-y-2 md:grid-cols-[minmax(0,1fr)_minmax(110px,20rem)_7.5rem]'>
+      {label}
+      {segments && segments.length > 0 ? (
+        <TooltipProvider delayDuration={150} disableHoverableContent>
+          <div
+            role='progressbar'
+            aria-valuemin={0}
+            aria-valuemax={100}
+            aria-valuenow={Math.min(percentageOfLimit, 100)}
+            className='bg-primary/20 order-last col-span-2 flex h-1.5 w-full overflow-hidden rounded-full md:order-none md:col-span-1'
```
Woldbye: Consider adding a bit more contrast to the empty bg color and lower end of the color spectrum. Its very hard to tell the difference if light is low on especially lightmode:

Image

Woldbye (dismissed):
Overall looks good - nice work. I had a few code suggestions to tighten it up a bit, but im fine merging it as is also :)

_One question_: 

What are your thoughts on this for self-hosters? Should we just always disable it, since they dont have billing? Technically we could still show the page without billing - the data layer is not billing coupled. That also introduces some questions about whether the code should be isBillingEnabled coupled at all.

flopper123 (changes requested):
Nice 😄 Code looks good, only had some UI/UX feedback:

1) The colors depend on the order and not the events themselves, should they have fixed colors per type?

Image

Image


2) If you look close at the progress bar, it seems as if it's like a C, zooming in seems to fix it as if it's a subpixel issue 🤔 

Image


3) Should we also show this the other places where we show event usage, like on `/billing`?

Image

dashboard/src/app/globals.css
Lindharden: I'll file an issue on this 👍

Lindharden:
> What are your thoughts on this for self-hosters? Should we just always disable it, since they dont have billing? Technically we could still show the page without billing - the data layer is not billing coupled. That also introduces some questions about whether the code should be isBillingEnabled coupled at all.

Disabled (and already is), billing shouldn't be accessible for self-hosters 😄 

> 1. The colors depend on the order and not the events themselves, should they have fixed colors per type?

I'd argue they shouldn't have fixed colors per type. The colors are closely related, so the user would have to remember that "medium blue is that" and "3/4 quarter of blue is this". Additionally, the current scale ensures that darkest = biggest, which is the intuitive design. That's what makes the bar read as a "meter". Coloring by type, means a light color blue can suddenly appear in between two darker colors, etc.

> 2. If you look close at the progress bar, it seems as if it's like a C, zooming in seems to fix it as if it's a subpixel issue 🤔

Interesting. I think I have fixed it, can you try to reproduce? 😄 

> 3. Should we also show this the other places where we show event usage, like on `/billing`?

No, this is meant for usage breakdown to understand where you can adjust events to "stay within limits" for instance. The billing page is a quick summary before you upgrade to understand what plan you need to choose. It's job on the billing page is basically: "here's your plan, here's the upgreade", a five-segment decomposition will make the billing page a diagnostic page instead.

flopper123 (approved):

PR #1041, opened 2026-08-04
Polished country flag icon rendering

flopper123:
FlagIcon no longer bakes in a drop shadow (too prominent in light mode). Its edge
definition now comes from a single overlay with per-side border tones, including
dark mode variants the old border lacked, plus a softer gloss gradient. The height
default switches from font-relative (!h-[1.15em]) to a plain h-4, and the component
now merges a caller-supplied className instead of silently discarding it, so call
sites can size flags themselves.

Reviewer notes:
- The className merge revives ErrorOccurrencePanel's previously inert h-6 w-6, so
  the error detail flag now renders 24px like the browser/OS/device icons next to it.
- With the height default no longer important, the active filter chips inherit
  FilterDescription's [&_svg]:size-3 (12px flag, was 13.8px). Everywhere else
  renders 16px, close to the old 1.15em at typical font sizes.

Before:
image

After:
image

Before:
image

After:
image

Dark mode:
image

image

### DIFF
--- a/dashboard/src/components/filters/CountryFlagLabel.tsx
+++ b/dashboard/src/components/filters/CountryFlagLabel.tsx
@@ -15,9 +15,13 @@ export function CountryFlagLabel({ countryCode, className, children }: CountryFl
   const code = countryCode.toUpperCase();
 
   return (
-    <span className={cn('inline-flex items-center gap-1.5', className)}>
+    <span className={cn('inline-flex items-center gap-1', className)}>
       {hasFlag(code) && (
-        <FlagIcon countryCode={code as FlagIconProps['countryCode']} countryName={getCountryName(countryCode, locale)} />
+        <FlagIcon
+          countryCode={code as FlagIconProps['countryCode']}
+          countryName={getCountryName(countryCode, locale)}
+          className=''
+        />
       )}
       {children}
     </span>
--- a/dashboard/src/components/filters/CountryFlagLabel.tsx
+++ b/dashboard/src/components/filters/CountryFlagLabel.tsx
@@ -15,12 +15,11 @@ export function CountryFlagLabel({ countryCode, className, children }: CountryFl
   const code = countryCode.toUpperCase();
 
   return (
-    <span className={cn('inline-flex items-center gap-1', className)}>
+    <span className={cn('inline-flex items-center gap-1.5', className)}>
       {hasFlag(code) && (
         <FlagIcon
           countryCode={code as FlagIconProps['countryCode']}
           countryName={getCountryName(countryCode, locale)}
-          className=''
         />
       )}
       {children}
--- a/dashboard/src/components/filters/CountryFlagLabel.tsx
+++ b/dashboard/src/components/filters/CountryFlagLabel.tsx
@@ -17,10 +17,7 @@ export function CountryFlagLabel({ countryCode, className, children }: CountryFl
   return (
     <span className={cn('inline-flex items-center gap-1.5', className)}>
       {hasFlag(code) && (
-        <FlagIcon
-          countryCode={code as FlagIconProps['countryCode']}
-          countryName={getCountryName(countryCode, locale)}
-        />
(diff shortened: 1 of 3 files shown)


dashboard/src/components/filters/CountryFlagLabel.tsx
```diff
@@ -15,9 +15,13 @@ export function CountryFlagLabel({ countryCode, className, children }: CountryFl
...
 
   return (
-    <span className={cn('inline-flex items-center gap-1.5', className)}>
+    <span className={cn('inline-flex items-center gap-1', className)}>
       {hasFlag(code) && (
-        <FlagIcon countryCode={code as FlagIconProps['countryCode']} countryName={getCountryName(countryCode, locale)} />
+        <FlagIcon
+          countryCode={code as FlagIconProps['countryCode']}
+          countryName={getCountryName(countryCode, locale)}
+          className=''
```
Woldbye: Whats the point of `className=''`?

dashboard/src/components/icons/FlagIcon.tsx
```diff
@@ -29,16 +30,15 @@ function FlagIconComponent({ countryCode, countryName, ...props }: FlagIconProps
     <span title={countryName || 'Unknown'} className='relative flex items-center justify-center'>
       <FlagComponent
         {...props}
-        className='shadow-foreground/50 dark:shadow-background/50 inline-block !h-[1.15em] rounded-xs shadow-sm dark:rounded-none'
```
Woldbye, line 32: This is actually a real regression on some pages. 

- The flags are now inconsistently sized on the `EventLogItem` 

- The flags are now oversized compared to the text labels on our landing page
Image

- The flag appears taller than the text across all surfaces I tested (aside from the active query filters). Its especially noticable on the filter dropdown:
Image


I did some research and it seems like the popular sizing solution is to make the flags 1em everywhere, so they always match the line height.

Woldbye (commented):
I can see you moved this to draft, while I was mid-review, but I already started my review, so I might as well complete it 👍 

I found a few inconsistencies all stemming from the new px specific height behavior instead of the preexisting em we need to look into.

dashboard/src/components/filters/CountryFlagLabel.tsx
flopper123: nothing, it's been removed 🫡

dashboard/src/components/icons/FlagIcon.tsx
flopper123: I've made it 1.1em 😄 That way the flag is slightly larger than the text, but not too

Woldbye: Agreed on 1.1em looks good boss! :D Im very happy with visuals now, will run a final verification round tonight before approving

dashboard/src/components/icons/FlagIcon.tsx
```diff
@@ -29,16 +30,16 @@ function FlagIconComponent({ countryCode, countryName, ...props }: FlagIconProps
...
+        className={cn('inline-block rounded-xs dark:rounded-none', className)}
         style={{
           imageRendering: 'auto',
           shapeRendering: 'geometricPrecision',
-          height: '1.15em',
+          height: '1.1em',
```
Woldbye, line 37: Quibble: How about removing the height declaration from style, and moving it into the className declaration. That way you dont have to do the "important" overwrites in consumers.

Woldbye (approved):
I like the sizes now - great work 👍 

Had a small optional suggestion that would clean this up for future consumers, but not required to merge it.

dashboard/src/components/icons/FlagIcon.tsx
flopper123: Sadly that wont work due to Shadcn adding svg size overrides. If we add it inline to className it'll fight for specificity, where it'll lose to shadcn 😒 
So this is the only way to gain consistent sizing sadly.

Lindharden (approved):

PR #1042, opened 2026-08-04
Make click-applied filters discoverable

Woldbye:
Clicking a row applies a filter silently; users neither notice it happened nor know where to remove it. All click paths go through `use-filter-click`, so feedback is added at that choke point. The Filters editor stays silent.

## Changes
- Single fixed-id toast on every click-applied change: "Filters updated" with green `+` / red `-` rows per actual diff, rendered by the same `FilterDescription` as the pills (icons and flags included); Undo restores the pre-click state, single level, 10s
- Clicking an already-applied filter is a no-op, and semantically unchanged filters keep their instance identity across replacements and undo, so pills never remount or re-pulse without a real change
- Filters trigger shows the active count as a primary badge instead of the "(N)" label
- Newly added pills get a brief entrance pulse (skipped on page load, reduced-motion aware)
- The sonner `Toaster` moved from the root layout into `(app)/layout.tsx`, inside the theme and intl providers, so toast content renders with app context (the 2025 reason for root mounting - a `/create` route outside the dashboard layout - no longer exists; every toast surface lives under `(app)`)

Filter-aware empty states were split out to betterlytics/betterlytics-internal#126.

Closes betterlytics/betterlytics-internal#47
Closes betterlytics/betterlytics-internal#127


Lindharden (changes requested):
Looks really nice! I found a few issues:

1. Undo restores a whole-state snapshot, not the announced change
The toast says "+ Chrome", but Undo does `state = prev`. Any change made through a non-click path while the toast is open (Filters editor Apply, saved-filter load, pill X, browser Back) is silently reverted too:
- `[]` -> click Chrome -> add `country=DK` in the editor -> Undo -> DK is gone as well.
- `[DK]` -> click Chrome -> X the DK pill -> Undo -> DK comes back.

I suggest applying the inverse diff (remove `added` if present, re-add `removed` if absent), or at least dismissing the toast from every non-click setter.

2. Fixed toast id has two sonner side effects

- `remainingTime` is only reset when `duration` changes, so after a hover pause the next click's toast inherits the leftover countdown (hover at 9s -> apply new filter -> the new toast disappears in ~1s).

I think we should look into unique id per toast + `toast.dismiss(previousId)`. This should also give the toast a new "appear animation", to more clearly indicate that it's a new action instead of overwriting the toast "silently".

3. There's some overflow issues on some states:

Image

I think it might make sense to simply make the toast more plain, either a simple
```Filter applied```

or

```
Filters applied                    [Undo]
Browser is Chrome, Browser version is 120
```

without showing the removed part, +/-, and/or icons

Woldbye:
@Lindharden 
1. Fixed, I did exactly as you suggested
2. Fixed by design by changing to dismiss + open new and unique ids per toast. The animation looks pretty slick I think (see GIF below)
3. I'd like to keep filter description, but I moved the UNDO button, and removed color on the +/- icons. I think it reads a bit more plain now.

msedge_WKay5TlF1b

4. I also added shrink-0 to operator and fixed overflow
image

Lindharden (approved):
Looks good to me now 👏

flopper123 (commented):
very nice 😄

PR #1043, opened 2026-08-06
Hardened ingest-time bot detection with shadow-first rule tiers

Lindharden:
Rebuilds ingest-time bot filtering around an evidence-first rule system: rules with ~zero false-positive risk reject events, everything else runs in shadow mode (recorded to the new `analytics.bot_events` table plus per-reason metrics, while the event still counts as human) until real data justifies promotion.

Closes betterlytics/betterlytics-internal#15

**Decisions:**

- Client timestamps are no longer trusted or validated: events are stamped with server receive time, and the 5-minute drift rejection is gone (it silently dropped real visitors with skewed clocks). Dev builds still honor payload timestamps so seeding tools can backdate.
- The UA blocklist is split into tiers: named product signatures enforce, format and generic-word heuristics shadow. New upstream patterns land in shadow by default, promotion is a reviewed one-line move, and the regen script verifies every tier against a human-browser corpus.
- Only three reasons reject: UA blocklist, referrer spam, and the client automation signal. Everything else is shadow-only: ASN rules (hosting and crawler-operator networks), prefetch, impossible resolutions, UA oddities, velocity, client hints, and stale browsers.
- The ticket's client-hints and stale-version signals ship in shadow mode: Chromium 89+ without a `sec-ch-ua` header (or with a contradicting major), and desktop Chromium majors more than ~3 years behind, both with thresholds to tighten from shadow data.
- The enforced blocklist also matches the User-Agent header, so a forged payload UA cannot mask an HTTP client.
- A per site+IP velocity window (120 events/min, IP anonymized like every derived signal) ships as shadow telemetry to learn the real rate distribution before any rate limiting.
- Consumer relay/CDN egress networks (Private Relay, Google One VPN) are excluded from the hosting-ASN list so shadow data is not dominated by real iOS and VPN users.
- `bot_events` keeps full history (no TTL) with ASN context; rows ride the same retrying insert pipeline as events, with per-table inserter metrics.
- Session replay ingestion is bot-gated (presign and finalize), and the tracker never loads replay for automated browsers. The tracker also flags headless Chrome via client-hints brands.
- A new `ClientRequest` extractor centralizes IP, User-Agent, sec-ch-ua, and prefetch extraction for all handlers.

Enforcement decisions for the shadow rules follow once bot_events data shows negligible false positives on known-good sites

**Reviewer notes:**

- ~15k of the ~16.5k added lines are vendored data files (bot patterns, referrer-spam domains, hosting ASNs). Start with `backend/src/bot_detection/`.
- Migration 38 creates `bot_events` and adds `asn` columns to `events`.
- Self-host: no new required config. Without MaxMind credentials the ASN shadow rules simply stay inactive.


backend/src/bot_detection/mod.rs
Woldbye, line 1: On line 241 you force the blocklist against the header UA - I fear this may have some unwarranted consequences, as it kills first party proxying,.

I suggest solving this by using your shadow tier to investigate if the forged post scenario is a real problem. How about we make header-UA-only hit shadow, keep payload-UA hits enforced, and promote once `bot_events` shows what actually arrives in that header?

**Note:** I literally cant comment on the exact line since the diff is too large (639 lines) is apparently more than github wants to show.

migrations/38_create_bot_events.sql
```diff
@@ -0,0 +1,20 @@
+-- No TTL: full bot-event history is the evidence base for tuning filter rules.
```
Woldbye, line 1: No TTL here + `apply_data_retention_policy` only touching `analytics.events` means this table is growing and unbounded forever, and it sits outside the retention window customers configure.

This becomes becomes a permanent store of real visitors user agent + URL + referrer that our privacy docs don't describe and our retention settings don't reach. Self-hosters get unbounded disk on top.

I'd like full bot-event history personally, but I fear the privacy concerns and we need to bound it for self-hosters either way (so it doesnt cause memory issues for them cause of this in the future).

backend/src/geoip_updater/mod.rs
```diff
@@ -30,43 +124,66 @@ pub struct GeoIpUpdater {
...
 
+        let enabled = config.geolocation_mode.is_enabled() && Self::has_credentials(&config);
+        let db_path = config.geoip_db_path.clone();
+        Self::with_database(config, database_url, db_path, enabled)
+    }
+
+    /// Creates an updater for the ASN database. Independent of geolocation mode:
+    /// ASN data drives bot detection, not geo reports.
+    pub fn new_asn(config: Arc<Config>) -> Result<(Self, GeoIpWatchRx)> {
+        let enabled = Self::has_credentials(&config);
```
Woldbye: This gates the asn download on credentials alone (independent of ENABLE_GEOLOCATION).

I think that would mean a self-hoster who sets `ENABLE_GEOLOCATION=false` but leaves `MAXMIND_LICENSE_KEY` in their .env (a very common case - people flip the flag, they don't delete the key) now downloads GeoLite2-ASN and does an `IP->ASN` lookup on every event, even persisting `asn` and `asn_org` for each row.

Our own docs say "When disabled, no geo data is collected" (see docs/src/content/installation/self-hosting.mdx). 

I think the right seam is its own flag (`ENABLE_ASN_LOOKUP`, default off) plus an update to docs.

backend/src/config.rs
```diff
@@ -120,6 +121,9 @@ impl Config {
                 } else {
                     PathBuf::from("assets/geoip/GeoLite2-Country.mmdb")
                 }),
+            asn_db_path: env::var("ASN_DB_PATH")
```
Woldbye, line 127: AsnService is constructed unconditionally, so every self hoster without MM creds gets "ASN database not found ... Lookups disabled until first update." at every boot, and every event still pays an anonymize_ip plus a cache insert of an empty AsnInfo. I think its worth it to make it Option and skipping the lookup entirely when there's no database.

backend/src/bot_detection/referrer_spam.txt
Woldbye, line 1: I think we should add a .gitattributes for the vendored lists? As you mentoin yourself almost all changes from this PR is auto-generated from `hosting_asns.txt` + `referrer_spam.txt` + `bot_patterns_shadow.txt`. Marking those three linguist-generated collapses them in Files changed and leaves ~1,900 reviewable lines, which might also fix the issue I flagged on the 600 line mod.rs being unable to comment exact line.

Note that I would deliberately not mark `bot_patterns.txt` or `bot_patterns_local.txt`.

Though I would pin all five to `text eol=lf`. They're include_str!'d, so on a CRLF checkout the embedded bytes differ and the binary isn't reproducible across dev machines. The CRLF handling in the scripts and in data_lines() already covers correctness, this just stops it being load-bearing.
So to summarize I recommend:
```
    backend/src/bot_detection/hosting_asns.txt         linguist-generated=true text eol=lf
    backend/src/bot_detection/referrer_spam.txt        linguist-generated=true text eol=lf
    backend/src/bot_detection/bot_patterns_shadow.txt  linguist-generated=true text eol=lf
    backend/src/bot_detection/bot_patterns.txt         text eol=lf
    backend/src/bot_detection/bot_patterns_local.txt   text eol=lf
```

Woldbye (changes requested):
Nice work! 

I verified myself that we still correctly support in-app webviews (I checked capacitorjs, Expo, Electron, and Tauri) as part of my review! Very nice. 

I flagged some important things during my review regarding selfhosters that may need an extra look.

Also, Simulate-events is broken if this merges as it posts via Node global fetch, sets only Content-Type and X-Forwarded-For. Node global fetch  sends `User-Agent: node` `node\b` is an enforced bot pattern. You might not use the script yourself, but its still the recommended approach by our CONTRIBUTIONS docs, so if we break it, we must update the docs and remove script.  I personally think the script is a good way for new developers to get started, so instead of removing it could we find a way to keep it working for localdevelopment.

static/analytics.js
```diff
@@ -121,7 +134,7 @@
         referrer: referrer,
         user_agent: userAgent,
         screen_resolution: screenResolution,
-        timestamp: Math.floor(Date.now() / 1000),
```
flopper123, line 124: Suggestion: Should we re-add the `timestamp` into the body here until next release? 🤔
Since if we do a roll-back after deployment, then our old backend wouldn't be able to support the body of cached scripts that users got in the time between deployment and rollback.

backend/src/bot_detection/mod.rs
```diff
@@ -1,12 +1,641 @@
...
+        && !user_agent.contains("Mobile")
+        && !user_agent.contains("Android")
+}
+
+fn parses_as_ip(user_agent: &str) -> bool {
+    crate::ip_parser::parse_ip_str(user_agent.trim()).is_some()
+}
+
+/// Walks parent domains so `sub.spam.com` is caught by a `spam.com` entry
+fn is_spam_referrer(referrer: &str) -> bool {
```
flopper123, line 305: Suggestion: Remove trailing `.` from the domain. `betterlytics.io` and `betterlytics.io.` both work as the exact same domain. So a spammer could send referrer as `spam.com.` and won't get caught here. So it may make sense to do `host.trim_end_matches('.')` 😄

backend/src/bot_detection/mod.rs
```diff
@@ -1,12 +1,641 @@
...
+
+    // The header UA is matched too: a forged POST can carry a clean payload UA
+    // while the real HTTP client identifies itself in the header
+    let ua = clip(user_agent);
+    let header_ua = clip(input.header_user_agent);
+    if BOT_MATCHER.is_match(ua) || (header_ua != ua && BOT_MATCHER.is_match(header_ua)) {
+        reasons.push(REASON_UA_BLOCKLIST);
+    } else if BOT_HEURISTIC_MATCHER.is_match(ua) {
+        reasons.push(REASON_UA_HEURISTIC);
+    }
```
flopper123: This seems to block our `simulate-events.js`, as Node sends `User-Agent: node` if not overriden. And UA `node` is on the blocklist. So `simulate-events.js` should probably set the `"User-Agent": event.user_agent` on the fetch headers 😄

migrations/38_create_bot_events.sql
```diff
@@ -0,0 +1,20 @@
...
+    user_agent String DEFAULT '',
+    screen_resolution String DEFAULT '',
+    event_name String DEFAULT '',
+    bot_reasons Array(LowCardinality(String)),
+    asn UInt32 DEFAULT 0,
+    asn_org LowCardinality(String) DEFAULT ''
+) ENGINE = MergeTree()
+PARTITION BY toYYYYMM(date)
+ORDER BY (site_id, date, timestamp);
+
```
flopper123: This seems to be missing the `SETTING non_replicated_deduplication_window = ...;` 😄

flopper123 (changes requested):

migrations/38_create_bot_events.sql
Lindharden: Table is now unused for self-hosters. TTL is deliberately removed from production. There is no privacy concerns with this table, it doesnt store any identifiers at all.

backend/src/geoip_updater/mod.rs
Lindharden: I have added env keys to control this better.

backend/src/config.rs
Lindharden: Fixed 👍

backend/src/bot_detection/mod.rs
Lindharden: Downgraded to shadow 👍

static/analytics.js
Lindharden: Sure, why not 👍

backend/src/bot_detection/mod.rs
Lindharden: Done 👍

backend/src/bot_detection/mod.rs
Lindharden: Fixed 👍

migrations/38_create_bot_events.sql
Lindharden: Good catch, fixed 👍

Woldbye (approved):
Looks good to me now 👍 Nice fixes

Though one leftover from my comment `docs/src/content/installation/self-hosting.mdx` still doesn't mention `ENABLE_ASN_LOOKUP` or `ENABLE_BOT_EVENT_LOG` - only .env.example got comments.. Would be nice to update it to be explicit about how it works now

flopper123 (approved):

Lindharden:
> Though one leftover from my comment `docs/src/content/installation/self-hosting.mdx` still doesn't mention `ENABLE_ASN_LOOKUP` or `ENABLE_BOT_EVENT_LOG` - only .env.example got comments.. Would be nice to update it to be explicit about how it works now

That's intended, because otherwise it goes live without being available in self-host! 😄

PR #1044, opened 2026-08-07
Made /dashboard redirect to the dashboards overview when the user has multiple dashboards

flopper123:
`/dashboard` now enters a dashboard directly only when the user has exactly one. With zero or several it redirects to `/dashboards` so the user picks, instead of landing in an arbitrary "first" dashboard (the old lookup was an unordered `findFirst`).

Closes betterlytics/betterlytics-internal#134

**Reviewer notes:** the new `findSoleUserDashboard` fetches with `take: 2` and returns the dashboard only when exactly one row comes back, so the zero and many cases share the `/dashboards` branch. Onboarding still uses `getFirstUserDashboardAction`, unchanged. Docs links keep pointing at `/dashboard` as the entry point.

### DIFF
--- a/dashboard/src/repositories/postgres/dashboard.repository.ts
+++ b/dashboard/src/repositories/postgres/dashboard.repository.ts
@@ -103,8 +103,8 @@ export async function findSoleUserDashboard(userId: string): Promise<Dashboard |
     }
 
     return DashboardSchema.parse(prismaUserDashboards[0].dashboard);
-  } catch {
-    console.error("Error while finding user's sole dashboard");
+  } catch (error) {
+    console.error("Error while finding user's sole dashboard:", error);
     throw new Error('Failed to find dashboard');
   }
 }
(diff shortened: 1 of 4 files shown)


dashboard/src/repositories/postgres/dashboard.repository.ts
```diff
@@ -85,6 +85,30 @@ export async function findFirstUserDashboard(userId: string): Promise<Dashboard
...
+    });
+
+    if (prismaUserDashboards.length !== 1) {
+      return null;
+    }
+
+    return DashboardSchema.parse(prismaUserDashboards[0].dashboard);
+  } catch {
+    console.error("Error while finding user's sole dashboard");
+    throw new Error('Failed to find dashboard');
```
Lindharden:
```suggestion
    console.error("Error while finding user's sole dashboard:", error);
    throw new Error('Failed to find dashboard');
```

Lindharden (dismissed):

Lindharden (approved):

PR #1045, opened 2026-08-08
Removed user settings from the session, cutting a database query from every auth check

Lindharden:
The session callback fetched user settings on every session resolution (every server action, tRPC request, and session poll) though only two layouts used them. Sessions now carry identity only: those layouts fetch settings directly, and the dead settings/userLastFetched JWT fields are gone.

Language keeps working as before, by a cheaper route. It stayed in sync across devices because the session callback rewrote the NEXT_LOCALE cookie wherever cookie writes were allowed; with that gone, i18n/request.ts resolves the language from the saved setting for signed-in rendering (request-memoized and shared with the layout fetch, so a page render still pays one settings query), with the cookie as the signed-out fallback, so a language change still reaches the user's other devices. The cookie write itself moved to the signIn event.

Closes betterlytics/betterlytics-internal#79

**Reviewer notes:** risk sits in i18n/request.ts. Static pages take their locale from the URL segment, so the new session lookup never runs during prerender; building this branch and its parent gives identical static/dynamic route classification. The signIn event skips new users on purpose: onboarding renders from the URL segment rather than the DB path and saves the locale it rendered in, so syncing the cookie from the freshly created default-language row would drop the user out of the language they signed up in.

### DIFF
--- a/dashboard/src/app/(app)/(protected)/layout.tsx
+++ b/dashboard/src/app/(app)/(protected)/layout.tsx
@@ -6,7 +6,7 @@ import { isUserInvitedDashboardMemberAction } from '@/app/actions/index.actions'
 import { env } from '@/lib/env';
 import { DevWidget } from '@/components/dev/DevWidget';
 import { getUserSubscription } from '@/repositories/postgres/subscription.repository';
-import { getUserSettings } from '@/services/account/userSettings.service';
+import { getCachedUserSettings } from '@/services/account/userSettings.service';
 import { UserSettingsProvider } from '@/contexts/UserSettingsProvider';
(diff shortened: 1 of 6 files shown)


Woldbye (approved):
Nice work, real performance improvement - I tested this thoroughly and found no new bugs! :) 

While testing this, I stumpled on a strange bug:
If you get invited to a dashboard, and go to /it/signup via oauth on (google for example), then after the signup, you end up with default locale on /dashboards.

I opened an issue on it at https://github.com/betterlytics/betterlytics-internal/issues/138 since I could recreate the issue on dev.betterlytics.io. 

Optional: The bug seems to be that the user-setting and the auth cookie disagrees. Since you directly touch this path, maybe you see an easy way to fix it immediately.

flopper123 (approved):

PR #1046, opened 2026-08-08
Added browser, OS, and device icons to the filter value dropdown and applied filter chips

Woldbye:
Browser, OS, and device type values now render their icon across the filter UI: the value dropdown options, the selected badges, and the applied filter chips (funnel labels too, via the shared `FilterDescription`). Mirrors #1025, which did the same for country flags.

Unrecognized values render no icon rather than the generic Globe/Monitor the tables fall back to, matching how `hasFlag()` gates country flags. The tables keep their fallbacks.

Closes betterlytics/betterlytics-internal#136

**Reviewer notes:**

- `FilterValueLabel`'s per-column wrapper components are replaced by one wrapper plus a column-to-icon switch. `CountryFlagLabel.tsx` folds into it and is deleted, since its `hasFlag()` guard is the shape all four cases now share.
- Guards resolve through the same normalization the icon components use (`resolveOSIcon` / `resolveDeviceIcon`, alongside the existing `resolveBrowser`), so a guard and its icon can't disagree on a value like "Mac OS".
- Icons are sized `size-[1.1em]` to match FlagIcon per #1041, so they track the text rather than a fixed px size. That required moving `[&_svg]:size-3` off `FilterDescription`'s container onto the `FilterColumnLabel` it was there for: as a descendant selector it outranked the icons' own size class. `FilterColumnDropdown`, the other `FilterColumnLabel` consumer, is unaffected.

msedge_XNKk75j9fW


Lindharden (approved):
Nice

flopper123 (approved):
very nice

PR #1050, opened 2026-08-10
Hid the What's New changelog button on self-host deployments

flopper123:
Adds an `enableChangelog` client feature flag (cloud only) and gates `<ChangelogModal />` behind it in `BATopbar`, so self-host topbars no longer render the sparkles / NEW button. The entries advertise subscriptions, invoices, pricing and public status pages that don't exist on self-host.

Closes betterlytics/betterlytics-internal#37

**Note on the ticket's premise:** the modal never auto-opened after an upgrade. It mounts closed; what rendered unconditionally was the topbar button (badged NEW until the user's `changelogVersionSeen` matches the current version), and the dialog opened on click. Removing the button removes the only route in. `/changelog` already redirects to `/` when not cloud, so the page needed no change.

**Reviewer notes:** the flag sits next to `enableBugReports` in `feature-flags-factory.ts` and needs no new env var, since `PUBLIC_IS_CLOUD` is the factory's only input. There is no server-side flag; `markChangelogSeenAction` and `changelogVersionSeen` are untouched and simply unreachable on self-host once the button is gone.

Typechecked only, not verified against a live self-host stack.


Woldbye (approved):
Exactly what I had in mind. Nice work 👍

Lindharden (approved):

Issue #52, opened 2025-05-07
Strip parameters from the url before inserting in database


Lindharden:
parameters are parsed in pipeline and stored in separate columns (utm campaign columns etc).

Thus it's redundant to store the entire url.

PR #1052, opened 2026-08-11
Migrated dashboard auth from next-auth to better-auth

Lindharden:
Replaced next-auth v4 with better-auth as the dashboard's auth engine: email+password, Google/GitHub OAuth, database-backed sessions, and 2FA via the twoFactor plugin. Existing accounts and bcrypt passwords carry over through a data migration (hashes move onto credential Account rows), and the v4 workarounds (manual session hack, dummy provider, string-based 2FA error protocol) are gone.

Closes betterlytics/betterlytics-internal#50
Closes betterlytics/betterlytics-internal#51
Closes betterlytics/betterlytics-internal#52

**Behavior changes and tradeoffs**

- Every active session is signed out once at cutover. No cookie bridge, by decision on #52.
- Existing 2FA enrollments are reset and need re-enrollment: better-auth derives TOTP keys from the raw stored string, so otpauth-era secrets can never produce matching codes. A one-time boot task disables them and, where email is configured, notifies affected users.
- bcrypt stays the permanent password hasher via a custom hash/verify pair. Offline rehash to scrypt is impossible, and this keeps a single hash format with zero migration risk.
- Disabling 2FA now asks for the account password instead of a TOTP code (plugin semantics). Enabling asks for the password and shows one-time backup codes.
- Backup codes are generated and stored at enrollment; redeeming them in the login UI is betterlytics/betterlytics-internal#54.
- The sign-in locale sync skips a user's first sign-in using a 60-second account-age window, since next-auth's isNewUser flag no longer exists. Outcome is equivalent.
- env.ts accepts NEXTAUTH_URL/NEXTAUTH_SECRET as fallbacks for the new BETTER_AUTH_* names, so stale selfhost entrypoints and un-renamed deploys keep booting.
- Admin bootstrap moved from the credentials authorize callback to server startup (instrumentation.ts).

**Reviewer notes**

- Reading order: `dashboard/src/lib/better-auth.ts` (engine config and hooks), the Prisma migration (renames plus credential-row backfill, sessions wiped), `src/auth/api-auth.ts` (the chokepoint keeps its return shape, so the ~110 action wrappers and tRPC context are untouched), then the client swap (mechanical useSession/signIn/signOut edits).
- Risk concentrates in the SQL migration and the login/2FA flows. The #50 characterization suite plus API-level E2E in a worktree covered password login against pre-migration hashes, the full 2FA cycle including the drift window, session revocation on reset/change-password, registration, admin bootstrap, and the env fallback.
- Companion changes ride separately: the selfhost repo branch (entrypoint derives BETTER_AUTH_SECRET with the unchanged label, README upgrade note) and production-setup (the env key rename can trail the deploy thanks to the fallback).
- Follow-up release: drop `User.totpSecret`, the one-time boot task, and selfhost's legacy env exports.


.env.example
```diff
@@ -87,9 +87,9 @@ s3_sse_enabled=false
 
 PUBLIC_BASE_URL=http://localhost:3000
 
-# NextAuth requisites
-NEXTAUTH_URL=http://localhost:3000
-NEXTAUTH_SECRET=your-secret
+# Auth (better-auth)
```
Woldbye, line 90: It might make sense to not change this: Wont it automaticly work if users dont change the env if we leave it unchanged?

If we do rename, consider a general name: AUTH_URL | AUTH_SECRET, as we will need these two for any auth vendor

Lindharden: Great idea. I've renamed to AUTH_ 😄

dashboard/prisma/schema.prisma
```diff
@@ -89,49 +89,54 @@ enum StatusPageIncidentStatus {
...
+  @@index([userId])
 }
 
 model User {
   id            String    @id @default(cuid())
   name          String?
-  email         String?   @unique
-  emailVerified DateTime?
+  email         String    @unique
+  emailVerified Boolean   @default(false)
```
Woldbye, line 129: Do we not lose important meta changing from DateTime to Boolean?

dashboard/src/services/auth/bootstrap.service.ts
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * ADMIN_PASSWORD stays authoritative so a locked-out operator can recover by
+ * setting a new value and restarting. The flip side is deliberate: changing the
+ * admin password in-app reverts on next boot.
+ */
+async function syncAdminPassword(userId: string, password: string): Promise<void> {
+  try {
+    if (await verifyUserPassword(userId, password)) return;
+
+    await updateUserPassword(userId, password);
```
Woldbye: On main changing the admin password in-app permanently retired the env value. Here every restart silently reverts it. 

Im not sure thats a good security tactic, since anyone who still knows the old env value keeps admin access across restarts.  WDYT?

dashboard/src/services/auth/bootstrap.service.ts
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * One-time cutover cleanup: the twoFactor plugin cannot verify legacy TOTP
+ * secrets (incompatible HMAC key derivation), so affected users get 2FA disabled
+ * and an email asking them to re-enroll. Delete once the release has run everywhere.
+ */
+export async function resetLegacyTwoFactor(): Promise<void> {
+  const legacyUsers = await findLegacyTwoFactorUsers();
+  if (legacyUsers.length === 0) return;
+
+  await clearLegacyTwoFactor();
```
Woldbye: You await this, so we should consider what happens if this process dies between this clear and the following `enqueueEmail` .

Most likely outcome is that the email-notices are lost for good on next boot

Woldbye (commented):
Overall very clean and well done 👍 

I had a few code comments LMK what you think

dashboard/src/services/auth/bootstrap.service.ts
Lindharden: Yes, I did this because we don't expose a CLI command or any other method to recover a lost admin account. But I agree that it might annoying that the password resets on every restart.

I reverted to the initial implementation and filed a follow-up issue for exposing some method to retrieve a lost root account: https://github.com/betterlytics/betterlytics-internal/issues/176 😄

dashboard/prisma/schema.prisma
Lindharden: Yes, I didn't consider "when" users verified as useful data. But I've readded a separate "emailVerifiedAt" column that we can use solely for audit trails. Then we can always remove it in the future if we decide it's useless 😄

See https://github.com/betterlytics/betterlytics/pull/1052/commits/db4c625a7183436f58d2649b8ce09a15537d9d56

dashboard/src/services/auth/bootstrap.service.ts
Lindharden: It's a bit hard to ensure mails happen transactionally, but we can reverse the order so emails are shipped before clearing legacy. That should make it eventually constistent 😄

Woldbye: Not sure what the best solution is. It's not a major concern: worst case they can click and retry if they never get the first email

Woldbye (dismissed):
Lgtm now :)

dashboard/src/lib/better-auth.ts
```diff
@@ -0,0 +1,168 @@
...
+        after: async (session) => {
+          try {
+            const user = await findUserById(session.userId);
+            if (!user?.createdAt) return;
+            if (Date.now() - user.createdAt.getTime() < FIRST_SIGN_IN_WINDOW_MS) return;
+
+            const settings = await getUserSettings(session.userId);
+            if (settings.language) {
+              await setLocaleCookie(settings.language);
+            }
```
flopper123, line 149: In `getUserSettings` it does:
```ts
export async function getUserSettings(...) {
    ...
    const settings = await UserSettingsRepository.findSettingsByUserId(userId);

    if (!settings) {
      return await createDefaultUserSettings(userId); // <-- It creates the default settings in here
    }
    ...
}
```

We now call the `createDefaultUserSettings` from this file, so I suggest we remove it from the `getUserSettings` as to not cause any potential clashes.
It may cause side effects, so I'm not too sure of this though.
Thoughts?

dashboard/src/components/userSettings/account/UserSecurityTotpSettings.tsx
```diff
@@ -13,79 +12,90 @@ import {
...
+      await navigator.clipboard.writeText(text);
+      setCopied(true);
+      setTimeout(() => setCopied(false), 2000);
+    } catch {
+      toast.error(failedMessage);
+    }
+  };
+
+  return { copied, copy };
+}
```
flopper123: Suggestion: Move this out to our shared hooks

dashboard/src/repositories/postgres/user.repository.ts
```diff
@@ -133,10 +155,14 @@ export async function updateUserPassword(userId: string, newPassword: string): P
...
-      where: { id: userId },
-      data: { passwordHash },
+    const result = await prisma.account.updateMany({
+      where: { userId, providerId: CREDENTIAL_PROVIDER_ID },
+      data: { password: passwordHash },
     });
+
+    if (result.count !== 1) {
+      throw new Error(`Expected exactly one credential account, updated ${result.count}`);
+    }
```
flopper123: Should this maybe be replaced by a single `update`, as it'll then auto-fail if more than one is found due to the `@@unique([providerId, accountId])` on the `Account` model:

```ts
    await prisma.account.update({
      where: { providerId_accountId: { providerId: 'credential', accountId: userId } },
      data: { password: passwordHash },
    });
```

dashboard/src/lib/better-auth.ts
```diff
@@ -0,0 +1,168 @@
...
+import { createDefaultUserSettings, getUserSettings } from '@/services/account/userSettings.service';
+import { createStarterSubscriptionForUser } from '@/services/billing/subscription.service';
+import { sendVerificationEmail } from '@/services/account/verification.service';
+import { enqueueEmail } from '@/services/email/email.service';
+import { createUserRecipientKey } from '@/services/email/recipient-key.service';
+import { setLocaleCookie } from '@/constants/cookies';
+import { isFeatureEnabled } from '@/lib/feature-flags';
+import { findUserById } from '@/repositories/postgres/user.repository';
+
+const BCRYPT_SALT_ROUNDS = 10;
```
flopper123: All these following files hardcoded the salt rounds and reuse the same bcrypt patterns:
`dashboard/src/lib/better-auth.ts`
`dashboard/src/repositories/postgres/user.repository.ts`
`dashboard/src/services/auth/bootstrap.service.ts`

I think it'd make sense to maybe create a `password.ts` in lib that only defines these once along with the bcrypt functions. Thoughts?

flopper123 (changes requested):
Very nice! 🤩 
I wasn't able to find any direct issues, and it all seems to work as expected.

Only have some small comments for it

dashboard/src/lib/better-auth.ts
Lindharden: We need it here, because for OAuth, the settings is created via the hook, which is a best-effort, but not a guarantee. And also, both are upserts so there shouldn't be any clashes 😄

Woldbye (dismissed):

flopper123 (approved):
Awesome! 😄 🙏

PR #1054, opened 2026-08-12
Session Replay for selfhost

flopper123:
Session replay on selfhost now runs against a Garage object store bundled into the selfhost image, and enabling replay without S3 storage fails fast at startup in both services (zod superRefine in the dashboard env parse, assert in the backend Config::new) instead of showing replay UI that errors at playback.

Both services split S3 access into two clients: a presigning client that signs against the public S3_ENDPOINT (browsers fetch those URLs and the signature covers the host), and a data-plane client on the new optional S3_INTERNAL_ENDPOINT for real server-side calls, which would otherwise hairpin through the public URL from inside the container. With S3_MANAGE_BUCKET_RULES=true the backend applies CORS and lifecycle rules to the bucket at startup (retention from REPLAY_RETENTION_DAYS, falling back to DATA_RETENTION_DAYS), so a selfhost deploy needs no manual bucket, credential, or CORS work.

Part of betterlytics/betterlytics-internal#158 (main-repo portion; compose, setup.sh, and docs land in betterlytics-selfhost).

Deviations from the ticket: fail-fast at startup replaces "treat the feature as disabled and warn"; the bundled store is Garage rather than MinIO; BYO-S3 is dropped as a supported path (the code still reads the standard S3_* vars, but only the bundled Garage setup is documented and supported). Issue text to be updated to match.

Reviewer notes: new env vars are S3_INTERNAL_ENDPOINT (both services), S3_MANAGE_BUCKET_RULES and REPLAY_RETENTION_DAYS (backend only); all optional and off by default, so cloud deploys are unaffected. A deploy with SESSION_REPLAYS_ENABLED=true but no S3 config now refuses to boot, intentionally. Not yet verified end to end against a live selfhost stack.


flopper123:
Replaced by https://github.com/betterlytics/betterlytics/pull/1075

PR #1055, opened 2026-08-13
Added the running version to the sidebar footer for self-hosted deploys

Woldbye:
The publish workflow now bakes the release tag into the selfhost image (build arg `APP_VERSION`, exposed to the app as a server-side env var), and the dashboard renders it as muted text in the sidebar footer. Release images show their exact tag, local and manual builds show `dev`, cloud hides it entirely (continuous deploys have no meaningful tag).

Closes betterlytics/betterlytics-internal#172

Deviations from ticket:
- The ticket asked for a public (browser-readable) env var. Implemented server-only (`APP_VERSION`, no `PUBLIC_` prefix) instead, for security: `PUBLIC_` keys are auto-served to anonymous callers by the public environment endpoint, and an exact version readable without auth lets anyone fingerprint an instance for known vulnerabilities. Selfhost instances lag releases, which is exactly when that matters. None of the surveyed competitors (umami, plausible, posthog, matomo) disclose their version pre-auth, matomo deliberately hides it. The version renders only on authenticated pages, gated off share pages and cloud.
- The cloud gate is a server-side flag (`isCloud` in feature-flags.ts) rather than the client factory named in the ticket, since the sidebar is a server component. Behavior is identical, cloud sets both flags.

Reviewer notes:
- The publish.yml build-arg path only runs on a real release. If it were ever miswired the image just shows `dev`, nothing breaks; worth one glance at the footer during the next release soak.
- No selfhost compose changes needed, the value ships as image ENV.

Verified: SSR footer shows `dev` with cloud flags off and nothing with them on; logged-out fetches of the public environment endpoint and of a share page contain no version in any form; a test image built with `--build-arg APP_VERSION=v9.9.9-test` reported that value at runtime (before the mechanical ENV rename; the tagged path gets its one-glance confirmation at the next release soak).

## Demo
image

image


Lindharden (commented):
I'm unsure about the shifts & spacing the version number causes in the main app sidebar.

I think we should consider moving it into the settings sidebar, since it's sidebar footer is currently completely empty (left aligned?) and anyone who cares about the version is usually heading to settings anyway. This also seems to be what other major providers does 🤔

Woldbye:
Fine by me :) though it will be harder to notice then

Woldbye:
@Lindharden I have moved it into the dashboard settings and left-aligned:
image

Lindharden (approved):

PR #1058, opened 2026-08-15
Documented Session Replay support for self-hosted instances

flopper123:
Updated the docs for Session Replay shipping on selfhost: the self-hosting guide gets a Session Replay section (bundled recording storage, `SESSION_REPLAYS_ENABLED`, `REPLAY_RETENTION_DAYS`), the cloud vs self-hosted comparison table now shows replay as supported, and the "Cloud only, requires S3-compatible storage" wording is removed from the FAQ, installation, and integration pages.

Part of betterlytics/betterlytics-internal#158 (docs portion, alongside betterlytics/betterlytics-selfhost#6 and betterlytics/betterlytics#1054).


flopper123:
Superseded by [#1105](https://github.com/betterlytics/betterlytics/pull/1105)

PR #1059, opened 2026-08-15
Gated the cloud-only features, contact, vs and demo pages off-cloud

Lindharden:
Selfhosted instances served the Betterlytics marketing and competitor-comparison pages on the operator's own domain, and `/demo` redirected to the literal `/share/undefined` with `DEMO_DASHBOARD_ID` unset. `/features`, `/contact`, `/vs/[competitor]` and `/demo` now redirect home when `IS_CLOUD` is false, using the same gate as `about/page.tsx`.

Closes betterlytics/betterlytics-internal#155

Reviewer notes: the gate is deliberately per-page rather than in the `(public)` layout, since that segment also holds accept-invite, forgot-password, reset-password and verify-email, which must stay reachable off-cloud. `generateStaticParams` on `/vs/[competitor]` is left alone (the selfhost image prerenders no page routes, so the runtime gate holds). The public topbar and footer still link to these routes off-cloud; that cleanup is tracked separately in betterlytics/betterlytics-internal#175.

Verification: build, lint and unit tests pass. Runtime redirect behavior was not exercised, so the `IS_CLOUD=false` acceptance check is still open.

### DIFF
--- a/dashboard/src/app/(app)/[locale]/(public)/contact/page.tsx
+++ b/dashboard/src/app/(app)/[locale]/(public)/contact/page.tsx
@@ -8,6 +8,8 @@ import { Button } from '@/components/ui/button';
 import { Card, CardHeader, CardTitle, CardDescription, CardFooter } from '@/components/ui/card';
 import type { SupportedLanguages } from '@/constants/i18n';
 import type { ReactNode } from 'react';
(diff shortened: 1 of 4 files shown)


Woldbye (approved):
LGTM 👍 Follows existing pattern exactly

PR #1061, opened 2026-08-16
Served web-vitals first-party from the analytics script's origin

Woldbye:
The tracker now derives its scripts base URL from the analytics script's own src origin and loads a vendored web-vitals 5.3.0 from there instead of unpkg, removing the only third-party request in the visitor path. An explicit `data-scripts-base-url` still wins. The selfhost image now ships `web-vitals.js` so it loads first-party there too (replay.js shipping is covered by #1075), and the docs describe the first-party loading and CSP.

`static/web-vitals.js` is the npm registry tarball's iife build, byte-identical below a 7-line license banner (entry added to THIRD-PARTY-LICENSES.md). It is deliberately never re-minified.

Do not merge until the pending security release has been tagged.

Closes betterlytics/betterlytics-internal#161
Closes betterlytics/betterlytics-internal#162


Dockerfile.selfhost
```diff
@@ -76,8 +76,9 @@ FROM node:24-slim AS static-builder
 
 RUN echo "Minifying static files..."
 WORKDIR /app
-COPY static/analytics.js .
-RUN npx --yes esbuild analytics.js --minify --outfile=analytics.min.js
+COPY static/analytics.js static/replay.js ./
+RUN npx --yes esbuild analytics.js --minify --outfile=analytics.min.js \
+ && npx --yes esbuild replay.js --minify --outfile=replay.min.js
```
flopper123: This is already done in https://github.com/betterlytics/betterlytics/pull/1075 🫡

docs/src/content/integration/web-vitals.mdx
```diff
@@ -66,6 +66,16 @@ Metrics are reported on visibility change or pagehide to capture stable final va
...
+## First-party loading and CSP
+
+The `web-vitals` library loads first-party from the same origin the analytics script itself was loaded from - no request leaves for any third-party host. If you set `data-scripts-base-url`, that base is used instead.
+
+If your site sets a Content-Security-Policy, `script-src` only needs to allow your Betterlytics origin (for the hosted service, `https://betterlytics.io`).
+
+<Callout type="info">
+If you proxy `analytics.js` through your own domain, the browser will request `web-vitals.js` (and `replay.js` for Session Replay) from that same domain. Either proxy those files too, or set `data-scripts-base-url` explicitly to your Betterlytics origin.
+</Callout>
+
```
flopper123: I have a feeling that this section may cause more confusion than clarity over this. Setting the `data-scripts-base-url` should only be relevant for self-hosters, and by then it would already allow the web-vitals script to load.

The callout is for people who proxy the `analytics.js`, however, I don't know if that's a thing that users actually do, at least not something we've instructed them to do.

So I'd suggest that some of this (if even) should be placed in the self-host section instead of for the web-vitals integration page.

Thoughts? 😄

docs/src/content/integration/session-replay.mdx
```diff
@@ -209,7 +209,7 @@ Use `data-disable-replay-on-urls` patterns to disable recording on routes like `
 - Consent not applied: if you rely on a CMP, either call `setReplayConsent(true)` after consent resolves, or render the script with `data-consent-replay="true"` only after consent is granted.
 - Disabled by route pattern: check `data-disable-replay-on-urls`; ensure current URL doesn’t match a disabled pattern.
 - Sampling too low: raise `data-replay-sample` during testing (e.g., 100) and revert later.
-- Script loading: if self‑hosting, confirm `data-scripts-base-url` is correct and `replay.js` is reachable.
+- Script loading: `replay.js` loads from the same origin as the analytics script by default. If you proxy `analytics.js` through your own domain, either proxy `replay.js` too or set `data-scripts-base-url` explicitly to your Betterlytics origin.
```
flopper123: I'd suggest we leave these changes out for the session replay docs update 😎

flopper123 (changes requested):
Very nice! Only some small comments 😄

flopper123 (approved):
Nice! 😄

PR #1064, opened 2026-08-16
Extracted a shared useCopy clipboard hook

Lindharden:
Eight components each had their own copy-to-clipboard code (writeText, a transient `copied` flag, a reset timer, and inconsistent error handling). They now share `useCopy` in `dashboard/src/hooks/use-copy.ts`, which takes an optional `failedMessage` (error toast, omit to fail silently) and `resetAfterMs`, and returns `{ copied, copy }` where `copy` resolves to whether the write succeeded so a call site can gate its own follow-up.

Closes betterlytics/betterlytics-internal#181

## Per call site

Mechanics only, no intended UX change. What each site shows today and after:

| Call site | Visible behavior | Change |
|---|---|---|
| `CopyButton.tsx` | icon swap, 1400ms | passes `resetAfterMs: 1400`; local `useCopyToClipboard` deleted |
| `UserSecurityTotpSettings.tsx` | icon swap 2000ms + `copyFailed` error toast | local `useCopy(failedMessage)` deleted, same message passed as an option |
| `IntegrationSheet.tsx` | Copy/Copied label swap 2000ms + `failedToCopy` error toast | keyed `copiedIdentifier` state collapsed to `copied` (the file has one copy target) |
| `CodeBlock.tsx` | icon swap, 2000ms | inline handler removed |
| `McpTokenManager.tsx` | icon swap, 2000ms | inline handler removed |
| onboarding `Integration.tsx` | icon swap 2000ms, silent on failure | keyed `copiedIdentifier` collapsed to `copied` (one copy target); still silent (no `failedMessage`) |
| `ErrorDetailHeader.tsx` | Share/Copied label + icon swap, 2000ms | inline handler removed |
| `StatusPagesClient.tsx` | `linkCopied` success toast, no transient state | see below |

## Reviewer notes

Two failure-path behaviors could not be preserved exactly, both because the old code did not wait on the clipboard promise:

- `ErrorDetailHeader` and `CopyButton` previously showed the confirmation state unconditionally, including when the write failed. They now show it only on success.
- `StatusPagesClient` previously showed the "Link copied" success toast unconditionally, again including on failure. It now shows that toast only on success, and a new `actions.linkCopyFailed` toast otherwise. That key is the only new string, added to all four locale files.

Success paths are unchanged everywhere, and `writeText` is still called synchronously inside the click handler at every site, so the browser user-gesture requirement still holds.

The hook also adds unmount cleanup of the reset timer, which four of the five inlined versions lacked.

### DIFF
--- a/dashboard/messages/da.json
+++ b/dashboard/messages/da.json
@@ -4075,6 +4075,7 @@
       "viewPage": "Vis side",
       "more": "Flere handlinger",
       "linkCopied": "Link kopieret til udklipsholderen",
+      "linkCopyFailed": "Kunne ikke kopiere linket",
       "openPublicPage": "Åbn offentlig side",
       "editPage": "Rediger side",
       "manageIncidents": "Administrer hændelser",
--- a/dashboard/messages/en.json
+++ b/dashboard/messages/en.json
@@ -4072,6 +4072,7 @@
       "viewPage": "View page",
       "more": "More actions",
       "linkCopied": "Link copied to clipboard",
+      "linkCopyFailed": "Failed to copy link",
       "openPublicPage": "Open public page",
       "editPage": "Edit page",
       "manageIncidents": "Manage incidents",
--- a/dashboard/messages/it.json
+++ b/dashboard/messages/it.json
@@ -4075,6 +4075,7 @@
       "viewPage": "Vedi pagina",
       "more": "Altre azioni",
       "linkCopied": "Link copiato negli appunti",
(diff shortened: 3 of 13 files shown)


Woldbye (approved):
Approved, good as is. 

Though, worth noting that the `navigator.clipboard` only exists in a secure context, so IF we start supporting self-hosters on http lan in the future, they wont be able to copy to clipboard. The fix is adding the `execCommand('copy')` fallback to the clipboard path (note that execCommand is deprecated, but its still the right solution for http), to future proof this.

PR #1065, opened 2026-08-16
Added a flag to disable the favicon proxy's DuckDuckGo egress

Lindharden:
Favicons still come from DuckDuckGo out of the box, but self-hosts that cannot (or would rather not) make that outbound request can now set `PUBLIC_ENABLE_FAVICON_FETCHING=false`. With it off the dashboard builds no `/api/favicons` URLs at all and renders the existing globe fallback, and the route short-circuits before touching the upstream, so direct hits and hotlinking cannot cause egress either. The flag is defined once in `createFeatureFlags`, so a future global internet-features switch can be AND-ed in there without touching call sites.

Closes betterlytics/betterlytics-internal#169

Deviation: the ticket says the disabled route returns "its negative response". It returns the same 404 but with `Cache-Control: no-store` instead of the shared 24h negative cache, so re-enabling the flag is not masked by a stale 404 sitting in a browser or proxy cache.

Reviewer notes: new env var, defaulting to true via a new `zStringBooleanDefaultTrue` helper, so deploys that never set it behave exactly as today. Documented in `.env.example`; docs-site coverage deferred to a future advanced env-var table.


flopper123 (dismissed):
Awesome 🥳

flopper123 (approved):

PR #1068, opened 2026-08-16
Removed the unused Prisma Accelerate extension

Lindharden:
## Summary

The Prisma client was wrapped in the Accelerate extension, which only engages on a `prisma://` connection string. We connect straight to our own Postgres over `postgresql://`, and nothing in the codebase uses `cacheStrategy`, so the extension was a per-query pass-through plus an eager engine start. With it gone the exported client is a plain `PrismaClient` again, so the `as PrismaClient` cast that existed only to hide the extension's modified type is removed and the dev global cache is typed instead of `unknown`.

Closes betterlytics/betterlytics-internal#67

## Reviewer notes

No self-host impact. Routing to Accelerate is decided by the connection string in Prisma core, not by this extension, and the stack cannot boot on an Accelerate-only URL anyway (`prisma migrate deploy`, role provisioning and the Rust backend all need the raw Postgres wire protocol). Prisma's hosted Postgres hands out direct `postgresql://` URLs, which a plain client handles fine.

## Verification

Unit tests, lint and build pass, and a read plus an update through the exported client succeed against a dev worktree Postgres. Sign-in was not exercised through the browser in this run.

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -32,7 +32,6 @@
     "@number-flow/react": "^0.5.11",
     "@opentelemetry/api": "^1.9.1",
     "@prisma/client": "^6.8.2",
-    "@prisma/extension-accelerate": "^1.3.0",
     "@radix-ui/react-accordion": "^1.2.12",
     "@radix-ui/react-alert-dialog": "^1.1.14",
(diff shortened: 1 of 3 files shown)


Woldbye (approved):
LGTM

PR #1069, opened 2026-08-16
Stopped building favicon requests for non-domain values

Lindharden:
`getFaviconUrl` now returns null for values that are not domain-shaped (whitespace anywhere, or no dot after normalization), so the demo sidebar's synthetic "Demo Dashboard" label never builds a request that `/api/favicons` would only reject. `DomainFavicon` already renders the globe fallback on null, so the visible result is unchanged. Guarding the builder rather than the label closes the failure class for any future placeholder value.

Closes betterlytics/betterlytics-internal#45

The guard is a strict superset of what `domainValidation` accepts, so every stored dashboard domain still resolves its favicon exactly as before.


dashboard/src/lib/favicons.ts
```diff
@@ -18,6 +20,12 @@ export function getFaviconUrl(domain?: string | null): string | null {
     return null;
   }
 
+  // The favicon route validates with this same schema, so placeholder labels like the demo
+  // sidebar's "Demo Dashboard" would only earn a 404 — filter them before requesting.
```
flopper123: I'd remove the comment. Code is self-describing.

flopper123 (changes requested):
Nice, just a small comment

flopper123 (dismissed):

flopper123 (approved):

PR #1070, opened 2026-08-16
Extracted shared BASiteQuery builder for MCP tools

Lindharden:
MCP tools each hand-built a `BASiteQuery` and filled dummy values for the fields they never read. `buildSiteQuery` in `src/mcp/utils/` now owns those defaults (granularity, queryFilters, userJourney, timezone), and `errors.ts`, `globalProperties.ts` and `userJourneys.ts` call it instead. New tools no longer copy the boilerplate, and a `BASiteQuery` field change touches one place.

Closes betterlytics/betterlytics-internal#8

Spec item 4 (making `userJourney` optional on `BASiteQuery` itself) was left out on purpose. It was flagged as an optional follow-up in the ticket and is a wider entity change than this cleanup.

Reviewer notes: no effective query values change. `user_journeys` still passes its real `maxSteps` and `limit`. The only literal difference is `list_global_properties`, whose filler `userJourney` goes from `{1, 1}` to the shared default `{3, 50}`; neither the global-properties repository nor the errors service reads `userJourney`, so the emitted queries are identical.

### DIFF
--- a/dashboard/src/mcp/tools/errors.ts
+++ b/dashboard/src/mcp/tools/errors.ts
@@ -6,6 +6,7 @@ import {
   dateOrderRefinement,
 } from '@/mcp/entities/mcp.entities';
 import { resolveTimeRange } from '@/mcp/utils/resolveTimeRange';
+import { buildSiteQuery } from '@/mcp/utils/buildSiteQuery';
 import {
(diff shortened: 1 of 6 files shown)


flopper123 (approved):
Nice

PR #1071, opened 2026-08-16
Locked resend verification email to the signed-in account

Lindharden:
`resendVerificationEmailAction` now runs through `withUserAuth` and mails the session user's own address, ignoring anything the client sends. Before this it was callable with no session and with any email, and its replies (success vs. generic failure vs. "wait N minutes") told the caller whether an account existed and whether it was already verified.

Closes betterlytics/betterlytics-internal#82

Reviewer notes:

- The action's return type changed to the standard `ServerActionResponse` envelope, so both callers now read `result.error.message` instead of `result.error`. They only surface it when `error.name === 'UserException'`; anything else falls back to the existing localized copy, so masked internal errors never reach a toast in raw English.
- The 5 minute cooldown and its "wait N minutes" text are unchanged in behaviour, just delivered as a `UserException` (the only user-safe message left; every other failure is logged server side and comes back generic). One side effect worth knowing: `withActionSpan` treats every `UserException` as an errored span and logs it, so ordinary cooldown hits will now appear as error-level logs and errored `action.resendVerificationEmailAction` spans in Grafana. That is the existing convention for `UserException` across the codebase (same as `assertRoleAuthorized`), so it was left alone rather than special-cased here.
- Unauthenticated callers hit `requireAuth`'s redirect to /signin before any account lookup runs. That redirect is thrown outside `withActionSpan` and re-thrown by `unstable_rethrow`, so it reaches the client as a real redirect rather than a masked error.
- New colocated test at `dashboard/src/app/actions/auth/verification.action.test.ts` exercises the real `withUserAuth` chain (only `@/auth/api-auth` and the service layer are mocked), including the no-session case and the "a client-supplied email is ignored" regression guard.

### DIFF
--- a/dashboard/messages/da.json
+++ b/dashboard/messages/da.json
@@ -3296,6 +3296,7 @@
     "termsOfServiceRequired": "Du skal acceptere servicevilkårene for at fortsætte.",
     "urlMustBeOnDomain": "URL skal være på {domain} eller et underdomæne",
     "monitorAlreadyExists": "En monitor for denne URL eksisterer allerede",
+    "verificationEmailCooldown": "{minutes, plural, one {Vent venligst # minut, før du anmoder om en ny bekræftelsesmail.} other {Vent venligst # minutter, før du anmoder om en ny bekræftelsesmail.}}",
     "statusPageSlugTaken": "Denne URL-slug er allerede i brug",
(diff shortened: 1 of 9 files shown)


Lindharden:
Superseded by https://github.com/betterlytics/betterlytics/pull/1079

PR #1072, opened 2026-08-17
Added copyable invite links for member invitations

Lindharden:
Inviting a member no longer relies on email delivery alone: creating an invite now shows a panel with the selectable accept-invite link (plus a note when emails are disabled and nothing was sent), and every pending invitation row gets a copy action with a tooltip. The copy helper falls back to `execCommand` on plain HTTP, which selfhost supports.

Closes betterlytics/betterlytics-internal#157

**Deviation:** spec 4 called for a browser-origin fallback when `PUBLIC_BASE_URL` is unset on non-cloud deploys. Instead the variable is now required (`z.string().url()`, fail-fast at startup), so the betterlytics.io default can never leak. Every environment already provides it: selfhost derives it from `DOMAIN` in `.env.base`, release image builds load it from `.env.production.example` via `.env.build`, and the dev env files set it. This avoids shipping a "was it explicitly configured" flag only the fallback needed.

**Reviewer notes:**

- Risk sits in `shared.env.ts` (no more default). Unit tests pass with the variable unset, so CI is unaffected.
- The cancelled/expired acceptance criterion needed no change: `acceptInvitation` already rejects non-pending and expired invitations server-side.
- New-user invitees still dead-end at signup while `ENABLE_REGISTRATION` is off, tracked in betterlytics/betterlytics-internal#39.
- The four locale files are mechanical string additions.


dashboard/src/lib/env/shared.env.ts
```diff
@@ -8,7 +8,7 @@ export const zStringBoolean = z
 
 export const sharedEmailEnvSchema = z.object({
   IS_CLOUD: zStringBoolean,
-  PUBLIC_BASE_URL: z.string().optional().default('https://betterlytics.io'),
+  PUBLIC_BASE_URL: z.string().url(),
```
Woldbye, line 17: Remember to add this explicitly to our production environment before deployment if its missing.

Woldbye, line 17: Requiring `PUBLIC_BASE_URL` crashes the app locally. 

It parses at module import in both the app and the worker. Locally when I try to build this, it literally crashes with `[TypeError: Cannot set property message of [object Object] which has only a getter]`, since there's no dashboard/.env next doesn't know the environment. 

The dotenv injection reaches the node runtime, but the edge sandbox only sees vars Next itself loaded from dashboard/.env*. Untill this branch, every single variable in `sharedEmailEnvSchema` had a default - so the module parser never failed anywhere.

Woldbye (changes requested):
I think this still needs some work. 

We need to find a better way to require `PUBLIC_BASE_URL`, as its breaking local builds (see code comment)

dashboard/src/components/CopyButton.tsx
```diff
@@ -17,7 +17,19 @@ function useCopyToClipboard(resetAfterMs = 1400) {
 
   const copy = useCallback(
     (text: string) => {
-      navigator.clipboard.writeText(text);
+      if (navigator.clipboard) {
+        navigator.clipboard.writeText(text);
```
flopper123: Quibble: Missing `await`, so this will always show the success message even though the copy may fail.

flopper123 (dismissed):
Very nice! Minor optional comment 😄

dashboard/src/lib/env/shared.env.ts
Woldbye: So I did some digging, and what I have diagnosed above is not quite correct.

The reason the build failed, is that this branch is so incredibly stale from main, that it still uses the old NEXTAUTH_SECRET and TOTP_SECRET_ENCRYPTION_KEY, which I obviously no longer have in my .env.

Im approving this as a result, and writing a follow up immediately that adds proper errors for runtime-zod validation errors: `[TypeError: Cannot set property message of [object Object] which has only a getter]` is simply unacceptable

Woldbye (dismissed):
Changed to approved, as I found the real error, and its due to this branch being based on an old version of main, with environments we no longer used.

flopper123 (approved):

PR #1073, opened 2026-08-17
Rate limited credential sign-in per account and surfaced throttling errors in the auth forms

Lindharden:
better-auth already rate limits its own endpoints per IP in production (sign-in and sign-up at 3 per 10s, password reset and verification email requests at 3 per 60s, 2FA at 3 per 10s, in-memory, on by default). This PR adds the dimension it cannot give us: a per-account throttle on credential sign-in (10 attempts per 10 minutes, sliding window), so rotating IPs no longer buys unlimited guesses against one account. The budget is cleared whenever a session is created for the account, so only sustained failures ever block a legitimate user. The 429 carries a Retry-After header, and the login form, signup auto-sign-in, and 2FA settings dialogs now show a "too many attempts" message instead of "invalid credentials" when throttled (all four locales).

Part of betterlytics/betterlytics-internal#12. The registration, password reset, and verification actions named there are covered by the stacked follow-up PRs that migrate those flows into better-auth (part of betterlytics/betterlytics-internal#55). PR 1 of 5 in that stack.

Reviewer notes:

- The limiter is in-process and bounded (10k keys, oldest evicted, no sweep timer), per the "in-memory per-process is acceptable" call in the ticket.
- Accepted residual: an attacker actively spraying failures at a known email can hold that account's password login closed while the spray lasts (window is 10 minutes, OAuth unaffected). That is inherent to any per-account limiter; clearing on success confines it to active attacks.
- Deploy caveat, pre-existing and independent of this PR: better-auth resolves the client IP from x-forwarded-for and deliberately ignores multi-value headers unless advanced.ipAddress.trustedProxies is configured. If the production proxy appends to XFF rather than overwriting it, every built-in per-IP limit collapses into one shared site-wide bucket (3 shared sign-in attempts per 10s). Worth verifying the prod nginx sets a single-value XFF before this posture is relied on.
- Follow-up candidate, not in this PR: the 2FA verify endpoints only have the per-IP 3/10s plugin rule, so TOTP/backup-code guessing per account is unbounded across IPs. Needs cookie- or user-keyed limiting and touches the backup-codes work in betterlytics/betterlytics-internal#54.

Verified live against a worktree stack: 10 wrong passwords then 429 with Retry-After; and 5 failures, 1 success, 6 failures produces no 429 (the success cleared the budget).

### DIFF
--- a/dashboard/messages/da.json
+++ b/dashboard/messages/da.json
@@ -757,7 +757,7 @@
             "invalidOtp": "Ugyldig bekræftelseskode. Tjek dine oplysninger og prøv igen.",
             "invalidCredentials": "Ugyldig e-mail eller adgangskode. Tjek dine oplysninger og prøv igen.",
             "generic": "Der opstod en fejl under login. Prøv venligst igen.",
-            "tooManyAttempts": "For mange forsøg. Vent et par minutter og prøv igen."
+            "tooManyAttempts": "For mange forsøg. Prøv igen senere."
           },
           "twoFactor": {
             "title": "To-faktor-godkendelse",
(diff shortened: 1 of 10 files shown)


Lindharden:
Closing without merging: we're moving the auth flows onto better-auth's own endpoints instead (better-auth ships built-in rate limiting), so we no longer want to introduce a custom limiter. #1074 has been rebased onto main without these commits.

PR #1074, opened 2026-08-18
Migrated change-password onto better-auth's endpoint

Lindharden:
The change-password dialog now calls better-auth's `/change-password` endpoint (`authClient.changePassword` with `revokeOtherSessions: true`); the custom server action, service function, and `verifyUserPassword` repository helper are deleted. The password-changed email is preserved via `databaseHooks.account.update.after` gated on the endpoint path, so it fires off the DB-confirmed password write and reuses the existing `sendPasswordChangedNotification` helper.

Part of betterlytics/betterlytics-internal#12 and betterlytics/betterlytics-internal#55. PR 1 of 4.

Closes https://github.com/betterlytics/betterlytics-internal/issues/193


Woldbye (commented):
LGTM except for one thing:

I can't find the per-session-token throttle from the PR text anywhere on the branch - only the 429 handling in the dialog. What actually guards /change-password right now is better-auth's built-in limiter: IP-keyed, 3 attempts/10s, production-only

dashboard/src/components/userSettings/account/ChangePasswordDialog.tsx
```diff
@@ -114,16 +114,23 @@ export default function ChangePasswordDialog({ open, onOpenChange }: ChangePassw
...
           newPassword: validated.newPassword,
+          revokeOtherSessions: true,
         });
-        if (result.success) {
+        if (!error) {
           toast.success(t('toast.success'));
           resetForm();
           onOpenChange(false);
+        } else if (error.code === 'INVALID_PASSWORD') {
+          toast.error(t('toast.invalidCurrentPassword'));
```
flopper123: Question: Should invalid password maybe be like `WEAK_PASSWORD` instead of a toast? 🤔

flopper123 (commented):
Very nice, only one comment 😄

dashboard/src/components/userSettings/account/ChangePasswordDialog.tsx
```diff
@@ -114,16 +114,23 @@ export default function ChangePasswordDialog({ open, onOpenChange }: ChangePassw
     try {
       const validated = ChangePasswordSchema.parse(passwords);
       startTransition(async () => {
-        const result = await changePasswordAction({
+        const { error } = await authClient.changePassword({
           currentPassword: validated.currentPassword,
           newPassword: validated.newPassword,
+          revokeOtherSessions: true,
```
flopper123: Question: Should this also be done on the backend to ensure password change is always revoking other sessions, like before? 🤔 Seems like it can be done in a better-auth hook

Lindharden:
> I can't find the per-session-token throttle from the PR text anywhere on the branch - only the 429 handling in the dialog. What actually guards /change-password right now is better-auth's built-in limiter: IP-keyed, 3 attempts/10s, production-only

PR was outdated. I've updated the PR description, the only guard now is better-auth's built-in limiter IP-keyed, 3 req/10s on `/change-password`, production-only, in-memory.

I'm revisiting token-based rate limit after all endpoints are migrated to better-auth, as none of these are regression 😄

flopper123 (approved):

PR #1075, opened 2026-08-18
Added ClickHouse session replay storage and proxied all replay traffic through the app layer

flopper123:
Session replay no longer requires S3 or a browser-reachable object store. All replay traffic now flows through our own services (script uploads to the backend, playback through a dashboard API route), with the storage backend behind that pipe selected by `REPLAY_STORAGE`: ClickHouse (new, selfhost default) or S3 (cloud, now fully internal). The presign handshake, the bundled Garage, and the finalize endpoint are gone; per-chunk metadata rides the upload, and errors attach via a slim `/replay/error` POST.

Closes betterlytics/betterlytics-internal#87

**Reviewer notes**

- Reading order: `backend/src/session_replay/` (upload path, `SegmentStore` seam), then `dashboard/src/repositories/*replaySegments*` + `app/api/replay/segment` (`ReplaySegmentReader` seam), then `static/replay.js`.
- Migration 40 adds `analytics.session_replay_segments` plus a `storage` marker on `session_replays`; the post-migrate step applies retention TTL from `REPLAY_RETENTION_DAYS` so both storage modes agree.
- New env `REPLAY_STORAGE=clickhouse|s3`. Unset resolves to `s3` when `S3_ENABLED=true`, else `clickhouse`, so existing deployments change nothing. `S3_INTERNAL_ENDPOINT` and the SSE handshake are removed (both unreleased).
- Upload hardening rides along: same site policies as `/event`, timestamps rejected outside now +/- 24h, 200 MB per-session cap. The per-site daily budget is deferred to betterlytics/betterlytics-internal#189.
- Script contract change: one PUT per chunk (keepalive on small bodies, so tail chunks survive page exit); `/replay/presign/put` and `/replay/finalize` no longer exist. A cached pre-deploy `replay.js` stops recording until it refreshes, worth checking the cloud cache TTL on that file at rollout.

Verified end-to-end on a worktree stack in both modes: ClickHouse (record, segments in CH, playback, error attach) and S3 via MinIO. Not verified against real AWS S3.


backend/src/session_replay/store.rs
```diff
@@ -0,0 +1,155 @@
...
+            Self::S3(s3) => {
+                s3.put_segment(
+                    &object_key(site_id, session_id, filename),
+                    bytes,
+                    gzip.then_some("gzip"),
+                )
+                .await
+                .map_err(StoreError::Storage)?;
+                Ok(())
+            }
```
Lindharden: Nit: perhaps extract each match body to a helper function for readability?

backend/src/session_replay/mod.rs
Lindharden: Can you add some spacing here, it's quite hard to read when it's not newlined into sections 😆

backend/src/session_replay/store.rs
```diff
@@ -0,0 +1,155 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::InvalidPayload(msg) => write!(f, "invalid payload: {}", msg),
+            Self::Storage(e) => write!(f, "storage failure: {}", e),
+        }
+    }
+}
+
+// Enum + static dispatch, not a trait: two known variants, no async_trait/dyn
+// machinery. Promote to a trait only if a third backend ever appears.
```
Lindharden:
```suggestion
```

backend/src/main.rs
```diff
@@ -235,11 +237,31 @@ async fn main() {
...
+            if config.enable_session_replay && config.replay_storage == ReplayStorage::S3 {
+                panic!("Failed to initialize S3 service (required for replay storage): {}", e);
+            }
             warn!("Failed to initialize S3 service: {}", e);
             None
         }
     };
 
+    if config.enable_session_replay && config.replay_storage == ReplayStorage::S3 {
+        info!("REPLAY_RETENTION_DAYS applies to ClickHouse data only; expire S3 objects under the 'site/' prefix with a bucket lifecycle rule");
```
Lindharden: Maybe it should be a warn?

```suggestion
        warn!("REPLAY_RETENTION_DAYS applies to ClickHouse data only; expire S3 objects under the 'site/' prefix with a bucket lifecycle rule");
```

backend/src/main.rs
```diff
@@ -235,11 +237,31 @@ async fn main() {
...
+    }
+
+    // Built only when replay is enabled, so the config assert has already validated
+    // the storage mode for this config.
```
Lindharden: I think this is quite implicit from the .then()

```suggestion
```

Lindharden (approved):

PR #1076, opened 2026-08-19
Migrated forgot/reset password onto better-auth

Lindharden:
Forgot/reset password now runs through better-auth's `requestPasswordReset` and `resetPassword` endpoints. The custom server actions, service functions, and the `PasswordResetToken` table are gone (the migration drops it); tokens live in better-auth's `Verification` table instead.

Three hardening pieces on top of the plain migration:

- Tokens are stored SHA-256 hashed through a `verification.storeIdentifier` override, so a leaked DB snapshot yields no usable link.
- Reset links are single-use per user: issuing a new one deletes the outstanding ones, and `onPasswordReset` clears the rest after redemption.
- OAuth-only accounts get neither a reset email nor a redeemable token. Redeeming one would silently attach a password login to an account that never had one.

Session revocation moves to better-auth's `revokeSessionsOnPasswordReset`, so `invalidateAllUserSessions` and its repository helper are deleted. The password-changed email is preserved via `onPasswordReset`. Both forms now show localized errors for invalid token, weak password, and 429, replacing the hardcoded English strings.

Closes betterlytics/betterlytics-internal#194. Part of betterlytics/betterlytics-internal#55. PR 2 of 4.

Accepted by design: a failure inside `sendResetPassword` (token pruning or email enqueue) is caught and logged by better-auth's `runInBackgroundOrAwait` and never reaches the caller, so `/request-password-reset` still returns `status: true` and the form shows its success message. That is intentional. The success response has to be indistinguishable from the unknown-email case or the endpoint becomes an account-enumeration oracle; a genuine send failure is visible in the error logs rather than to the requester.


Woldbye (dismissed):
Nice work :) Looks good to me, couldnt find a single flaw in the threat model

dashboard/src/lib/better-auth.ts
```diff
@@ -83,14 +110,21 @@ export const auth = betterAuth({
...
+      // Redeeming a reset for an OAuth-only account would attach a password login to it
+      if (ctx.path === '/reset-password') {
+        const token = ctx.body?.token ?? ctx.query?.token;
+        if (typeof token === 'string' && token) {
+          const userId = await findResetTokenUserId(resetTokenStoredIdentifier(token));
+          if (userId && !(await findCredentialAccount(userId))) {
+            throw new APIError('BAD_REQUEST', { message: 'Invalid token', code: 'INVALID_TOKEN' });
+          }
+        }
+      }
```
flopper123: Question: Instead of blocking the reset password for OAuth-only users at reset time, would it make sense to do so at request time instead, such that better-auth can validate the OAuth situation prior to minting the token?
So adding something like:
```ts
if (ctx.path === '/request-password-reset') {
  const email = ctx.body?.email;
  if (typeof email === 'string') {
    const user = await findUserByEmail(email);
    if (user && !(await findCredentialAccount(user.id))) {
      return ctx.json({
        status: true,
        message: 'If this email exists in our system, check your email for the reset link',
      });
    }
  }
}
```

dashboard/src/repositories/postgres/user.repository.ts
```diff
@@ -213,7 +214,7 @@ export async function anonymizeUser(userId: string): Promise<void> {
       prisma.account.deleteMany({ where: { userId } }),
       prisma.session.deleteMany({ where: { userId } }),
       prisma.twoFactor.deleteMany({ where: { userId } }),
-      prisma.passwordResetToken.deleteMany({ where: { userId } }),
+      prisma.verification.deleteMany({ where: { identifier: { startsWith: RESET_TOKEN_PREFIX }, value: userId } }),
```
flopper123: Question: is the `RESET_TOKEN_PREFIX` important here, or should we simply delete many where value is userId, something like:
```ts
prisma.verification.deleteMany({ where: { value: userId } }),
```

flopper123 (changes requested):
Looks good. Two small questions only.
If they don't matter then it's ready for approval 😄

flopper123 (dismissed):

flopper123 (approved):

PR #1078, opened 2026-08-20
Migrated signup to better-auth

Lindharden:
Email/password signup now goes through better-auth's `/sign-up/email`: SignupForm calls `authClient.signUp.email`, and `registerUserAction`, `registerNewUser`, and `registerUser` are deleted. The user-create hooks are now the single provisioning path for credential and OAuth users alike: the before hook rejects sign-ups without terms acceptance and stamps `termsAcceptedAt`/`termsAcceptedVersion`, and the after hook creates the starter subscription and the settings row with the signup language, and sends the verification email when `enableAccountVerification` is on.

Sign-up is gated on `enableRegistration` inside better-auth (`disableSignUp`), so self-host deploys with registration off refuse at the endpoint. The prisma adapter gains `transaction: true`, so the user and credential account commit atomically; subscription and settings stay best-effort in the after hook, same as OAuth signups today.

Closes betterlytics/betterlytics-internal#195. Part of betterlytics/betterlytics-internal#55. PR 3 of 4.

### DIFF
--- a/dashboard/src/lib/better-auth.ts
+++ b/dashboard/src/lib/better-auth.ts
@@ -186,8 +186,7 @@ export const auth = betterAuth({
             extra.termsAcceptedAt = new Date();
             extra.termsAcceptedVersion = CURRENT_TERMS_VERSION;
           }
-          if (Object.keys(extra).length === 0) return;
-          return { data: { ...user, ...extra } };
+          return { data: { ...user, ...extra, createdAt: new Date(), updatedAt: new Date() } };
         },
         after: async (user, ctx) => {
           try {
--- a/dashboard/src/lib/better-auth.ts
+++ b/dashboard/src/lib/better-auth.ts
@@ -16,6 +16,8 @@ import { setLocaleCookie } from '@/constants/cookies';
 import { isFeatureEnabled } from '@/lib/feature-flags';
 import { findUserById, findCredentialAccount } from '@/repositories/postgres/user.repository';
 import { PasswordSchema } from '@/entities/auth/password.entities';
+import { CURRENT_TERMS_VERSION } from '@/constants/legal';
+import { SUPPORTED_LANGUAGES, type SupportedLanguages } from '@/constants/i18n';
 import {
   RESET_TOKEN_EXPIRY_SECONDS,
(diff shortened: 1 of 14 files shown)


dashboard/src/lib/better-auth.ts
```diff
@@ -136,20 +157,28 @@ export const auth = betterAuth({
...
+          const extra: Record<string, unknown> = {};
+          if (user.emailVerified) {
+            extra.emailVerifiedAt = new Date();
+          }
+          if (ctx?.path === '/sign-up/email') {
+            extra.name = user.name?.trim() || null;
+            extra.termsAcceptedAt = new Date();
+            extra.termsAcceptedVersion = CURRENT_TERMS_VERSION;
+          }
+          return { data: { ...user, ...extra, createdAt: new Date(), updatedAt: new Date() } };
```
flopper123: The new email signups seems to now be missing a `role` that the old `registerUser` had 🤔

dashboard/src/lib/better-auth.ts
```diff
@@ -35,15 +38,19 @@ const PASSWORD_POLICY_FIELDS: Record<string, string> = {
...
+  const language = (body as { language?: unknown } | undefined)?.language;
+  return SUPPORTED_LANGUAGES.includes(language as SupportedLanguages) ? (language as SupportedLanguages) : undefined;
+}
+
 export const auth = betterAuth({
   appName: 'Betterlytics',
   baseURL: env.AUTH_URL,
   secret: env.AUTH_SECRET,
-  database: prismaAdapter(prisma, { provider: 'postgresql' }),
+  database: prismaAdapter(prisma, { provider: 'postgresql', transaction: true }),
```
flopper123, line 50: The user create after hook will still be called even when the transaction fails:
```ts
databaseHooks: {
    user: {
      create: {
        after: ...
``` 

It may make sense to add a guard to the hook, maybe something like:
```ts
if (!(await findUserById(user.id))) return;
```

flopper123 (changes requested):
Very nice! 😄 

Only found a small issue

flopper123 (dismissed):

flopper123 (approved):

PR #1079, opened 2026-08-22
Migrated email verification to better-auth

Lindharden:
Email verification now runs on better-auth: sign-up sends via `sendOnSignUp`, resends call `/send-verification-email` from the client, and the emailed link is verified by better-auth's `/verify-email`, which redirects to our page with `?verified=1` or `&error=<code>`. The custom token table, repository, entities, crypto utils and the verify/resend server actions are deleted, and a migration drops `VerificationToken`.

The 5 minute resend cooldown is now enforced by the pg-boss singleton (one verification email per user per 5 minutes) on top of better-auth's built-in per-IP limit (3/min) on the resend endpoint. A throttled enqueue surfaces as a 429 instead of a silent success, with a localized cooldown toast on both resend surfaces. `emailVerifiedAt` stays in sync via a user update before-hook.

Closes betterlytics/betterlytics-internal#196. Part of betterlytics/betterlytics-internal#55. PR 4 of 4.


dashboard/src/services/account/verification.service.ts
```diff
@@ -1,166 +1,26 @@
...
+
+export async function sendVerificationEmail(
+  user: { id: string; email: string; name: string | null },
+  url: string,
+): Promise<void> {
+  if (!isFeatureEnabled('enableAccountVerification')) return;
+  const result = await enqueueEmail({
+    type: 'email-verification',
+    recipientKey: createUserRecipientKey(user.id),
+    campaignKey: 'email-verification',
```
flopper123, line 18: Campaign key is now just a constant, so I think it may incorrectly dedupe consecutive verification emails 🤔

dashboard/src/lib/better-auth.ts
```diff
@@ -64,6 +64,11 @@ export const auth = betterAuth({
       await sendPasswordChangedNotification(user.id, user.email, user.name ?? null);
     },
   },
+  emailVerification: {
+    sendVerificationEmail: ({ user, url }) => sendVerificationEmail({ ...user, name: user.name ?? null }, url),
```
flopper123: I think this may brake unverified emails for OAuth signups as the URL will then point to `/dashboards` (for Github for instance where you don't need a verified email from them). I think it makes sense to do something like:
```ts
sendVerificationEmail: ({ user, url }) => {
  const link = new URL(url);
  link.searchParams.set('callbackURL', '/verify-email?verified=1');
  return sendVerificationEmail({ ...user, name: user.name ?? null }, link.toString());
},

```

flopper123 (changes requested):
Awesome, only a few findings 🥳

dashboard/src/services/account/verification.service.ts
Lindharden: The dedupe is intentional, it's the 5 min cooldown. `throttleSeconds: 300` on the email type maps to pg-boss singletonSeconds, so the constraint is `(name, singleton_on, singleton_key)` with `singleton_key = email-verification:user:<id>` one verification email per user per 5-minute slot 😄

flopper123 (approved):

PR #1080, opened 2026-08-22
2 fa backup codes

Lindharden:
Adds single-use backup codes so a TOTP user who loses their authenticator can still sign in. Codes are shown once at enrollment with copy and download, can be regenerated from security settings, and are accepted at the second-factor step of login.

Closes betterlytics/betterlytics-internal#54

The login input carries the risk. Codes are mixed case and compared exactly, so the field normalizes without re-casing: it strips non-alphanumerics, caps length, and re-inserts the hyphen. It also sets autoCapitalize="none", without which mobile keyboards capitalize the first character and silently reject valid codes.

image

Closes https://github.com/betterlytics/betterlytics-internal/issues/54


flopper123 (approved):
Nice, works great 😄

PR #1084, opened 2026-08-23
fix(funnels): persist funnel step order explicitly (#963)

### LINKED ISSUE #963: Saved funnel step order is not guaranteed
Funnel steps aren't persisted with an explicit order, so loading a saved funnel can return them in a different sequence than the user defined. Because funnels are order-dependent, this silently changes both the displayed steps and the computed conversion numbers. Most likely to surface after a funnel is edited.

Order-dependence confirmed by swapping only the condition order over the same events (A then B): 

`windowFunnel(3600)(ts, A, B) → 2`
`windowFunnel(3600)(ts, B, A) → 1`

MohammadYusif:
## Summary

- Funnel steps were stored without any order column and read back without an `orderBy`, so the sequence a user authored survived only as long as Postgres happened to return the rows in insertion order.
- `updateFunnel` deletes every step and recreates it, which is precisely when that stops holding — the new rows reuse space freed by the deleted ones, in whatever order the free-space map hands it back. That matches the report that this is most likely to surface after a funnel is edited.
- The array order is load-bearing downstream: `getFunnelDetails` feeds the per-step conditions positionally into ClickHouse `windowFunnel(...)`, which is order-dependent, and `toFunnel` zips `funnelSteps[i]` against `visitors[i]`. So a reshuffle silently changed both the displayed steps and the computed conversion numbers. The edit dialog's drag-to-reorder was effectively discarded on save.
- Fixed by storing the order explicitly, following the existing `StatusPageMonitor.position` precedent rather than inventing a new pattern.

## Changes

- `dashboard/prisma/schema.prisma`: add `position Int @default(0)` to `FunnelStep`.
- `dashboard/prisma/migrations/20260823120000_add_funnel_step_position/migration.sql`: add the column and backfill existing rows with `ROW_NUMBER() OVER (PARTITION BY "funnelId" ORDER BY "id") - 1`. Prisma mints one cuid per step in write order and cuids sort lexicographically by creation time, so ordering by the primary key recovers the authored order for existing data. This is best-effort recovery for rows written before the column existed — happy to drop the backfill and let everything default to 0 if you'd rather not rely on that.
- `dashboard/src/repositories/postgres/funnels.repository.ts`: write `position` from the array index in `createFunnel` and `updateFunnel`, and add `orderBy: { position: 'asc' }` to the shared `FUNNEL_STEPS_INCLUDE` so both read paths return steps in stored order.
- `dashboard/src/repositories/postgres/funnels.repository.test.ts`: new tests covering position assignment on create, renumbering after a reorder on update, and both read paths ordering by position. All four fail without the repository change.

`FunnelStepFilter` ordering is deliberately unchanged — filters within a step are `AND`-ed, so their order is not semantically significant.

Fixes #963

### DIFF
--- a/dashboard/prisma/migrations/20260823120000_add_funnel_step_position/migration.sql
+++ b/dashboard/prisma/migrations/20260823120000_add_funnel_step_position/migration.sql
@@ -1,0 +1,14 @@
+-- AlterTable
+ALTER TABLE "FunnelStep" ADD COLUMN     "position" INTEGER NOT NULL DEFAULT 0;
+
+-- MigrateData: the step order was never stored, so recover the best signal still available.
+-- Prisma mints one cuid per step in the order the steps are written, and cuids sort
+-- lexicographically by creation time, so ordering by the primary key restores the authored
+-- order for the funnels the application has written.
+UPDATE "FunnelStep" AS "step"
+SET "position" = "ordered"."position"
+FROM (
+    SELECT "id", (ROW_NUMBER() OVER (PARTITION BY "funnelId" ORDER BY "id") - 1)::INTEGER AS "position"
+    FROM "FunnelStep"
+) AS "ordered"
+WHERE "step"."id" = "ordered"."id";
--- a/dashboard/prisma/migrations/20260823120000_add_funnel_step_position/migration.sql
+++ b/dashboard/prisma/migrations/20260823120000_add_funnel_step_position/migration.sql
@@ -2,13 +2,20 @@
 ALTER TABLE "FunnelStep" ADD COLUMN     "position" INTEGER NOT NULL DEFAULT 0;
 
 -- MigrateData: the step order was never stored, so recover the best signal still available.
--- Prisma mints one cuid per step in the order the steps are written, and cuids sort
--- lexicographically by creation time, so ordering by the primary key restores the authored
--- order for the funnels the application has written.
+-- Prisma mints one cuid per step in write order and cuid v1 sorts lexicographically by
+-- creation time, so the primary key recovers the authored order for every step the
+-- application has written. Some steps backfilled by 20251130102124 got gen_random_uuid()
+-- instead, which carries no order at all; nothing has rewritten those rows since, so their
+-- physical position is still the order that migration inserted them in.
 UPDATE "FunnelStep" AS "step"
 SET "position" = "ordered"."position"
 FROM (
-    SELECT "id", (ROW_NUMBER() OVER (PARTITION BY "funnelId" ORDER BY "id") - 1)::INTEGER AS "position"
+    SELECT
+        "id",
+        (ROW_NUMBER() OVER (
+            PARTITION BY "funnelId"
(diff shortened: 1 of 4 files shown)


Woldbye (dismissed):
Thank you for your contribution @MohammadYusif 

Nice fix, I think its exactly the right seam 👍

dashboard/prisma/migrations/20260823120000_add_funnel_step_position/migration.sql
```diff
@@ -0,0 +1,14 @@
...
+-- Prisma mints one cuid per step in the order the steps are written, and cuids sort
+-- lexicographically by creation time, so ordering by the primary key restores the authored
+-- order for the funnels the application has written.
+UPDATE "FunnelStep" AS "step"
+SET "position" = "ordered"."position"
+FROM (
+    SELECT "id", (ROW_NUMBER() OVER (PARTITION BY "funnelId" ORDER BY "id") - 1)::INTEGER AS "position"
+    FROM "FunnelStep"
+) AS "ordered"
+WHERE "step"."id" = "ordered"."id";
```
Lindharden, line 21: Nice work! The diagnosis in the description is right and I appreciate
that you traced it through to `windowFunnel` and `toFunnel` rather than just
adding the column.

One change needed before this can go in, and it's in the backfill.

`ORDER BY "id"` is sound reasoning for rows Prisma wrote as cuid v1 leads with a
base36 millisecond timestamp, so lexicographic order is creation order. But not
every `FunnelStep` row has a cuid. The `20251130102124_add_filter_steps_as_model`
migration backfilled steps from `Funnel.queryFilters` using `gen_random_uuid()`,
and any funnel that hasn't been re-saved since still has those. A v4 UUID carries
no ordering at all, so those funnels get shuffled.

```suggestion
-- AlterTable
ALTER TABLE "FunnelStep" ADD COLUMN     "position" INTEGER NOT NULL DEFAULT 0;

-- MigrateData: the step order was never stored, so recover the best signal still available.
-- Prisma mints one cuid per step in write order and cuid v1 sorts lexicographically by
-- creation time, so the primary key recovers the authored order for every step the
-- application has written. Some steps backfilled by 20251130102124 got gen_random_uuid()
-- instead, which carries no order at all; nothing has rewritten those rows since, so their
-- physical position is still the order that migration inserted them in.
UPDATE "FunnelStep" AS "step"
SET "position" = "ordered"."position"
FROM (
    SELECT
        "id",
        (ROW_NUMBER() OVER (
            PARTITION BY "funnelId"
            ORDER BY CASE WHEN "id" LIKE '%-%' THEN NULL ELSE "id" END NULLS LAST, "ctid"
        ) - 1)::INTEGER AS "position"
    FROM "FunnelStep"
) AS "ordered"
WHERE "step"."id" = "ordered"."id";
```

From my tests, with the `ORDER BY id` version, 5 funnels come out in a different order than they're in
now, and with my proposed suggestion, 0 seems to come out in the wrong order.

Lindharden (changes requested):
Thank you for your contribution @MohammadYusif 🙏

Lindharden (approved):

PR #1085, opened 2026-08-23
Moved monitor alert emails into the worker so SMTP self-hosters receive them

Lindharden:
The Rust backend no longer sends email. Monitor down, recovery and SSL alerts are enqueued as pg-boss `send-email` jobs and rendered by the worker's existing pipeline, so SMTP-only self-hosters now receive them and alerts gain retries, per-recipient dedup and consistent branding. The backend's MailerSend-only email stack is deleted.

Closes betterlytics/betterlytics-internal#43

**Deviation from the ticket:** spec 6 wanted a cross-stack test as the guard against pg-boss upgrades. It was built, but it needed a Postgres service on the unit-test job plus a throwaway database per run, which was too much overhead for the risk, so it was dropped. pg-boss is pinned to an exact version instead, with a comment on the queue definition pointing at the Rust insert. The acceptance criterion "CI fails if a pg-boss upgrade breaks the backend's job insert" is therefore not met: bumping pg-boss is now a deliberate, reviewed act.

**Reviewer notes**
- New env vars `JOB_QUEUE_DATABASE_URL` and `POSTGRES_JOBQUEUE_RW_PASSWORD`, and a new `jobqueue_rw` role granted on the `pgboss` schema by `provision_roles.js`. With uptime monitoring enabled the backend refuses to boot without them. Self-host provisioning is a separate PR: betterlytics/betterlytics-selfhost#10.
- Risk is concentrated in `backend/src/jobqueue/mod.rs`, which mirrors pg-boss's own insert (queue defaults, and `policy` copied onto the row so the exclusive singleton index applies).


backend/src/jobqueue/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+use tokio_postgres::error::SqlState;
+use tokio_postgres::types::Json;
+use tracing::info;
+
+use crate::postgres::{PostgresError, PostgresPool};
+
+/// Mirrors pg-boss's own insert (`pg-boss/dist/plans.js` insertJobs) for the pinned version:
+/// queue defaults come from `pgboss.queue`, and `policy` must be copied onto the row for the
+/// exclusive-policy singleton index to apply. Always returns one row: `queue_exists`, and `id`
+/// (NULL when the queue is missing or an identical job is already pending).
```
flopper123, line 34: Consider adding a check or something for this. A `pnpm update` could break this silently (though it'll probably be caught during testing) 😄

flopper123 (dismissed):
Very nice! 😁
Only one small comment, but still approved 👍

backend/src/jobqueue/mod.rs
Lindharden: I have moved pgboss schema creation to app initialization and made Rust fail-fast when schema version is incorrect: https://github.com/betterlytics/betterlytics/pull/1085/commits/37b37547a21296aa5a4a1823034e5f172c3f7649

flopper123 (approved):

PR #1089, opened 2026-09-04
Name the failing variables when env validation fails instead of a masked TypeError

Woldbye:
Env validation failures now name every failing variable instead of crashing with `TypeError: Cannot set property message of [object Object] which has only a getter`.

## Description of why

- zod v3's `ZodError.message` is getter-only, and Next's instrumentation error wrapping assigns to `err.message`, so any env schema failure at boot erased the real issues behind a generic TypeError - any missing required variable was undiagnosable.
- Closes betterlytics/betterlytics-internal#210

## Changes

- Added a `parseEnv(label, schema)` helper: `safeParse`, then a plain writable-message `Error` listing every issue as `NAME: reason`, tagged with the failing schema (`app`/`shared`/`worker`) and pointing at `.env.example`
- Routed the three env modules through it; validation timing is unchanged (eager at build/boot)
- Unit tests, including a regression test that the thrown error is not a `ZodError` and its `message` is writable

## Additional Notes

Failure output after this change, with `AUTH_SECRET` and `INTEGRATION_ENCRYPTION_KEY` commented out:

```
Error: An error occurred while loading instrumentation hook: Invalid environment variables (app): AUTH_SECRET: Required; INTEGRATION_ENCRYPTION_KEY: Required. Compare your .env against .env.example.
    at parseEnv (src\lib\env\parse-env.ts:12:9)
    at __TURBOPACK__module__evaluation__ (src\lib\env.ts:106:28)
    at <unknown> (C:\Users\cwr\personal\betterlytics\dashboard\.next\server\chunks\_f05743ca._.js:30:16)
    at async registerOpenTelemetry (src\instrumentation.ts:43:21)
    at async Module.register (src\instrumentation.ts:2:3)
  10 |     .map((issue) => `${issue.path.join('.') || '(env)'}: ${issue.message}`)
  11 |     .join('; ');
> 12 |   throw new Error(`Invalid environment variables (${label}): ${issues}. Compare your .env against .env.example.`);
     |         ^
  13 | }
  14 |
```

### DIFF
--- a/dashboard/src/lib/env/parse-env.ts
+++ b/dashboard/src/lib/env/parse-env.ts
@@ -1,0 +1,13 @@
+import { z } from 'zod';
+
+// Plain Error on purpose: Next assigns to err.message while ZodError's is getter-only, masking the issues
+export function parseEnv<T extends z.ZodType>(label: string, schema: T, source: unknown = process.env): z.output<T> {
+  const result = schema.safeParse(source);
+  if (result.success) {
+    return result.data;
+  }
+  const issues = result.error.issues
+    .map((issue) => `${issue.path.join('.') || '(env)'}: ${issue.message}`)
+    .join('; ');
+  throw new Error(`Invalid environment variables (${label}): ${issues}. Compare your .env against .env.example.`);
+}
--- a/dashboard/src/lib/env/parse-env.ts
+++ b/dashboard/src/lib/env/parse-env.ts
@@ -9,5 +9,5 @@ export function parseEnv<T extends z.ZodType>(label: string, schema: T, source: 
   const issues = result.error.issues
     .map((issue) => `${issue.path.join('.') || '(env)'}: ${issue.message}`)
     .join('; ');
-  throw new Error(`Invalid environment variables (${label}): ${issues}. Compare your .env against .env.example.`);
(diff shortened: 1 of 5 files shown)


dashboard/src/lib/env/parse-env.ts
```diff
@@ -0,0 +1,13 @@
...
+// Plain Error on purpose: Next assigns to err.message while ZodError's is getter-only, masking the issues
+export function parseEnv<T extends z.ZodType>(label: string, schema: T, source: unknown = process.env): z.output<T> {
+  const result = schema.safeParse(source);
+  if (result.success) {
+    return result.data;
+  }
+  const issues = result.error.issues
+    .map((issue) => `${issue.path.join('.') || '(env)'}: ${issue.message}`)
+    .join('; ');
+  throw new Error(`Invalid environment variables (${label}): ${issues}. Compare your .env against .env.example.`);
```
Lindharden:
```suggestion
  throw new Error(`Invalid environment variables (${label}): ${issues}. Check the environment passed to the dashboard.`);
```

Lindharden (approved):
LGTM 👏

flopper123 (approved):

Issue #153, opened 2025-06-05
Average visit duration on SummaryCard seems incorrect


flopper123:
![Image](https://github.com/user-attachments/assets/f901128b-7d6b-4ee6-bb4e-3871b5b1ab6c)

Lindharden:
Mine appears correct. I'm unsure why you're getting different results?

![Image](https://github.com/user-attachments/assets/91d18818-de7a-47ef-8b4d-86a076665bee)

PR #1090, opened 2026-09-05
Moved custom status domain routing from the proxy into the app

flopper123:
The dashboard now routes custom status domains itself: a Node-runtime middleware classifies every request's Host against the status page table (cached, coalesced, bounded) and rewrites `/` to the domain page, 404s everything else on that host, and 503s with Retry-After when the lookup cannot be answered. Any reverse proxy can now sit in front of the app with no path allowlist, slug regex, or rewrite, which unblocks the selfhost setup (#153) and the own-proxy docs (#168).

Closes betterlytics/betterlytics-internal#167

Reviewer notes
- Read in order: `lib/status-host-routing.ts` (pure policy: reserved hosts, route decision), `services/analytics/statusHost.service.ts` (cache and DB lookup), `middleware.ts` (wiring, now `runtime: 'nodejs'` with a `/((?!_next/).*)` matcher; the old path exclusions moved inside so status hosts can 404 them).
- One reserved-host policy now serves both claiming and routing: `isOwnNamespace` delegates to `isOwnHost`. Off cloud only the app host and its `www` are reserved, so a selfhoster on `analytics.acme.com` can claim `status.acme.com`.
- `ExpiringSet` replaces the hand-rolled negative cache in the `ask` guard and backs the new caches; no behaviour change there.
- The `/status/domain/<host>` path still passes on a status host so the current prod Caddy rewrite keeps working. Deploy the app first, then simplify the proxy.
- No new env vars. `STATUS_PAGE_DOMAIN` is ignored off cloud; the CNAME target shown in the editor is the bare app host off cloud.

Verified with tsc and the three new vitest files. Not verified against a live proxy or Postgres outage; the manual checks are a `pnpm build`, a curl with a custom Host header, unpublishing a page (404), and stopping Postgres (503 only for cold hosts).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -131,7 +131,9 @@ BACKGROUND_JOBS_ENABLED=false
 PUSHOVER_APP_TOKEN=
 INTEGRATION_ENCRYPTION_KEY=your-32-character-encryption-key
 
-# Public status pages. The status-host proxy (TLS, rewrites, rate limiting) lives in the
-# production setup repo; these two are consumed by the app itself.
+# Public status pages. Host routing for custom status domains lives in the app itself and needs
+# no proxy config. STATUS_PAGE_ASK_SECRET is required whenever public status pages are enabled
+# outside development (the selfhost setup derives it) and is used only by a proxy doing on-demand
+# TLS through /api/status-page/ask.
 STATUS_PAGE_DOMAIN=status.localhost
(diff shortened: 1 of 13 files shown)


Lindharden (approved):
Nice 👏

Lindharden (approved):

PR #1094, opened 2026-09-06
Dropped aws-lc from the backend build and made S3 setup fail fast on missing keys

Lindharden:
The backend compiled aws-lc (a C/asm crypto library built via cmake) twice over, once through our own `rustls`/`tokio-rustls` defaults and once through the AWS SDK's built-in HTTPS client, while `probe.rs` already installs ring as the process-wide provider. Everything now uses ring only, the S3 client is built directly from the `S3_*` env vars instead of going through aws-config's credential chain, and a handful of dead or over-featured dependencies are gone (`config`, `lazy_static`, `serde_yaml`, prometheus' protobuf encoder, sysinfo's disk/network/user probing). cmake is no longer needed in the build images.

Clean `cargo build --release` on a dev machine: 9m 08s before, 6m 13s after. Total compile CPU time 3177s to 2323s, crates 403 to 335.

Depends on #1075.

Reviewer notes:
- Self-host contract change: with `S3_ENABLED=true`, `S3_BUCKET`, `S3_ACCESS_KEY_ID` and `S3_SECRET_ACCESS_KEY` are now required and the backend exits at startup naming the missing var.
- No change for R2 or MinIO: endpoint, path style and static keys work as before.

### DIFF
--- a/Dockerfile.selfhost
+++ b/Dockerfile.selfhost
@@ -7,7 +7,6 @@ RUN apt-get update && apt-get install -y \
     pkg-config \
     libssl-dev \
     build-essential \
-    cmake \
     git \
     && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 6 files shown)


flopper123 (approved):
Nice 😎

PR #1095, opened 2026-09-06
Accepted pasted URLs as dashboard domains and previewed the cleaned domain

Lindharden:
Pasting a full URL (`https://www.example.com/some/page?x=1`) into the domain field now resolves to `example.com` instead of failing the format check. When the input gets cleaned, the create dialog and onboarding step show "Will be created as example.com" under the field so the result is never a surprise, and the helper text no longer tells users to avoid `https://`.

Everything after the first `/`, `?`, `#` or `:` is dropped, not just a trailing slash.

Closes betterlytics/betterlytics-internal#187


flopper123 (approved):
Nice, works great 😄

Issue #155, opened 2025-06-06
Add browser icons where browsers are displayed


Lindharden:
Prepend browser icon (ideally as svgs) in front of entries:

Devices page:
![Image](https://github.com/user-attachments/assets/ca86dda4-f991-4947-bca1-42801a3f8b30)
![Image](https://github.com/user-attachments/assets/973b9942-2178-4d87-8d41-703a3e1884e9)

Overview page:

![Image](https://github.com/user-attachments/assets/c7f15b5c-db3d-4b1f-8dc7-395ec1674b1e)

PR #1100, opened 2026-09-08
Hide links to cloud-gated routes in public nav and footers off-cloud

flopper123:
## What

Off-cloud (`IS_CLOUD=false`), the public topbar and footers no longer link to routes that redirect home on selfhost deploys.

- New server `CloudOnly` wrapper (`components/CloudOnly.tsx`) on `isFeatureEnabled('isCloud')`.
- `PublicTopBar`: Features and Pricing (desktop + mobile menu) gated via an `isCloud` prop from the `(public)` layout. The topbar is a client component with one render site, so a prop needs no env provider.
- `(public)` layout renders `MinimalFooter` instead of `Footer` off-cloud. `Footer` itself is untouched and cloud-only.
- `MinimalFooter`: privacy/terms/contact are cloud-only; off-cloud it shows Documentation, GitHub and Discord instead (existing `public.footer` keys), so selfhosters keep a docs and community support path. Language selector and copyright stay in both modes.
- No translation changes.

## Why

#155 gated the marketing and legal routes off-cloud, so every one of these links dead-ended through a redirect to `/` and then signin on selfhosted instances.

## Reviewer notes

- Deviates from spec 4 of the issue ("external links stay untouched"): off-cloud the full footer is replaced, so Status and Bluesky go with it. Docs, GitHub and Discord are kept because every open-core reference we checked (Plausible CE, PostHog, Sentry, GitLab) keeps docs plus a community support link on self-host while dropping vendor billing, status and legal.
- On-cloud output is unchanged.
- Verified with `tsc --noEmit`; manual check off-cloud on `/signin`, `/signup`, `/forgot-password`, `/reset-password`, `/accept-invite`, `/verify-email` still to be done.

Closes betterlytics/betterlytics-internal#175

### DIFF
--- a/dashboard/src/app/(app)/[locale]/(public)/layout.tsx
+++ b/dashboard/src/app/(app)/[locale]/(public)/layout.tsx
@@ -2,11 +2,12 @@ import { Footer } from '@/components/footer/Footer';
 import PublicTopBar from '@/components/topbar/PublicTopBar';
 import { SUPPORTED_LANGUAGES } from '@/constants/i18n';
 import ThemeToggleFab from '@/components/ThemeToggleFab';
+import { isFeatureEnabled } from '@/lib/feature-flags';
 
 export default async function LocaleLayout({ children }: { children: React.ReactNode }) {
   return (
     <div className='flex min-h-screen flex-col justify-between'>
-      <PublicTopBar />
+      <PublicTopBar isCloud={isFeatureEnabled('isCloud')} />
       <div className='flex flex-1 flex-col'>{children}</div>
       <Footer />
       <ThemeToggleFab />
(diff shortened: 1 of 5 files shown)


dashboard/src/app/(app)/[locale]/(public)/layout.tsx
```diff
@@ -1,14 +1,18 @@
...
 export default async function LocaleLayout({ children }: { children: React.ReactNode }) {
+  const isCloud = isFeatureEnabled('isCloud');
+
   return (
     <div className='flex min-h-screen flex-col justify-between'>
-      <PublicTopBar />
+      <PublicTopBar isCloud={isCloud} />
       <div className='flex flex-1 flex-col'>{children}</div>
-      <Footer />
+      {isCloud ? <Footer /> : <MinimalFooter />}
```
Comment (Woldbye) [inline dashboard/src/app/(app)/[locale]/(public)/layout.tsx:15]: Quibble:
The topbar already handles the conditional toggling internally. It seems to me like footer should follow the same pattern for consistency (i.e. pass isCloud, and conditionally show content internally)

Woldbye (approved):
Nice work. Had a minor quibble but looks good to me as is!

dashboard/src/app/(app)/[locale]/(public)/layout.tsx
Comment (flopper123) [reply dashboard/src/app/(app)/[locale]/(public)/layout.tsx]: Done 🫡

Lindharden (approved):

PR #1101, opened 2026-09-13
Streamed whole session replays in one request

flopper123:
The replay player now loads a session through a single `GET /api/replay/segments` NDJSON stream (one segment per line, gzipped as it goes) instead of one request per segment plus a manifest call. ClickHouse mode streams straight from the query; S3 mode composes the per-segment reads with a small prefetch window. Playback starts on the first line.

Reviewer notes
- Risk is in `route.ts` and `web-stream-pipeline.ts`: `stream.pipeline` so a source error ends the response and a client disconnect cancels the ClickHouse query, and the same callback releases the concurrency slot (3 open streams per user, 24 per process, 429 with `Retry-After`).
- Cache headers are age based: a session whose `ended_at` is over an hour old is cached immutably, younger ones for 60 s. There is no completion flag.
- `LIMIT 1 BY filename` collapses duplicate segment rows from ingest retries.
- Client: the reader is cancelled on early exit and the container effect resets on unmount. A malformed line fails the session loudly (our script cannot produce one, so it would mean a format bug).

Not verified live: cancel propagation, the concurrency cap, and the error path. Local Node 20.12 has a `Readable.toWeb` backpressure bug (fixed in 20.17+, production image is 20.20) that drains the whole stream instantly, so those checks need a dev server on a newer Node. Headers, dedupe, tsc, lint and vitest are verified. S3 mode not exercised.


Lindharden (changes requested):
The new concurrency cap (24 per process) is a hard limit on a user-facing feature, but hitting it is currently invisible: the route returns 429 with no log line or metric, and the client shows a generic "Failed to fetch segments" without retrying. The first time we'd hear about it is from a user.

Could we add a couple of prom-client metrics here, following the pattern in `mcp/metrics.ts`?

- A gauge of currently open replay streams (increment on acquire, decrement in the release callback). This is what tells us we're approaching 24 before anyone is refused.
- A counter of refused streams, labelled by reason (`user` / `process`). Process-level refusals should be zero in normal operation, so a single hit is worth an alert.

A stream duration histogram would be a nice bonus, since lifetime x concurrent viewers is what decides whether 24 is enough

PR #1103, opened 2026-09-18
Let invited users and the first account sign up when registration is disabled

Lindharden:
`ENABLE_REGISTRATION=false` (the self-host default) also blocked people who had been invited to a dashboard: the invite link bounced signed-out users to sign-in with no way to create an account. Registration now only governs unsolicited sign-ups. An invitation, or an empty instance, unlocks sign-up for that one address, matching how Plausible, PostHog and Grafana handle closed instances.

Reviewer notes:
- The gate moved from better-auth's static `disableSignUp` into `hooks.before` (`/sign-up/email`) and `databaseHooks.user.create.before`, so it covers OAuth too. `signupGate.service.ts` is the one place that decides. The invite exemption requires the token, not just knowledge of the address; provider-verified OAuth emails may match on address alone.
- `/accept-invite/<token>` now sends new addresses to `/signup?invite=`, existing ones to `/signin?callbackUrl=`, and both forms return to the invite afterwards. The sign-up form shows which dashboard the invite is for and locks the email. `toSafeRelativePath` validates `callbackUrl`; better-auth already validates its own `callbackURL`, this covers ours.
- `ADMIN_EMAIL` / `ADMIN_PASSWORD` are optional. With both empty, `/signin` on an empty instance redirects to `/signup` and the first sign-up creates the account. Whoever reaches it first wins, the same model as the products above; worth a line in the self-host docs. `bl-wt` still sets both, so dev spinups are unchanged.
- Self-host sign-ups no longer show or require the terms checkbox. Every downstream consumer of `termsAcceptedAt` (modal, onboarding, dashboard creation) was already cloud-gated; the sign-up form was the only place that was not.
- Cloud is unchanged in practice: registration is enabled there, so `getSignupAllowance` short-circuits without touching the database, and the terms check still runs.

Not verified: OAuth sign-up through an invite (no provider credentials locally). Email/password path verified end to end on a fresh database.


dashboard/src/repositories/postgres/user.repository.ts
```diff
@@ -41,6 +41,15 @@ export async function findUserByEmail(email: string): Promise<User | null> {
   return await findUserBy({ email: email.toLowerCase() });
 }
 
+export async function countUsers(): Promise<number> {
+  try {
+    return await prisma.user.count();
```
flopper123, line 46: This one doesn't account for deleted users 😄 If you delete admin root user, you get stuck

dashboard/src/lib/auth/safe-redirect.ts
```diff
@@ -0,0 +1,15 @@
+const PLACEHOLDER_ORIGIN = 'http://placeholder.invalid';
+
+// Same-origin paths only; anything that escapes the placeholder origin is an open redirect
+export function toSafeRelativePath(value: string | null | undefined, fallback: string): string {
+  if (!value || !value.startsWith('/') || /[\s\\]/.test(value)) {
```
flopper123: This check fails when Next gives repeated query keys as array.
For instance, if you go to `/signin?callbackUrl=a&callbackUrl=b` it'll fail 😄

dashboard/prisma/migrations/20260919120000_reset_user_role_to_instance_admin/migration.sql
```diff
@@ -0,0 +1,13 @@
+-- MigrateData: "User"."role" now marks the instance admin. Historically every email/password
+-- sign-up was defaulted to 'admin' (OAuth sign-ups were not), so the existing values carry no
+-- meaning. Clear them, then promote the earliest live account: on seeded self-host installs
+-- that is the boot-time admin by construction. From here on only the first sign-up sets it.
+UPDATE "User" SET "role" = NULL;
```
flopper123, line 5: This will not only clear the admin role, it'll clear all user roles. So for prod, it would clear all existing roles, instead of be targeted to selfhost setups only 😄

dashboard/src/app/(app)/[locale]/(signup)/signup/SignupForm.tsx
```diff
@@ -110,20 +123,28 @@ export default function SignupForm({ providers }: SignupFormProps) {
...
           }
 
           baEvent('onboarding-account-created');
 
+          if (redirectTo) {
+            router.push(redirectTo);
+            return;
+          }
+
           const { hadInvitations } = await handlePotentialInvitationsOnAccountCreation();
```
Comment (flopper123) [inline dashboard/src/app/(app)/[locale]/(signup)/signup/SignupForm.tsx:148]: This now early-returns before going the `handlePotentialInvitationsOnAccountCreation` check. Is that on purpose? 😄

flopper123 (changes requested):
Very nice! 😄

dashboard/prisma/migrations/20260919120000_reset_user_role_to_instance_admin/migration.sql
Lindharden: Intentional, role is unused.

dashboard/src/repositories/postgres/user.repository.ts
Lindharden: Deliberate, and I went back and forth on it. Deleting is a soft delete, so the row stays (but this is not the case for historical selfhost instances). If we counted only live users, an instance whose sole account deleted itself would treat the next sign-up as the first user and hand them the admin role, on an instance that still holds dashboards and data. Counting every row means that instance stays closed instead

dashboard/src/app/(app)/[locale]/(signup)/signup/SignupForm.tsx
Comment (Lindharden) [reply dashboard/src/app/(app)/[locale]/(signup)/signup/SignupForm.tsx:148]: On purpose. redirectTo is only set when the sign-up came through an invite link, and it points at the accept-invite page

Comment (flopper123) [reply dashboard/src/app/(app)/[locale]/(signup)/signup/SignupForm.tsx:148]: Okay, however, if you invite a user to multiple dashboards, but only send them one invitation link, they'll then need to visit `/dashboards` to accept the remaining invitation links. I guess that's a good enough approach 🤔

dashboard/src/lib/auth/safe-redirect.ts
Lindharden: Addressed

flopper123 (approved):

dashboard/src/app/(app)/[locale]/(signup)/signup/SignupForm.tsx
Comment (Lindharden) [reply dashboard/src/app/(app)/[locale]/(signup)/signup/SignupForm.tsx:148]: Yes but I checked how others handle this. All accept exactly the invitation behind the link, and leave any others pending for the person to accept or decline individually. Auto-accepting everything would make joining one dashboard a side effect of joining another, and an invitation should stay a choice. So I'm keeping this as is

PR #1104, opened 2026-09-19
Used the full visitor IP for GeoIP lookups instead of the anonymized one

flopper123:
GeoIP now resolves country, subdivision and city from the real visitor IP instead of the /24-truncated one, and the lookup cache is keyed by the full IP so the result is exact. Nothing stored changes; the full IP only ever lives in the in-memory moka cache (20 min idle TTL).

Closes betterlytics/betterlytics-internal#231

Reviewer notes: measured on GeoLite2-City (2026-09-18) with 300k random IPv4s, the truncated lookup differed for 0.4% of addresses (0.03% flipped country, mostly cloud ranges). IPv6 was already lossless. The ASN lookup keeps its /24 key on purpose: the same measurement on GeoLite2-ASN showed zero bot-verdict flips.


Lindharden (changes requested):
The public privacy policy says the IP is "immediately anonymized ... before any processing", and repeats it under "Immediate anonymization". After this PR the full IP is used for the geo lookup and sits in an in-memory cache for up to 20 minutes idle. Using the full IP transiently for geolocation is normal practice and fine under GDPR, but the policy should say what actually happens. Suggested wording for the geographic section:

`We determine the visitor's location from their IP address using a GeoIP database. The IP address is used only in memory for this lookup and is anonymized (last octet removed for IPv4, first 64 bits kept for IPv6) before it is used for anything else. The full IP is never stored.`

